PR #1, opened 2020-06-02
[SDK-284] Initial tooling setup

jdm717:
## What

Initial tooling setup for the repo:
- Base package.json with initial lerna, lerna run, and custom scripts
- GitHub workflows for building, linting, and testing
- Vscode configuration


lerna.json
```diff
@@ -0,0 +1,5 @@
+{
+  "npmClient": "yarn",
+  "version": "independent",
```
danschultz, line 3: Can we change to fixed versioning? We should discuss more, but I'm leaning towards a model where all our SDK packages have the same version. This'll make it easier for developers who need to pull in multiple packages and keep their versions consistent.

README.md
```diff
@@ -1,2 +1,80 @@
...
+
+## Structure
+
+- `./packages`: The packages to be made available on NPM. This directory contains
+  subdirectories that represent different categories of packages:
+  - `./libs`: The libs directory contains libraries that are intended to be
+    reusable across other applications and/or libraries. These are normally
+    published to our private NPM repository.
+  - `./sdks`: The sdks directory contains NPM packages for Vertex's public facing
+    SDKs.
```
danschultz, line 14: Thinking that we keep the packages in this repo flat. I think we'll end up with around 10-12 packages which is manageable with a flat structure.

lerna.json
jdm717: Yeah, I think that makes sense. It'd be a bit nicer to have the versions truly indicate compatibility rather than having to guess and check. Is there a specific version you'd want to move to initially? Seems like we could do a version bump to `0.3.3` for everything to match the current version of the viewer, or we could go with a higher version (like `0.4.0` or something)

PR #2, opened 2020-06-02
[SDK-284] Add libs and sdks

jdm717:
## What

Moves our existing libraries and SDKs over from vertex-web.

## Dependencies

- https://github.com/Vertexvis/vertex-web-tools/pull/11 (build tooling will need to be publicly available before this build can pass)


lerna.json
```diff
@@ -1,5 +1,8 @@
 {
   "npmClient": "yarn",
   "version": "independent",
-  "packages": []
+  "packages": [
+    "packages/libs/*",
+    "packages/sdks/*"
```
danschultz, line 6: https://github.com/Vertexvis/vertex-web-sdks/pull/1/files#r434188839

PR #3, opened 2020-06-02
[SDK-637] Support connecting to FSS

jdm717:
## What

Adds support for connecting to FSS by adding a `streamingClient` option that can be passed with the network configuration. Setting this to `platform` will open a connection to FSS instead of ISS.

As part of setting up the ability to load a scene, support for a new `urn:vertexvis:platform:scene:id` URN was added.

Note: currently, some of the existing behavior around manipulating scenes will not be functional until we complete https://jira.vertexvis.net/browse/SDK-639.

## Testing 

- Configure the SDK to point at `platdev` 
```
viewer.config = {
  network: {
    apiHost: 'https://platform.platdev.vertexvis.io',
    renderingHost: 'wss://stream.platdev.vertexvis.io',
    streamingClient: 'platform'
  },
};
```
- Provide a scene id through `viewer.load('urn:vertexvis:platform:scene:id')`
- Verify an image is rendered

## Dependencies

- https://github.com/Vertexvis/vertex-web-tools/pull/11 (build tooling will need to be publicly available before this build can pass)


jdm717:
Closing for cleaner PR in https://github.com/Vertexvis/vertex-web-sdk/pull/5

PR #4, opened 2020-06-03
[SDK-638] FSS interactions

jdm717:
## What

Adds support for rotate/zoom/pan through the FrameStreamingClient.

As part of this work, I had to modify the camera methods in `graphics3d` to support operations based only on the position/lookat/up vectors of a camera.

## Notes

- Currently using FSS as the streaming client will display the view cube, but it is not possible to interact with, as it is using our `viewer.scene()` api, which will be updated with https://jira.vertexvis.net/browse/SDK-639

## Testing

- Load a platform scene as described in https://github.com/Vertexvis/vertex-web-sdks/pull/3 
  - Verify that mouse interactions work as expected
- Load an eedc scene 
  - Verify that mouse interactions work as expected

## Dependencies

https://github.com/Vertexvis/vertex-web-tools/pull/11


jdm717:
Closing for cleaner PR in https://github.com/Vertexvis/vertex-web-sdk/pull/6

PR #6, opened 2020-06-08
[SDK-638] FSS Interactions

jdm717:
## What

Adds support for rotate/zoom/pan through the FrameStreamingClient.

As part of this work, I had to modify the camera methods in `graphics3d` to support operations based only on the position/lookat/up vectors of a camera.

## Notes

- Currently using FSS as the streaming client will display the view cube, but it is not possible to interact with, as it is using our `viewer.scene()` api, which will be updated with https://jira.vertexvis.net/browse/SDK-639

## Testing

- Load a platform scene as described in https://github.com/Vertexvis/vertex-web-sdks/pull/3 
  - Verify that mouse interactions work as expected
- Load an eedc scene 
  - Verify that mouse interactions work as expected

## Dependencies

https://github.com/Vertexvis/vertex-web-tools/pull/11


packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -1,22 +1,24 @@
-import { Scene, Camera } from '@vertexvis/graphics3d';
+import { Camera, Scene } from '@vertexvis/graphics3d';
 import { Dimensions, Point, Vector3 } from '@vertexvis/geometry';
 import { EventEmitter } from '@stencil/core';
 import { TapEventDetails, TapEventKeys } from './tapEventDetails';
 import { FrameStreamingClient } from '../frame-streaming-client';
+import { vertexvis } from '@vertexvis/frame-stream-protos';
```
danconner-vertex, line 6: very small nit: is this where the linter puts this import? would expect it to be with others from `@vertexvis`. also may not be used?

jdm717: It does look like this is the case, thinking it potentially has to do with the fact that `vertexvis` is lowercase, but the linter prefers to put this import toward the bottom of the list

jdm717:
@danconner-vertex thanks for taking a look! I ended up adding back the initial support for the view-cube, if you don't mind taking another quick look at that (it's copied over aside from https://github.com/Vertexvis/vertex-web-sdk/pull/6/files#diff-5f475b4a9ed40f1d756782f2b89d8fd0R182-R194)

packages/viewer/src/components/viewer-view-cube/readme.md
```diff
@@ -0,0 +1,17 @@
...
+
+
+<!-- Auto Generated Below -->
+
+
+## Properties
+
+| Property | Attribute | Description | Type                      | Default     |
+| -------- | --------- | ----------- | ------------------------- | ----------- |
+| `viewer` | --        |             | `HTMLVertexViewerElement` | `undefined` |
```
danconner-vertex, line 12: do we want to make sure this has some useful content? or not important right now?

jdm717: I would probably be of the opinion that it's not important right now (our dev portal currently only has the README for the viewer component). Certainly open to feedback on that, but I'm thinking the documentation would be good to update once this view cube is fully functional

danconner-vertex: makes sense to me. will approve when it's ready to be switched over to master.

PR #12, opened 2020-06-27
[SDK-738] Add operations to modify the scene camera

danschultz:
## What

This adds operations to the scene to modify its camera. Example usage:

```ts

// update the position
let camera = scene.camera().update({ position: Vector3.forward() });

// rotate around an axis
camera = camera.rotateAroundAxis(Math.PI / 2, Vector3.up());

// fit to a bounding box
camera = camera.fitToBoundingBox(BoundingBox.create(...));

// finally queue a frame to render
camera.render()
```

## Ticket

https://jira.vertexvis.net/browse/SDK-738

## Test Plan

Unit tests

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/stream-api/src/streamApi.ts
+++ b/packages/stream-api/src/streamApi.ts
@@ -8,8 +8,6 @@ type ResponseHandler = (
   response: vertexvis.protobuf.stream.IStreamResponse
 ) => void;
 
-type RecursiveRequired<T> = { [P in keyof T]-?: RecursiveRequired<T[P]> };
-
 export class StreamApi {
   private onResponseDispatcher = new EventDispatcher<
     vertexvis.protobuf.stream.IStreamResponse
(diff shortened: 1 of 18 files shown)


packages/stream-api/src/streamApi.ts
```diff
@@ -51,30 +49,30 @@ export class StreamApi {
...
 
   public replaceCamera({
     camera,
-  }: RecursiveRequired<
-    vertexvis.protobuf.stream.IUpdateCameraPayload
-  >): Promise<vertexvis.protobuf.stream.IStreamResponse> {
-    return this.send({
+  }: vertexvis.protobuf.stream.IUpdateCameraPayload): Promise<
+    vertexvis.protobuf.stream.IStreamResponse
+  > {
```
jdm717, line 61: With removing the `RecursiveRequired` type here, it looks like we'll lose any requirement that properties expected by FSS (like a camera with three vectors) are present. This `RecursiveRequired` type might not work with these changes due to the `frameCorrelationId` being a truly optional property, but I'm curious what your thoughts are on how/if we could continue to require properties like the camera. Seems like without some sort of enforcement here it becomes a bit more work to interact with these APIs

danschultz: I've created a ticket for improvements to Stream API. https://jira.vertexvis.net/browse/SDK-740

PR #14, opened 2020-07-01
[SDK-735] introducing raycaster class

danconner-vertex:
## What

adding method on stream api and raycasting class to perform 'hitItems' requests on the frame streaming service.

## Ticket

https://jira.vertexvis.net/browse/SDK-735

## Test Plan

get token and valid scene, then use something similar to below to make a hit request on tap:
```
viewer.addEventListener('tap', async function(event) {
  const { position } = event.detail;
  const raycaster = await scene.raycaster();
  raycaster.hitItems(position);

});
```
verify message was published to stream, and a valid result was returned.

n/a

n/a

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/42


packages/stream-api/yarn.lock
```diff
@@ -598,10 +598,10 @@
     eslint-plugin-prettier "^3.1.0"
     prettier "^1.19.1"
 
-"@vertexvis/frame-streaming-protos@^0.0.16":
-  version "0.0.16"
-  resolved "https://registry.yarnpkg.com/@vertexvis/frame-streaming-protos/-/frame-streaming-protos-0.0.16.tgz#a534347e5923e78bc2463ef1f6c7598462b097a2"
-  integrity sha512-H1kxC7QZ1kfQt7KvnwRo3oYDNrw7okPXwBrtoDY4tvFkwA+Jkb65X6vmDelA1ZJZKEA16FefI245s9Ti7DGLSA==
+"@vertexvis/frame-streaming-protos@^0.0.19":
+  version "0.0.19"
+  resolved "https://npm.ops.vertexvis.io/@vertexvis%2fframe-streaming-protos/-/frame-streaming-protos-0.0.19.tgz#b03b06252a6144d43ccb95e5613eefdeb85bcb83"
```
danschultz, line 603: This is causing the build to fail. Most of us probably have `@vertexvis` mapped to our private registry. We should probably add an `.npmrc` file to the root of the project that maps `@vertexvis` to our public NPM repo.

```
@vertexvis:registry=https://registry.npmjs.org/
registry=https://registry.npmjs.org/
```

packages/stream-api/src/streamApi.ts
```diff
@@ -77,7 +91,7 @@ export class StreamApi {
   ): Promise<vertexvis.protobuf.stream.IStreamResponse> {
     return new Promise(resolve => {
       const subscription = this.onResponse(response => {
-        if (response.frame != null) {
+        if (response.frame != null || response.hitItems) {
```
danschultz, line 94: This could be error prone, as you might have simultaneous requests for hit items. I think it'd be better to pass an optional request ID to `sendRequest` and switch the behavior if it's supplied to check that the response contains the request ID.

danconner-vertex: @danschultz how does this look 6850723ebd175963696e2b934b4b214b696a5772?

packages/stream-api/src/streamApi.ts
```diff
@@ -15,6 +15,8 @@ export class StreamApi {
 
   private messageSubscription?: Disposable;
 
+  private requestSubscriptions: { [requestId: string]: Disposable } = {};
```
danschultz, line 18: This doesn't look like it's used.

PR #17, opened 2020-07-15
[SDK-740] stream api improvements

danconner-vertex:
## What

- stream-api should be able to tag a request as a fire-and-forget type of requests. In this scenario, a returned promise should resolve immediately without a response.
- stream-api should be able to tag a request as a request-response. In this scenario, a returned promise should resolve with the response it receives from FSS, or with an error.
- stream-api should expose typings of request bodies with optional fields marked as nullable, and required fields marked as non-null.

## Ticket

https://jira.vertexvis.net/browse/SDK-740

## Test Plan

TBD

TBD

TBD

## Dependencies

n/a

### DIFF
--- a/packages/stream-api/src/streamApi.ts
+++ b/packages/stream-api/src/streamApi.ts
@@ -54,112 +54,72 @@ export class StreamApi {
 
   public startStream(
     data: StartStreamPayload,
-    withResponse: boolean
-  ): Promise<vertexvis.protobuf.stream.IStreamResponse>;
-
-  public startStream(...args: unknown[]): unknown {
-    const request = this.parseArgs<
-      vertexvis.protobuf.stream.StartStreamPayload
-    >(args);
-    return this.sendRequest({
-      requestId: request.requestId,
-      startStream: { ...request.data },
-    });
+    withResponse = true
+  ): Promise<vertexvis.protobuf.stream.IStreamResponse | void> {
+    return this.sendRequest(
+      {
+        startStream: data,
+      },
+      withResponse
+    );
   }
-
-  public beginInteraction(): Promise<void>;
 
   public beginInteraction(
-    withResponse: boolean
-  ): Promise<vertexvis.protobuf.stream.IStreamResponse>;
-
-  public beginInteraction(...args: unknown[]): unknown {
-    const request = this.parseArgs<
-      vertexvis.protobuf.stream.BeginInteractionPayload
-    >(args);
-    return this.sendRequest({
-      requestId: request.requestId,
-      beginInteraction: { ...request.data },
-    });
+    withResponse = true
+  ): Promise<vertexvis.protobuf.stream.IStreamResponse | void> {
+    return this.sendRequest(
+      {
+        beginInteraction: {},
+      },
+      withResponse
+    );
   }
-
-  public replaceCamera({ camera }: ReplaceCameraPayload): Promise<void>;
 
   public replaceCamera(
     { camera }: ReplaceCameraPayload,
-    withResponse: boolean
-  ): Promise<vertexvis.protobuf.stream.IStreamResponse>;
-
-  public replaceCamera(...args: unknown[]): unknown {
-    const request = this.parseArgs<
-      vertexvis.protobuf.stream.IUpdateCameraPayload
-    >(args);
-    const requestPayload = {
-      requestId: request.requestId,
-      updateCamera: { ...request.data },
-    };
-    return this.sendRequest(requestPayload);
+    withResponse = true
+  ): Promise<vertexvis.protobuf.stream.IStreamResponse | void> {
+    return this.sendRequest({ updateCamera: { camera } }, withResponse);
   }
-
-  public hitItems({ point }: HitItemsPayload): Promise<void>;
 
   public hitItems(
     { point }: HitItemsPayload,
-    withResponse: boolean
-  ): Promise<vertexvis.protobuf.stream.IStreamResponse>;
-
-  public hitItems(...args: unknown[]): unknown {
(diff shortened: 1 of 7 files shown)


packages/stream-api/src/streamApi.ts
```diff
@@ -38,78 +43,125 @@ export class StreamApi {
...
+        camera: request.data as vertexvis.protobuf.stream.ICamera,
       },
-    });
+    };
+    return this.sendRequest(requestPayload);
   }
 
+  public hitItems({
+    point,
+  }: vertexvis.protobuf.stream.IHitItemsPayload): Promise<void>;
```
danconner-vertex, line 99: unlikely this will be a call anyone wants to do, but providing makes the generic `sendRequest` simpler. will keep looking at alternatives.

packages/stream-api/src/streamApi.ts
```diff
@@ -38,78 +43,125 @@ export class StreamApi {
...
+
+  public endInteraction(...args: unknown[]): unknown {
+    const request = this.parseArgs(args);
     return this.sendRequest({
+      ...request.requestId,
       endInteraction: {},
     });
   }
 
+  private parseArgs(args: unknown[]): ParseArgsRequest {
```
danschultz, line 130: Consider making this a generic method, this way you don't have to do explicit casting.

```ts
private parseArgs<T>(args: unknown[]): ParseArgsRequest[T] {
  ...
}

public hitItems(...args: unknown[]): unknown {
  const req = this.parseArgs[vertexvis.protobuf.stream.IHitItemsPayload](args);
  req.data; // vertexvis.protobuf.stream.IPoint
}
```

packages/stream-api/src/streamApi.ts
```diff
@@ -1,13 +1,18 @@
...
-import { Disposable, EventDispatcher } from '@vertexvis/utils';
+import { vertexvis, google } from '@vertexvis/frame-streaming-protos';
+import { Disposable, EventDispatcher, UUID } from '@vertexvis/utils';
 
 type ResponseHandler = (
   response: vertexvis.protobuf.stream.IStreamResponse
 ) => void;
 
+interface ParseArgsRequest {
+  requestId?: google.protobuf.IStringValue;
```
danschultz, line 12: Do the protobufs require request ID's to be of type `IStringValue`? Or do they support `requestId?: IStringValue | string`? If the later, might be a bit simpler to just represent this as `requestId?: string`.

packages/stream-api/src/streamApi.ts
danconner-vertex: is the `[T]` mapping to `data` somehow? it's giving me an error when I use that syntax, not finding examples or docs on using in that way.

packages/stream-api/src/streamApi.ts
danconner-vertex: right now, it can be `IStringValue` or `null`.

packages/stream-api/src/streamApi.ts
danconner-vertex: worked on this here 2a8ab2ecd65df1f22f09543f94dc67b02f34f798

packages/viewer/src/scenes/raycaster.ts
```diff
@@ -17,7 +17,9 @@ export class Raycaster {
    */
   public async hitItems(
     point: Point.Point
-  ): Promise<vertexvis.protobuf.stream.IStreamResponse> {
-    return this.stream.hitItems(UUID.create(), { point });
+  ): Promise<vertexvis.protobuf.stream.IHitItemsResult> {
+    return this.stream
+      .hitItems(UUID.create(), { point })
```
danschultz, line 22: Maybe it'd be better for `stream-api` to assume that you want a response and have the package generate the request ID for you, instead of passing it in. We could use a boolean as the last argument to indicate this is a fire-and-forget type request, which won't supply a request ID. What do you think?

packages/stream-api/src/streamApi.ts
```diff
@@ -36,80 +50,131 @@ export class StreamApi {
...
+
+  public replaceCamera(...args: unknown[]): unknown {
+    const request = this.parseArgs<
+      vertexvis.protobuf.stream.IUpdateCameraPayload
+    >(args);
+    const requestPayload = {
+      requestId: request.requestId,
+      updateCamera: { ...request.data },
+    };
+    return this.sendRequest(requestPayload);
```
danschultz, line 101: It appears we can simplify this implementation quite a bit now :) I don't think there's a need to use method overloading anymore if we use a default for the `withResponse` argument.

IMO we should default `withResponse` to `true`. I think the majority of situations you're going to want a response, and ignoring the response is a more exceptional use-case.

```ts
  public replaceCamera(
    { camera }: ReplaceCameraPayload,
    withResponse: boolean = true
  ): unknown {
    return this.sendRequest({ updateCamera: camera }, withResponse);
  }

  private sendRequest(
    request: vertexvis.protobuf.stream.IStreamRequest,
    withResponse: boolean
  ): Promise<void | vertexvis.protobuf.stream.IStreamResponse> {
    if (withResponse) {
      return new Promise(resolve => {
        const requestId = UUID.create();
        const subscription = this.onResponse(response => {
          if (request.requestId?.value == requestId) {
            resolve(response);
            subscription.dispose();
          }
        });
        this.websocket.send(
          vertexvis.protobuf.stream.StreamMessage.encode({ request }).finish()
        );
      });
    } else {
      this.websocket.send(
        vertexvis.protobuf.stream.StreamMessage.encode({ request }).finish()
      );
      return Promise.resolve();
    }
  }
```

danconner-vertex: yeah, I was thinking that way, just not all the way there yet :)

PR #18, opened 2020-07-20
[SDK-776] Applying material/show/hide operations 

MadisonEhlers-Vertex:
## What

This work is designed to expose apis to consumers of vertex-web-sdk to enable them to do `.itemOperation()`s on scenes. 
The available parameters here are the following: 
- Query:
`.where()`
- Query options: `.withItemId()`, `.or()`, `.and()`, `.withSuppliedId()`. (note nested queries are not supported, so and/ors cannot be used in conjunction with this pr)

- Operations:
1. `.materialOverrideFromHex()` Accepts a string in hex form: `#c9919f`
2. `materialOverride()` Accepts a Material object.
3. `.show()` shows all selected items from the query
4. `.hide()` hides all selected items from the query

Here is an example of the material override:

platform scene operations

Here is an example of hiding a part (the whole model)
platform hide

## Ticket

https://vertexvis.atlassian.net/browse/SDK-776
https://vertexvis.atlassian.net/browse/SDK-691

## Test Plan

using these operations, you should see the selected parts change with the selected operations. ie, given an execution on a material override operation with a valid partId or supplied Part id, you should see those parts change colors. 

example:
```

      import {ColorMaterial} from './build/index.esm.js';
...
          scene
            .items()
            .where(q => q.withItemId('a885f431-d39f-4293-b9d1-cb97a8573840'))
            .materialOverride(ColorMaterial.fromHex('#42bff5'))
            .show()
            .execute();
```

N/A

N/A

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/54


packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+    return this;
+  }
+}
+
+export class SceneQueryBuilder implements SceneItemQuery {
+  private query: SelectorBuilder<ItemSelectorBuilder> | undefined;
+
+  public where(
+    _query: SelectorBuilder<ItemSelectorBuilder>
```
danschultz, line 236: if there's a naming collision, convention is to make the member variable use a `_`.

packages/viewer/src/index.html
```diff
@@ -35,8 +35,38 @@
...
+          console.log('tapping: ', event);
+          const { position } = event.detail;
+          const scene = await viewer.scene();
+          const raycaster = await scene.raycaster();
+
+          // const test = scene.material("1234").material("1111").execute();
+
+          const abc = scene
+            .itemOperation()
+            .where(q => q.itemId('abc').or().q.suppliedId('123'))
```
danschultz, line 51: `.or().q` is `.q` needed?

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+export interface SelectorBuilder<T> {
+  /**
+   * Returns the built selector.
+   */
+  build(): T;
+
+  /**
+   * Returns a conditional builder to perform an `or` operation.
+   */
+  or(): SelectorBuilder<OrSelector>;
```
danschultz, line 151: I'd recommend splitting the operations (`or`, `and`, etc) and the `build` method into separate interfaces. Maybe `SceneQuery`? Then `SelectorBuilder` could extend from this new interface. This way the types won't allow you to call `.build()` inside a where clause.

packages/viewer/src/scenes/scene.ts
```diff
@@ -3,6 +3,58 @@ import { Frame } from '../types';
...
+  ItemSelectorBuilder,
+  SceneItemQuery,
+} from './operations';
+
+export class SceneItemOperationsExecutor
+  implements SceneItemOperations<SceneItemOperationsExecutor> {
+  protected builder = new SceneOperationBuilder();
+  private queryExecutor: SceneItemQueryExecutor;
+  public constructor(
+    queryExecutor?: SceneItemQueryExecutor,
```
danschultz, line 19: you can also do `queryExecutor = new SceneItemQueryExecutor()`, which is terser IMO.

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+export type ItemOperation = ShowItemOperation | HideItemOperation;
+
+export interface SceneItemQuery {
+  where(
+    query: SelectorBuilder<ItemSelectorBuilder>
+  ): SceneItemOperationsExecutor;
+}
+
+export interface SceneItemOperations<T> {
+  material(color: string): T;
```
danschultz, line 189: Let's rename this to `materialOverride`. Also, I think we should introduce a type that represents the material. This way we can extend this to support different types of materials in the future. There's other fields on our ColorMaterial too, so having a separate type would allow others to be specific in what gets updated. See https://github.com/Vertexvis/vertex-api/blob/master/src/main/resources/schemas/ColorMaterial.yml for how this is defined in vertex-api.

We could also have a helper `ColorMaterial.fromHex()` which just sets the diffuse field.

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+
+/**
+ * A selector builder to perform boolean `or` operations.
+ */
+export class OrSelectorBuilder implements SelectorBuilder<OrSelector> {
+  private builders?: SelectorBuilder<ItemSelector>[] = [];
+
+  private parent: SelectorBuilder<ItemSelector>;
```
danschultz, line 59: what's `parent` intended for?

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+    this.query = { type: 'item-id', value: itemId };
+    return this;
+  }
+}
+
+/**
+ * A selector builder to perform boolean `or` operations.
+ */
+export class OrSelectorBuilder implements SelectorBuilder<OrSelector> {
+  private builders?: SelectorBuilder<ItemSelector>[] = [];
```
danschultz, line 57: No need to make this optional, as you're assigning it to a default empty array.

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+    this.builders.push(new ItemSelectorBuilder().withItemId(partId));
+    return this;
+  }
+}
+
+/**
+ * A selector builder to perform boolean `and` operations.
+ */
+export class AndSelectorBuilder implements SelectorBuilder<AndSelector> {
+  private builders?: SelectorBuilder<ItemSelector>[] = [];
```
danschultz, line 92: same here

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+   */
+  withItemId(itemId: string): this;
+}
+
+interface ChangeMaterialOperation {
+  type: 'change-material';
+  color: string;
+}
+
+export type SceneOperation = ChangeMaterialOperation;
```
danschultz, line 178: `ChangeMaterialOperation` is an operation that happens at an item level.

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,241 @@
...
+interface AndSelector {
+  type: 'and';
+  selectors: ItemSelector[];
+}
+
+export interface OperationDefinition {
+  operation: ItemOperation | SceneOperation;
+}
+
+export interface SelectorBuilder<T> {
```
danschultz, line 142: We also need to handle queries that operate on everything, e.g. `showAll()`, `hideAll()`, `clearAllMaterialOverrides()`. Maybe we could so something like:

```ts
scene.where(q => q.all())
  .show()
  .execute()
```

When calling `.all()` it should return a terminal type that doesn't allow further chaining.

packages/viewer/src/index.html
MadisonEhlers-Vertex: yeah, it should be. This wasn't working yet.

packages/viewer/src/scenes/operations.ts
MadisonEhlers-Vertex: Is there backend support for this yet?

packages/viewer/src/scenes/operations.ts
MadisonEhlers-Vertex: copied this from vertex-web, haven't finished this yet.

packages/viewer/src/index.html
MadisonEhlers-Vertex: .q() should not be needed and does not work. Addressed.

packages/viewer/src/scenes/operations.ts
MadisonEhlers-Vertex: I have split these out now, but am having a bit of trouble converting my `Selector` to a `SelectorBuilder`. Was wondering if you could provide any suggestions or insight into this problem?

packages/viewer/src/scenes/scene.ts
MadisonEhlers-Vertex: this piece was removed

packages/viewer/src/scenes/operations.ts
MadisonEhlers-Vertex: Chatted with Jared/Kevin about this. Currently there is no backend support for the showAll, hideAll, clear all. Might be worth doing as a follow up item.

packages/stream-api/src/streamApi.ts
```diff
@@ -92,6 +92,17 @@ export class StreamApi {
     );
   }
 
+  public createSceneAlteration(
+    request: vertexvis.protobuf.stream.ICreateSceneAlterationRequest
+  ): Promise<vertexvis.protobuf.stream.IStreamResponse> {
```
danschultz, line 97: Pattern here is to include a `withResponse` as the last argument that defaults to `true`.

packages/viewer/src/index.html
```diff
@@ -35,8 +35,21 @@
...
+          const { position } = event.detail;
+          const scene = await viewer.scene();
+          const raycaster = await scene.raycaster();
+          scene
+            .itemOperation()
+            .where(q => q.withItemId('a885f431-d39f-4293-b9d1-cb97a8573840'))
+            .materialOverrideFromHex('#c9919f')
+            .show()
+            .execute();
+        });
```
danschultz, line 51: Please revert before merging.

packages/viewer/src/scenes/__tests__/scene.spec.ts
```diff
@@ -22,6 +29,8 @@ describe(Scene, () => {
     });
   });
 
+  // describe(Scene.prototype.itemOperation)
+
```
danschultz, line 33: Should remove this unnecessary code.

packages/viewer/src/scenes/colorMaterial.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return {
+    opacity: 100,
+    glossiness: opacity || 10,
+    ambient: {
+      ...color,
+    },
+    diffuse: {
+      ...color,
+    },
```
danschultz, line 26: AFAIU, colors should only affect `diffuse`. The others are fixed values.

packages/viewer/src/scenes/scene.ts
```diff
@@ -3,6 +3,75 @@ import { Frame } from '../types';
...
+  protected builder = new SceneOperationBuilder();
+  public constructor(
+    private sceneViewId: UUID.UUID,
+    private commands: CommandRegistry,
+    private query: ItemSelectorBuilder
+  ) {}
+
+  public materialOverrideFromHex(color: string): SceneItemOperationsExecutor {
+    return this.materialOverride(fromHex(color));
+  }
```
danschultz, line 28: Unless there's a strong reason to, my preference would be to not have these helpers. These helpers could get pretty numerous, and would be better to organize them on the material type, e.g. `ColorMaterial`.

packages/viewer/src/scenes/scene.ts
```diff
@@ -3,6 +3,75 @@ import { Frame } from '../types';
...
+  Selector,
+  ItemSelectorBuilder,
+} from './selectors';
+import { CommandRegistry } from '../commands/commandRegistry';
+import { UUID } from '@vertexvis/utils';
+
+export class SceneItemOperationsExecutor
+  implements SceneItemOperations<SceneItemOperationsExecutor> {
+  protected builder = new SceneOperationBuilder();
+  public constructor(
```
danschultz, line 20: nit, spacing between properties and constructor.

packages/viewer/src/scenes/colorMaterial.ts
MadisonEhlers-Vertex: If I didnt pass all of the values, it did not work.

packages/viewer/src/scenes/selectors.ts
```diff
@@ -0,0 +1,236 @@
...
+   * @param suppliedId The customer ID of an item to match
+   */
+  withSuppliedId(suppliedId: string): this;
+
+  /**
+   * Selects a item that has the given Vertex ID.
+   *
+   * @param itemId The ID of the item to match.
+   */
+  withItemId(itemId: string): this;
```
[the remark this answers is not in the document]
MadisonEhlers-Vertex: This gets back to how that builder is structured. Would appreciate some thoughts on a direction/implementation for this builder problem.

danschultz:
@MadisonEhlers-Vertex here's a gist with some code snippets and example usage. https://gist.github.com/danschultz/830dda36b588a453fad25b405e6eac6f 

Couple notes:

* These queries are immutable. Each operation will return a new query that can't be modified outside of the query building process.
* OR and AND query types encode our current restrictions around boolean operations, e.g. not being able to combine and/or queries.
* A valid query is represented by the `TerminalQuery` interface. This way the caller can't return an incomplete query. For example you couldn't do `query.withItemId('1').or()`. That makes no sense :)
* Calling `TerminalQuery.build()` will return an `Expression`. This expression will then need to be converted into an object that is accepted by `stream-api`. One nice thing about this, is you basically have a AST that represents the query expression. You can convert this into whatever format you want.. a protobuf object, a string, etc.

packages/viewer/src/scenes/colorMaterial.ts
danschultz: Ya, but I don't think we should set `ambient`, `specular`, `emissive` to `color`. Those should probably be some constants that are hardcoded.

MadisonEhlers-Vertex:
> @MadisonEhlers-Vertex here's a gist with some code snippets and example usage. https://gist.github.com/danschultz/830dda36b588a453fad25b405e6eac6f
> 
> Couple notes:
> 
> * These queries are immutable. Each operation will return a new query that can't be modified outside of the query building process.
> * OR and AND query types encode our current restrictions around boolean operations, e.g. not being able to combine and/or queries.
> * A valid query is represented by the `TerminalQuery` interface. This way the caller can't return an incomplete query.
> * Calling `TerminalQuery.build()` will return an `Expression`. This expression will then need to be converted into an object that is accepted by `stream-api`. One nice thing about this, is you basically have a AST that represents the query expression. You can convert this into whatever format you want.. a protobuf object, a string, etc.

Thanks for the help @danschultz! I have updated the pr to reflect these changes.

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,69 @@
...
+  type: 'show';
+}
+
+interface HideItemOperation {
+  type: 'hide';
+}
+
+export interface OperationDefinition {
+  operation: ItemOperation;
+}
```
danschultz, line 13: This looks like a superfluous type. Just pass `ItemOperation`.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -47,7 +49,133 @@ export function startStream(
...
+
+    return stream.createSceneAlteration(request);
+  };
+}
+
+function buildSceneOperation(
+  query: QueryExpression,
+  operations: OperationDefinition[]
+): vertexvis.protobuf.stream.ISceneOperation {
+  const operationTypes: vertexvis.protobuf.stream.IOperationType[] = buildOperationTypes(
```
danschultz, line 76: Type defs are unnecessary within function definitions. Just do `const operationTypes = buildOperationTypes()`

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -47,7 +49,133 @@ export function startStream(
...
+    case 'supplied-id':
+      return {
+        item: {
+          sceneItemQuery: buildSceneItemQuery(query),
+        },
+        operationTypes,
+      };
+    default:
+      return {};
+  }
```
danschultz, line 101: This can be simplified

```ts
function buildSceneOperation(
  query: QueryExpression,
  operations: OperationDefinition[]
): vertexvis.protobuf.stream.ISceneOperation {
  const operationTypes = buildOperationTypes(operations);

  switch (query.type) {
    case 'and':
    case 'or':
      return {
        [query.type]: {
          queries: query.expressions.map(exp => ({
            sceneItemQuery: buildSceneItemQuery(exp),
          })),
          operationTypes,
        },
      };
    case 'item-id':
    case 'supplied-id':
      return {
        item: {
          sceneItemQuery: buildSceneItemQuery(query),
        },
        operationTypes,
      };
    default:
      return {};
  }
}
```

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -47,7 +49,133 @@ export function startStream(
...
+        hex: sceneViewId,
+      }),
+      operations: pbOperations,
+    };
+
+    return stream.createSceneAlteration(request);
+  };
+}
+
+function buildSceneOperation(
```
danschultz, line 72: This adds a lot of new behavior. I'd recommend moving `buildSceneOperation` and the functions it calls into a new module.

packages/stream-api/src/__tests__/streamApi.spec.ts
```diff
@@ -45,6 +46,16 @@ describe(StreamApi, () => {
     });
   });
 
+  describe('send createSceneAlteration', () => {
+    beforeEach(() => api.connect(url));
+
+    it('should send a request with Id', () => {
+      const request: vertexvis.protobuf.stream.ICreateSceneAlterationRequest = {};
```
danschultz, line 53: This type def shouldn't be necessary. Removes the need for an import and makes this line terser.

packages/viewer/src/commands/__tests__/streamCommands.spec.ts
```diff
@@ -21,4 +25,84 @@ describe('streamCommands', () => {
...
+        sceneViewId,
+        builtQuery,
+        operations
+      )({
+        stream,
+        tokenProvider: tokenProvider,
+        config,
+      });
+
+      console.log(fromHex('#FF00AA'));
```
danschultz, line 70: Unnecessary logging.

packages/viewer/src/commands/__tests__/streamCommands.spec.ts
```diff
@@ -21,4 +25,84 @@ describe('streamCommands', () => {
       expect(stream.startStream).toHaveBeenCalledWith({ ...dimensions });
     });
   });
+
+  describe('createSceneAlteration', () => {
+    it('sends a create alteration request with the given params', async () => {
+      const sceneViewId: UUID.UUID = UUID.create();
+      const sceneItemId: UUID.UUID = UUID.create();
```
danschultz, line 32: Unnecessary type defs.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -47,7 +49,133 @@ export function startStream(
...
+export function createSceneAlteration(
+  sceneViewId: UUID.UUID,
+  query: QueryExpression,
+  operations: OperationDefinition[]
+): Command<Promise<vertexvis.protobuf.stream.IStreamResponse>> {
+  return ({ stream }: CommandContext) => {
+    const pbOperations: vertexvis.protobuf.stream.ISceneOperation[] = [
+      buildSceneOperation(query, operations),
+    ];
+    const request: vertexvis.protobuf.stream.ICreateSceneAlterationRequest = {
```
danschultz, line 61: Unnecessary type defs

packages/viewer/src/scenes/colorMaterial.ts
```diff
@@ -0,0 +1,49 @@
...
+      g: 0.2,
+      b: 0.2,
+      a: 64,
+    },
+    emissive: {
+      r: 0.1,
+      g: 0,
+      b: 0.1,
+      a: 1,
+    },
```
danschultz, line 47: The ranges of these values should be between 0-255. Since these rgb values are so close to 0, I'd suggest making them 0 so they don't appear as special values. Also, `a` (alpha) isn't used internally, might as well be set to 0 as well. Maybe it's clearer if we just use the constructor function `Color.create(0, 0, 0)`.

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  private operation(...args: any[]): this {
+    const operation = args[0];
+    if (args.length === 1) {
+      this.operations.push({ operation });
+    } else if (args.length === 2) {
+      this.operations.push({ operation });
+    }
+    return this;
+  }
```
danschultz, line 68: We never pass more than a single argument to `operation`. Might as well consolidate this implementation to `private operation(operation: ItemOperation): SceneOperationBuilder`

packages/viewer/src/scenes/operations.ts
```diff
@@ -0,0 +1,69 @@
...
+   * Constructs the scene operations and returns a definition describing each
+   * operation.
+   */
+  public build(): OperationDefinition[] {
+    return this.operations.concat();
+  }
+
+  public materialOverride(color: ColorMaterial): SceneOperationBuilder {
+    return this.operation({ type: 'change-material', color });
+  }
```
danschultz, line 48: Since the types in `queries.ts` are immutable, I think it'd be good to keep things consistent make this builder immutable as well, e.g. return a new `SceneOperationBuilder` with the concated operation.

packages/viewer/src/scenes/scene.ts
```diff
@@ -3,6 +3,52 @@ import { Frame } from '../types';
...
+import { UUID } from '@vertexvis/utils';
+
+/**
+ * A class that is responsible for building operations for a specific scene.
+ * This executor requires a query, and expects `execute()` to be invoked in order
+ * for the changes to take effect.
+ */
+export class SceneItemOperationsExecutor
+  implements SceneItemOperations<SceneItemOperationsExecutor> {
+  protected builder = new SceneOperationBuilder();
```
danschultz, line 19: This can be private.

packages/viewer/src/scenes/scene.ts
```diff
@@ -3,6 +3,52 @@ import { Frame } from '../types';
...
+  public constructor(
+    private sceneViewId: UUID.UUID,
+    private commands: CommandRegistry,
+    private query: QueryExpression
+  ) {}
+
+  public materialOverride(color: ColorMaterial): SceneItemOperationsExecutor {
+    this.builder.materialOverride(color);
+    return this;
+  }
```
danschultz, line 30: This class can be immutable as well.

packages/viewer/src/scenes/operations.ts
MadisonEhlers-Vertex: Not sure what you're implying here. If I do a new `SceneOperationBuilder().materialOverride()` this will be a recursive call.

packages/viewer/src/index.html
```diff
@@ -28,20 +29,61 @@
...
+          // Set the size of the color picker
+          width: 320,
+          // Set the initial color to pure red
+          color: "#f00"
+        });
+
+        colorPicker.on('color:change', async (givenColor) => {
+          color = givenColor;
+        });
+        viewer.addEventListener('tap', async function(event) {
```
danschultz, line 68: we probably don't want to have the tokens and color picker in master.

MadisonEhlers-Vertex: I didn't mean to commit this. Will revert.

PR #22, opened 2020-07-29
[SDK-791] Support for query by all, and clear operations

MadisonEhlers-Vertex:
## What

Support for query by `.all()` and support to clear materials overrides. 

ie:
```
            await scene
              .items()
              .all()
              .hide()
              .execute();

           await scene
              .items()
              .where(q => q.withItemId(YOUR_ITEM_ID))
              .clear()
              .execute()
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-791

## Test Plan

Ensure that the given operations are applying to all for the all query, and that operations are being cleared on the clear operation

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/scenes/operations.ts
+++ b/packages/viewer/src/scenes/operations.ts
@@ -8,6 +8,10 @@ interface HideItemOperation {
   type: 'hide';
 }
 
+interface ClearItemOperation {
+  type: 'clear';
+}
+
 export interface ChangeMaterialOperation {
   type: 'change-material';
   color: ColorMaterial;
@@ -16,12 +20,14 @@ export interface ChangeMaterialOperation {
 export type ItemOperation =
   | ShowItemOperation
   | HideItemOperation
-  | ChangeMaterialOperation;
+  | ChangeMaterialOperation
+  | ClearItemOperation;
 
 export interface SceneItemOperations<T> {
   materialOverride(color: ColorMaterial): T;
   show(): T;
   hide(): T;
+  clear(): T;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/viewer/src/scenes/operations.ts
```diff
@@ -16,12 +20,14 @@ export interface ChangeMaterialOperation {
...
   | HideItemOperation
-  | ChangeMaterialOperation;
+  | ChangeMaterialOperation
+  | ClearItemOperation;
 
 export interface SceneItemOperations<T> {
   materialOverride(color: ColorMaterial): T;
   show(): T;
   hide(): T;
+  clear(): T;
```
MadisonEhlers-Vertex, line 30: Do we want to call this `.clear()` or should we call this `.clearOverrides()`?

Not sure what sort of verbiage we want to expose in the sdk

packages/viewer/src/scenes/scene.ts
```diff
@@ -55,6 +55,15 @@ export class SceneItemOperationsExecutor
     );
   }
 
+  public clear(): SceneItemOperationsExecutor {
```
danschultz, line 58: `clear` is pretty vague about what's being cleared. How about `clearMaterialHighlight()`?

packages/viewer/src/scenes/queries.ts
```diff
@@ -148,4 +148,16 @@ export class SceneItemQueryExecutor {
...
+    const allExpresion: QueryExpression = {
+      type: 'all',
+    };
+
+    return new SceneItemOperationsExecutor(
+      this.sceneViewId,
+      this.commands,
+      allExpresion
+    );
+  }
```
danschultz, line 162: We shouldn't need this method. There's already an `RootQuery.all` that can be accessed via: `scene.items().where(q => q.all())`. With this, there'd be two places where query expressions can be defined. Should keep it to a single place so things are consistent.

MadisonEhlers-Vertex: Yeah, I had considered that as an alternative option. If that's the way we wanna represent that, sounds good to me! 👍

PR #24, opened 2020-07-30
[SDK-797] Support CI releases via Lerna

danschultz:
## What

Updates our CI/CD pipeline to use `lerna publish from-package` internally. This will automatically check if a version exists in NPM and if not, publish the package.

Made some workflow changes to the pipeline as well. There's now two workflows:

* `branch.yml`: Is run on any git branch. It contains 3 jobs: `build`, `test` and `lint`. Lint and build run in parallel, so you'll get notified sooner if that stage fails.
* `master.yml`: Is run on any merge to master. It contains 3 jobs: `build`, `detect-publish`, `publish`. Publish will only run if a version change is detected in the `lerna.json` file. If a version change is detected, we'll publish packages to NPM and create a GH release.

Also added dependency caching to reduce the time for builds.

There's some setup duplication among the scripts. Plan is to pull these out into a GH action so we can share across jobs / repos.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-797

## Test Plan

1. Merge
2. Create a relaese
2. Verify that packages are released

N/A

N/A

## Dependencies

N/A


.github/workflows/master.yml
```diff
@@ -0,0 +1,91 @@
...
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout changes"
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
jdm717, line 15: nit: do these jobs need a `fetch-depth` of 2? Doesn't look like we're working with anything but the current commit

.github/workflows/master.yml
```diff
@@ -0,0 +1,91 @@
...
+          path: packages/*/dist
+  detect-publish:
+    runs-on: ubuntu-latest
+    outputs:
+      publish: ${{ steps.detect_publish.outputs.publish }}
+    steps:
+      - name: "Checkout changes"
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
jdm717, line 49: Same as above

.github/workflows/master.yml
```diff
@@ -0,0 +1,91 @@
...
+          echo ::set-output name=publish::$result
+  publish:
+    runs-on: ubuntu-latest
+    needs: [build, detect-publish]
+    if: needs.detect-publish.outputs.publish == 1
+    steps:
+      - name: "Checkout changes"
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
jdm717, line 64: Same as above

.github/workflows/master.yml
danschultz: 🤷‍♂️  probably a relic from copying the code from somewhere else. let me try with just pulling the last commit.

PR #25, opened 2020-07-30
[SDK-798] Support Graceful Reconnection

MadisonEhlers-Vertex:
## What

when given a server request from fss to the client to gracefully reconnect, 
we want to establish a new WS connection with a reconnect stream so the stream doesn't die. 

## Ticket

https://vertexvis.atlassian.net/browse/SDK-798

## Test Plan

Ensure that after a timeout duration, you can still render. 

N/A

N/A

## Dependencies

N/A


packages/stream-api/src/streamApi.ts
```diff
@@ -62,6 +84,26 @@ export class StreamApi {
...
   }
 
+  public async reconnect(
+    data: ReconnectPayload,
+    withResponse = true
+  ): Promise<Disposable> {
+    if (this.urlProvider == null) {
+      throw new Error('Unable to connect as no Url provider has been set');
+    }
+    await this.websocket.connect(this.urlProvider);
```
danschultz, line 94: We should try to keep the responsibility of the client simpler and have this method just be responsible for sending a reconnect message. In other words, the consumer would call:

```ts
await api.connect();
api.reconnect();
```

Understand there's a little weirdness with the naming, but think it'd be better if we keep the convention of methods that accept payloads aren't doing anything more than sending a message over the WS.

packages/stream-api/src/webSocketClient.ts
```diff
@@ -89,7 +91,9 @@ export class WebSocketClient {
     this.reopenAttempt += 1;
 
     try {
-      await this.connect(urlProvider);
+      if (!this.retryDisabled) {
+        await this.connect(urlProvider);
+      }
```
danschultz, line 96: A better way to do this would be to use `clearTimeout` for the timer that's created above.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -50,6 +50,20 @@ export function startStream(
...
 
+export function reconnect(
+  streamId: UUID.UUID,
+  dimensions: Dimensions.Dimensions
+): Command<Promise<Disposable>> {
+  return ({ stream }: CommandContext) => {
+    return stream.reconnect({
+      streamId: new vertexvis.protobuf.core.Uuid({
+        hex: streamId,
+      }),
```
danschultz, line 61: Wrapping this in a `new vertexvis.protobuf.core.Uuid` is unnecessary I think. You should be able to do:
```ts
{
  streamId: { hex: streamId }
}
```

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -349,9 +350,13 @@ export class Viewer {
 
     return new Promise(async resolve => {
       try {
-        await this.commands.execute('stream.connect', {
-          sceneId: scene.id,
-        });
+        const dispose: Disposable = await this.commands.execute(
```
danschultz, line 353: `await this.commands.execute<Disposable>` is more idiomatic in TS.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -50,6 +50,20 @@ export function startStream(
   };
 }
 
+export function reconnect(
+  streamId: UUID.UUID,
+  dimensions: Dimensions.Dimensions
+): Command<Promise<Disposable>> {
```
danschultz, line 56: Looks like this should return a `vertexvis.protobuf.stream.IStreamResponse` instead.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -369,6 +374,32 @@ export class Viewer {
...
+          this.dimensions
+        );
+        this.streamDisposables.push(dispose);
+        resolve(streamId);
+      } catch (e) {
+        this.errorMessage = 'Unable to maintain connection to Vertex';
+        throw new WebsocketConnectionError(this.errorMessage, e);
+      }
+      resolve(streamId);
+    });
```
danschultz, line 400: There's quite a bit of duplication here with `connectStreamingClient`. Consider extracting this out into another method that can be shared between `connectStreamingClient` and `reconnectStreamingClient`. You can use a function to parameterize the command execution which is different between these calls. Rough implementation might look like:

```ts
  private connectStream(connection: Promise<Disposable>): Promise<void> {
    this.stream = new StreamApi(
      new WebSocketClient(),
      this.stream.getUrlProvider()
    );

    return new Promise(async () => {
      try {
        const dispose = await connection;
        this.setupStreamListeners();
        this.streamDisposables.push(dispose);
      } catch (e) {
        this.errorMessage = 'Unable to maintain connection to Vertex';
        throw new WebsocketConnectionError(this.errorMessage, e);
      }
    });
  }

  private async connectStreamingClient(resource: string): Promise<string | null | undefined> {
    const scene = SceneResource.fromUrn(resource);
    await this.connectStream(
      this.commands.execute<Disposable>('stream.connect', { sceneId: scene.id })
    );

    const response = await this.stream.startStream({ ...this.dimensions });
    
    // note: this is different from previous implementation, and returns the scene view id
    // vs settings it. think this is better as it keeps the implementation more like a pure function.
    return response.startStream.sceneViewId.hex;
  }

  private async reconnectStreamingClient(streamId: string): Promise<void> {
    await this.connectStream(
      this.commands.execute<Disposable>('stream.reconnect', { sceneId: scene.id })
    );

    await this.stream.reconnect({ streamId, ...this.dimensions });
  }
```

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -570,11 +620,4 @@ export class Viewer {
   private initializeCredentials(): void {
     this.activeCredentials = parseToken(this.token);
   }
-
-  private getBackgroundColor(): Color.Color | undefined {
-    if (this.containerElement != null) {
-      const colorString = window.getComputedStyle(this.containerElement);
-      return Color.fromCss(colorString.backgroundColor);
-    }
-  }
```
danschultz, line 579: Can we keep this around, and document its for future use?

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: I think the implementations are different enough, and in one use case we want a new stream, where in another instance we do not (the initial start up), especially after the other pr feedback has been addressed. PTAL

packages/stream-api/src/webSocketClient.ts
MadisonEhlers-Vertex: Not sure which timer you are referring to here.

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: On second thought, I have made a common function for just the execution of the connection to the stream

packages/stream-api/src/webSocketClient.ts
danschultz: It's the timer id that's returned on line 83. Instead you should store the ID of that timer, and then call `clearTimeout` when `close` is called:

```js
const timer = setTimeout(() => console.log("hi"), 100);
clearTimeout(timer);
```

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -369,6 +366,41 @@ export class Viewer {
...
+    });
+  }
+
+  private reconnectStreamingClient(streamId: UUID.UUID): Promise<string> {
+    this.streamDisposables.forEach(d => d.dispose());
+
+    this.stream = new StreamApi(
+      new WebSocketClient(),
+      this.stream.getUrlProvider()
+    );
```
danschultz, line 388: Looking again at this, can't we just reuse the existing `StreamApi` and `stream.connect` for reconnections? Would make the implementation quite a bit simpler because we're just leveraging existing code. We'd also not need to pass down those `StreamProvider`s.

The implementation of `reconnectStreamingClient` would look something like:

```ts
  private reconnectStreamingClient(streamId: UUID.UUID): Promise<string> {
    return new Promise(async resolve => {
      await this.connectStream(
        this.commands.execute('stream.connect', {
          // You can pass `loadedSceneId` here instead of having a new `stream.reconnectWebSocket` command
          sceneId: this.loadedSceneId,
        })
      );

      // We can just use `StreamApi.reconnect` without having to define a new command.
      this.stream.reconnect({
        streamId: { hex: streamId },
        dimensions: this.dimensions,
      });
      resolve(streamId);
    });
  }
```

You don't need an array of stream disposables or the call to `setupStreamListeners` anymore.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -130,6 +130,7 @@ export class Viewer {
 
   private isResizing?: boolean;
   private sceneViewId?: UUID.UUID;
+  private streamDisposables: Disposable[] = [];
```
danschultz, line 133: See prior comment, but I don't think this array of disposables anymore if we don't recreate `StreamApi`.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -50,6 +50,26 @@ export function startStream(
...
+): Command<Promise<vertexvis.protobuf.stream.IStreamResponse>> {
+  return ({ stream }: CommandContext) => {
+    return stream.reconnect({
+      streamId: {
+        hex: streamId,
+      },
+      dimensions,
+    });
+  };
+}
```
danschultz, line 71: See prior comment, but these commands probably aren't necessary anymore.

packages/stream-api/src/streamApi.ts
```diff
@@ -62,6 +87,25 @@ export class StreamApi {
     );
   }
 
+  public async reconnectWebSocket(): Promise<Disposable> {
+    if (this.urlProvider == null) {
+      throw new Error('Unable to connect as no Url provider has been set');
+    }
+    return this.connect(this.urlProvider);
+  }
```
danschultz, line 95: Doesn't look like this is called.

packages/stream-api/src/streamApi.ts
```diff
@@ -46,10 +63,18 @@ export class StreamApi {
     this.messageSubscription?.dispose();
   }
 
+  public getUrlProvider(): UrlProvider | undefined {
+    return this.urlProvider;
+  }
```
danschultz, line 68: Shouldn't need if we can reuse the existing `StreamApi` instance.

packages/stream-api/src/streamApi.ts
```diff
@@ -27,17 +32,29 @@ export class StreamApi {
...
+  private onRequestDispatcher = new EventDispatcher<
+    vertexvis.protobuf.stream.IStreamRequest
+  >();
+
   private messageSubscription?: Disposable;
 
   public constructor(
-    private websocket: WebSocketClient = new WebSocketClient()
+    private websocket: WebSocketClient = new WebSocketClient(),
+    private urlProvider?: UrlProvider
```
danschultz, line 43: This arg wouldn't be needed either.

packages/stream-api/src/streamApi.ts
```diff
@@ -27,17 +32,29 @@ export class StreamApi {
...
 
-  public async connect(urlProvider: UrlProvider): Promise<Disposable> {
-    await this.websocket.connect(urlProvider);
+  public async connect(urlProvider?: UrlProvider): Promise<Disposable> {
+    if (urlProvider != null) {
+      this.urlProvider = urlProvider;
+    }
+    if (this.urlProvider == null) {
+      throw new Error('Unable to connect as no Url provider has been set');
+    }
```
danschultz, line 52: `urlProvider` can be a required arg, and wouldn't need these null checks.

packages/viewer/src/commands/commandRegistry.ts
```diff
@@ -9,11 +9,13 @@ interface CommandDefinition<R> {
   thisArg: any | undefined;
 }
 
+export type StreamProvider = () => StreamApi;
```
danschultz, line 12: We wouldn't need this either.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -306,9 +307,9 @@ export class Viewer {
   @Method()
   public async load(resource: string): Promise<void> {
     if (this.commands != null && this.dimensions != null) {
-      this.loadedSceneId = this.connectStreamingClient(resource);
+      const sceneId = this.connectStreamingClient(resource);
 
-      await this.loadedSceneId;
+      this.loadedSceneId = await sceneId;
```
danschultz, line 312: These temp variables aren't necessary. This can be simplified to

```ts
this.loadedSceneId = await this.connectStreamingClient(resource)
```

packages/stream-api/src/streamApi.ts
```diff
@@ -27,17 +32,29 @@ export class StreamApi {
...
 
-  public async connect(urlProvider: UrlProvider): Promise<Disposable> {
+  public async connect(
+    urlProvider: UrlProvider,
+    reconnect?: boolean
+  ): Promise<Disposable> {
+    if (reconnect) {
+      console.log('creating a new websocket client');
+      this.websocket = new WebSocketClient();
+    }
```
danschultz, line 52: Let's hop on a call to discuss, but this shouldn't be necessary.

packages/stream-api/src/webSocketClient.ts
MadisonEhlers-Vertex: I have updated this to save the timer, and clear it on a close event.

packages/stream-api/src/webSocketClient.ts
```diff
@@ -95,11 +81,41 @@ export class WebSocketClient {
...
+
+  private removeWebSocketListeners(): void {
+    if (this.webSocket != null) {
+      this.webSocket.removeEventListener('message', this.handleMessage);
+      this.webSocket.removeEventListener('open', () =>
+        this.onOpen(Promise.resolve)
+      );
+      this.webSocket.removeEventListener('error', () =>
+        this.onError(Promise.reject)
+      );
```
danschultz, line 104: This won't actually remove the listeners, and will result in a memory leak. You need to pass the same reference to the function that passed to `addEventListener`.

danschultz: Here's what I'd recommend doing. Define a function that closes over the function references and return a disposable that can later be called to clean up the WS.

```ts
  private addWebSocketListeners = (
    ws: WebSocket,
    resolve: VoidFunction,
    reject: VoidFunction
  ): Disposable => {
    const onOpen = (): void => this.onOpen(resolve);
    const onError = (): void => reject();
    ws.addEventListener('message', this.handleMessage);
    ws.addEventListener('open', onOpen);
    ws.addEventListener('error', onError);
    ws.addEventListener('close', this.onClose);

    return {
      dispose: () => {
        ws.removeEventListener('message', this.handleMessage);
        ws.removeEventListener('open', onOpen);
        ws.removeEventListener('error', onError);
        ws.removeEventListener('close', this.onClose);
      },
    };
  };

  private removeWebSocketListeners(): void {
    this.listeners?.dispose();
  }
```

packages/stream-api/src/webSocketClient.ts
```diff
@@ -15,12 +15,18 @@ export class WebSocketClient {
   private webSocket?: WebSocket;
   private onMessageDispatcher = new EventDispatcher<MessageEvent>();
   private reopenAttempt = 0;
+  private urlProvider?: UrlProvider;
+  private timer?: NodeJS.Timer;
```
danschultz, line 19: This type isn't correct because we're not in a Node environment. I would recommend changing the type to `number` and using `window.setTimeout`.

MadisonEhlers-Vertex: image

this is the type provided by `setTimeout`

MadisonEhlers-Vertex: used window.setTimeout resolves this. addressed.

packages/stream-api/src/webSocketClient.ts
MadisonEhlers-Vertex: Thanks!

PR #26, opened 2020-07-31
toggle es5 build for easier use of component

danconner-vertex:
## What

tweaking build to import / use in app.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-760


danconner-vertex:
going a different route here.

PR #38, opened 2020-08-07
[SDK-829] Fix reconnection

danschultz:
## What

There was a bug where reconnecting to a WS would fail.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-829

## Test Plan

1. View a model
2. Wait 45 seconds
3. Try to rotate the model


packages/viewer/src/commands/streamCommands.ts
```diff
@@ -2,35 +2,30 @@ import { vertexvis } from '@vertexvis/frame-streaming-protos';
...
+        `/stream-keys/${streamKey}/session`,
+        Uri.parse(config.network.renderingHost)
+      );
 
-        return {
-          url: Uri.toString(uri),
-          protocols: ['ws.vertexvis.com'],
-        };
-      } else {
-        throw new InvalidCredentialsError(`Provided credentials are invalid.`);
```
MadisonEhlers-Vertex, line 32: won't we still want this to throw this exception here? Maybe stream key is invalid?

PR #42, opened 2020-08-13
[SDK-850] Default to prod environment for websdk

MadisonEhlers-Vertex:
## What

Expose production environment for vertex web sdk and use it as the default. 

To leverage the viewer and use a development environment, you will have to pass the env to the viewer. 

```
        viewer.configEnv = 'platdev';
```

## Test Plan

Ensure rendering works for platprod and platdev

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/config/config.ts
+++ b/packages/viewer/src/config/config.ts
@@ -32,7 +32,6 @@ function getEnvironmentConfig(environment: Environment): Config {
   switch (environment) {
     case 'platdev':
       return platdevConfig;
-    case 'platprod':
     default:
       return platprodConfig;
   }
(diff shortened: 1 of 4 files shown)


packages/viewer/src/config/config.ts
```diff
@@ -21,12 +21,25 @@ const platdevConfig: Config = {
...
+    renderingHost: 'wss://stream.platprod.vertexvis.io',
+  },
+};
+
 function getEnvironmentConfig(environment: Environment): Config {
-  return platdevConfig;
+  switch (environment) {
+    case 'platdev':
+      return platdevConfig;
+    case 'platprod':
```
danschultz, line 35: This line is unnecessary.

PR #44, opened 2020-08-17
[SDK-800] Consume viewer as an NPM package

danschultz:
## What

This fixes issues that prevented the viewer from being consumed as an NPM package. The root cause was there was multiple TS library helpers being included in our bundles and causing compile errors in Webpack. Fixed this by enabling `importHelpers` in our TS configs and adding `tslib` as a dependency.

Also started extracted shared Rollup and TS configs into a new internal `@vertexwebsdk/build` package.

Example of importing the viewer using a bundler like Webpack:

```js
import { defineCustomElements } from '@vertexvis/viewer/loader';

async function main() {
  const viewer = document.querySelector("viewer");
  await viewer.load("urn:vertexvis:stream-key:123");
  console.log("Loaded!");
}

definedCustomElement(window).then(() => main());
```

If you want to target IE11 or Edge <= 18, you'll also need to import `applyPolyfills`.

```js
import { applyPolyfills, defineCustomElements } from '@vertexvis/viewer/loader';

function main() {
  console.log("Loaded!");
}

applyPolyfills().then(() => definedCustomElement(window)).then(() => main());
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-800

## Test Plan

Tested by including the viewer as an NPM dependency to an example Webpack project.

Also validated that viewer can be included as a `<script>` tag in HTML.

N/A

N/A


internal/build/src/rollup.ts
```diff
@@ -0,0 +1,55 @@
...
+type Config = {
+  /**
+   * Indicates if the generated config will output both Node and browser
+   * bundles. This is useful for library projects that need to support both
+   * environments.
+   */
+  isMultiPlatform?: boolean;
+};
+
+export function rollupConfig({ isMultiPlatform = false }: Config = {}):
```
MadisonEhlers-Vertex, line 21: Not sure if there'd be any easy way to test this so that this doesn't regress?

danschultz: This is a bit difficult to test, but added some rudimentary ones.

PR #53, opened 2020-08-27
[SDK-796] Support for Transactional Queries

MadisonEhlers-Vertex:
## What

You should be able to do multiple operations in one transactional request to the server. IE, if you want to do a show only on an item, today you would have to do two separate executions:
```
   await scene.items().where(q => q.all()).hide().execute();
   await scene.items().where(q => q.withItemId(ITEM_ID)).show().execute();
```

This PR includes changes to allow you to pass multiple operations inside one execution context. This will break syntax of existing consumers, but will break type checking if you happen to be using typescript. 

Introducing the new syntax for item operations:
```
      scene
        .items(
          op => [
            op.where(q => q.all()).hide(),
            op.where(q => q.withItemId(itemId).or().withSuppliedId(suppliedId)).show(),
            op.where(q => q.all()).materialOverride(ColorMaterial.fromHex("#ff1122"))
          ])
        .execute();
```

Now instead of items returning the scene item operations executor, that entity has been modified to become a builder where you may include one or many operations. The `items()` expects a function that returns either a single op, or an array of ops. Ie, you can use the above syntax, or simply just do one operation in your items invocation:

```
      scene
        .items(
            op => op.where(q => q.withItemId(itemId)).hide())
        .execute();
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-796

## Test Plan

Try to bundle multiple operations. Your first frame that responds should have applied all of the operations you specified. 
For example, you can do a show only lie this:
```
      scene
        .items(
          op => [
            op.where(q => q.all()).hide(),
            op.where(q => q.withItemId(YOUR_ITEM)).show()
          ])
        .execute();
```

Item operations

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/scenes/scene.ts
+++ b/packages/viewer/src/scenes/scene.ts
@@ -4,7 +4,7 @@ import { Camera } from './camera';
 import { Dimensions } from '@vertexvis/geometry';
 import { Raycaster } from './raycaster';
 import { ColorMaterial } from './colorMaterial';
-import { SceneItemOperations, SceneOperationBuilder } from './operations';
+import { SceneItemOperations, SceneOperationBuilder, ItemOperation } from './operations';
 import { QueryExpression, SceneItemQueryExecutor } from './queries';
 import { CommandRegistry } from '../commands/commandRegistry';
 import { UUID } from '@vertexvis/utils';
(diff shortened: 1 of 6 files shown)


packages/viewer/src/scenes/scene.ts
```diff
@@ -89,8 +115,33 @@ export class Scene {
...
+  ): ItemsOperationExecutor {
+    const sceneOperations: TerminalItemOperationBuilder = operations(
+      new SceneItemQueryExecutor(this.sceneViewId, this.commands)
+    );
+    if (Array.isArray(sceneOperations)) {
+      const operationList = sceneOperations.reduce(
+        (acc, builder: SceneItemOperationsBuilder) =>
+          acc.concat(builder.build()),
+        []
+      );
```
danschultz, line 134: `operationList` will actually be typed as `any[]`. property better to do:

```ts
      const operationList = sceneOperations.reduce(
        (acc, builder) => acc.concat(builder.build()),
        [] as QueryOperation[]
      );
```

packages/viewer/src/scenes/scene.ts
```diff
@@ -89,8 +115,33 @@ export class Scene {
...
+      return new ItemsOperationExecutor(
+        this.sceneViewId,
+        this.commands,
+        operationList
+      );
+    } else {
+      return new ItemsOperationExecutor(this.sceneViewId, this.commands, [
+        sceneOperations.build(),
+      ]);
+    }
```
danschultz, line 144: You can remove this if/else by doing something like:

```ts
    const sceneOperations = operations(new SceneItemQueryExecutor());
    const ops = Array.isArray(sceneOperations)
      ? sceneOperations
      : [sceneOperations];

    const operationList = ops.reduce(
      (acc, builder) => acc.concat(builder.build()),
      [] as QueryOperation[]
    );
```

PR #54, opened 2020-08-28
[SDK-912] Adds support for reporting frame timings

danschultz:
## What

This adds acknowledgements for frame requests in the Web SDK. Whenever a `DrawFrameRequest` is received, the SDK will respond with a `DrawFrameResult` that includes the duration of sending the frame from the server to the client.

This also includes:
* Bug fixes for emitting `frameDrawn` and `frameReceived` events. We were not correctly emitting events before.
* Cleanup and improvements to how we mock and stub WS messages. I introduced a `WebSocketClientMock` class to assist in testing our websocket communication.
* More utils. There's now functions for `Async.delay` and `Async.timeout` to delay and reject promises that take too long.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-912

## Test Plan

1. Frame timing metrics should appear in Datadog.

## Dependencies

- [ ] FSS PR


packages/viewer/src/rendering/renderer.ts
```diff
@@ -0,0 +1,80 @@
...
+function timeRender(api: StreamApi, renderer: Renderer): Renderer {
+  return async req => {
+    const requestedAt = new Date();
+    const resp = await renderer(req);
+
+    api.replyResult(resp.id, {
+      drawFrame: {
+        timing: {
+          sequenceNumber: resp.frame.sequenceNumber,
+          sendToReceiveDuration: toProtoDuration(requestedAt, resp.sentAt),
```
danschultz, line 30: This needs to be updated to take into account offsets.

packages/stream-api/src/time.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export function toProtoDuration(
+  start: Date,
+  end: Date
+): google.protobuf.Duration {
+  const millis = end.getTime() - start.getTime();
+  return google.protobuf.Duration.create(parseEpochMillis(millis));
+}
+
+/* eslint-disable padding-line-between-statements */
```
MadisonEhlers-Vertex, line 47: Do we need to disable the eslint rules here? Would prefer to block the rule as a whole if we're going to add these disables in files.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -36,6 +36,7 @@ export function connect({
   };
 }
 
+// TODO(dan): remove, we can call the api directly.
```
MadisonEhlers-Vertex, line 39: You planning to do anything with this in this pr? Also why would we remove this? I thought going through FSS was to reduce latency on requests.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -396,77 +422,68 @@ export class Viewer {
...
         );
         const fitTo = Rectangle.fromDimensions(this.dimensions);
         const fit = Rectangle.containFit(fitTo, imageRect);
 
         const scaleX = fit.width / imageRect.width;
         const scaleY = fit.height / imageRect.height;
 
-        const startXPos = imagePosition.x * scaleX;
-        const startYPos = imagePosition.y * scaleY;
+        const startXPos = imageAttributes.imageRect.x * scaleX;
```
MadisonEhlers-Vertex, line 474: Will these changes resolve this issue? https://vertexvis.atlassian.net/browse/SDK-909

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -541,7 +558,7 @@ export class Viewer {
   }
 
   private setupStreamListeners(): void {
-    this.stream.onRequest(request => this.handleStreamRequest(request));
+    this.stream.onRequest(msg => this.handleStreamRequest(msg.request));
```
MadisonEhlers-Vertex, line 561: Could probably be renamed `onMessage` now

packages/viewer/src/components/viewer/viewer.tsx
danschultz: Not yet. Have no idea what's causing that annoying bug.

packages/viewer/src/components/viewer/viewer.tsx
danschultz: `onRequest` will still filter messages to only include requests. I changed what gets passed to it, because we're including additional timing headers on the message that were not available on `StreamRequest`.

packages/viewer/src/commands/streamCommands.ts
danschultz: That comment is perhaps misleading. I meant that you can use `StreamApi` directly. I'm considering removing the concept of the `CommandRegistry` altogether. Turns out it's need has diminished over time.

packages/stream-api/src/time.ts
danschultz: Without this the method overloads are not grouped. IMO it's clearer that these are overloaded methods without whitespace between them.

Method overloads are fairly rare, so I think it's okay to add in these exceptions vs globally disabling the lint rule. I'll usually consider removing the lint rule only if I find myself adding an exception every few days.

packages/stream-api/src/index.ts
```diff
@@ -3,3 +3,10 @@ export * from './streamApi';
 export * from './webSocketClient';
 
 export * from './connection';
+
+export * from './encoder';
+
+export * from './time';
+
+// TODO(dan): should be moved to a separate bundle.
```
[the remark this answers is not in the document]
danschultz: Need to investigate the effort with this, but will try to make it part of this PR.

packages/stream-api/src/time.ts
MadisonEhlers-Vertex: Sounds good to me. Thanks for the explanation.

packages/viewer/src/commands/streamCommands.ts
MadisonEhlers-Vertex: Ah, I understand. Perhaps removing the TODO, or updating the text if you choose to leave it in upon merging.

danschultz: Annotated this with `@deprecated`.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -36,6 +36,9 @@ export function connect({
   };
 }
 
+/**
+ * @deprecated Call StreamApi.createSceneAlteration directly.
```
MadisonEhlers-Vertex, line 40: 👍

danschultz:
> LGTM! The timing being reported looks like what I'd expect for the send->receive of a frame. Not seeing anything related to the processing/paint related timings, will we want to add that as part of this PR/ticket? Makes sense as follow-up to me, just curious

Ya, I'll do follow ups as necessary. These set of changes have become pretty large.

PR #64, opened 2020-09-16
[SDK-953] Handling concurrent loads of same resource

MadisonEhlers-Vertex:
## What

Handling loading the same stream key twice concurrently. 

## Test Plan

Try loading the same stream key twice, concurrently. This action should not fail. 

```
          viewer.load(`urn:vertexvis:stream-key:${streamKey}`); 
          viewer.load(`urn:vertexvis:stream-key:${streamKey}`); 
```

N/A

N/A

## Dependencies

N/A


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -310,8 +311,21 @@ export class Viewer {
     await this.unload();
```
danschultz, line 311: `unload` should only be called if we initiate a new WS connection.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -310,8 +311,21 @@ export class Viewer {
...
 
     if (this.commands != null && this.dimensions != null) {
-      this.resource = LoadableResource.fromUrn(urn);
-      await this.connectStreamingClient(this.resource);
+      const loadableResource = LoadableResource.fromUrn(urn);
+
+      const isCurrentlyLoadingResource =
+        this.connectingPromise != null &&
+        this.resource != null &&
+        this.resource.id === loadableResource.id;
```
danschultz, line 319: Technically it's possible to have different types of resources. So you also need to check if the resource types are equal.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -310,8 +311,21 @@ export class Viewer {
...
+      if (isCurrentlyLoadingResource) {
+        return;
+      }
```
danschultz, line 322: Instead of returning eagerly, it's generally easier to understand if the code that should be executed is contained within the condition block.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -310,8 +311,21 @@ export class Viewer {
...
+      this.resource = loadableResource;
+
+      this.connectingPromise = this.connectStreamingClient(this.resource);
+
+      await this.connectingPromise;
+      this.connectingPromise = undefined;
```
danschultz, line 328: This should probably be cleared when `unload` is called as well. Otherwise a hanging connection is going to prevent a second call to load without anyway to get the viewer out of that state.

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: This is handled within the function. It checks if there is a streamDisposable present, otherwise does nothing.

danschultz (approved):

PR #69, opened 2020-09-21
[SDK-870] [SDK-906] Update tsconfig to use shared

MadisonEhlers-Vertex:
## What

Updating to use shared tsconfig

## Ticket

https://vertexvis.atlassian.net/browse/SDK-870
https://vertexvis.atlassian.net/browse/SDK-906
## Test Plan

General regression testing. 
- Viewer
- alterations
- hit items
- camera adjustments

queries should now support a list of ids `q.withItemIds([...])`

Removed deprecated scene alterations stream request in favor of calling the stream api directly.


packages/viewer/src/components.d.ts
```diff
@@ -41,7 +41,7 @@ export namespace Components {
           * Internal API.
           * @private
          */
-        "registerCommand": <R, T>(id: string, factory: CommandFactory<R>, thisArg?: T) => Promise<Disposable>;
+        "registerCommand": <R, T>(id: string, factory: CommandFactory<R>, thisArg?: T | undefined) => Promise<Disposable>;
```
danconner-vertex, line 44: was this a result of the config change?

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -161,7 +164,7 @@ export class InteractionApi {
    */
   public async endInteraction(): Promise<void> {
     if (this.isInteracting()) {
-      this.currentCamera = null;
+      this.currentCamera = undefined;
```
danconner-vertex, line 167: what was the difference in the tsconfig that prompted all the changes from null to undefined?

packages/viewer/src/interactions/tapInteractionHandler.ts
```diff
@@ -87,18 +87,21 @@ export class TapInteractionHandler implements InteractionHandler {
...
-    if (Point.distance(this.pointerDownPosition, pointerUpPosition) <= 1) {
-      this.interactionApi?.tap(
-        this.getCanvasPosition(pointerUpPosition),
-        keyDetails
-      );
+    if (
+      this.pointerDownPosition != null &&
+      Point.distance(this.pointerDownPosition, pointerUpPosition) <= 1
+    ) {
+      const position = this.getCanvasPosition(pointerUpPosition!);
```
danconner-vertex, line 101: should we be concerned about this lint warning?

packages/viewer/src/scenes/__tests__/queries.spec.ts
```diff
@@ -72,4 +72,50 @@ describe(RootQuery, () => {
...
+      itemId.toString(),
+    ]);
+
+    /*
+
+              {
+          value: suppliedId,
+          type: 'supplied-id',
+        },
+      */
```
danconner-vertex, line 88: is this comment useful?

packages/viewer/src/scenes/__tests__/scene.spec.ts
```diff
@@ -13,8 +13,11 @@ describe(Scene, () => {
   } as unknown) as CommandRegistry;
   const renderer = jest.fn();
   const sceneViewId: UUID.UUID = UUID.create();
+  const streamApi = {
+    createSceneAlteration: jest.fn(),
+  };
   const scene = new Scene(
-    new StreamApi(),
+    (streamApi as unknown) as StreamApi,
```
danconner-vertex, line 20: this looks odd, why do we have to cast it twice?

packages/viewer/src/types/frame.ts
```diff
@@ -48,12 +66,24 @@ export const fromProto = (
...
+        lookAt: Vector3.create({
+          x: sceneAttributes.camera.lookAt?.x || undefined,
+          y: sceneAttributes.camera.lookAt?.y || undefined,
+          z: sceneAttributes.camera.lookAt?.z || undefined,
+        }),
+        up: Vector3.create({
+          x: sceneAttributes.camera.up?.x || undefined,
+          y: sceneAttributes.camera.up?.y || undefined,
+          z: sceneAttributes.camera.up?.z || undefined,
+        }),
```
danconner-vertex, line 83: I'm not too great at typescript types, but sensing this being necessary is a sign a type could be more useful.

packages/viewer/src/commands/streamCommands.ts
```diff
@@ -44,29 +40,6 @@ export function connect({
...
-      operations: pbOperations,
-    };
-
-    return stream.createSceneAlteration(request);
-  };
-}
-
 export function registerCommands(commands: CommandRegistry): void {
   commands.register('stream.connect', connect);
-  commands.register('stream.createSceneAlteration', createSceneAlteration);
```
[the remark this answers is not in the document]
MadisonEhlers-Vertex: This function was deprecated. Originally requests went through the command registry, now it calls the stream api directly.

packages/viewer/src/components.d.ts
MadisonEhlers-Vertex: yes it was, It is automatically updating for me here.

packages/viewer/src/interactions/interactionApi.ts
MadisonEhlers-Vertex: The difference is that the ts config of `--strictNullChecks` is now turned on. Would be happy to discuss turning it off, but this was the result of pulling the shared tsconfig.

packages/viewer/src/scenes/__tests__/queries.spec.ts
MadisonEhlers-Vertex: It was at one point. :) Will remove

packages/viewer/src/types/frame.ts
MadisonEhlers-Vertex: These are actually both typed. The issue here, I believe is related to an above comment on strict null types. This may be one argument for removing that strict rule. This code is trying to set the type `(number|null);` to `(number | undefined)` and it is unable to do so. Removing the strict rule would prevent these changes from being necessary.

MadisonEhlers-Vertex: @jdm717 @danschultz What are your guys thoughts on this rule?

jdm717 (commented):
Overall lgtm, had some general questions/nits

packages/viewer/src/types/frame.ts
jdm717: Personally I think I'd prefer to leave that rule enabled. It adds a bit of overhead, but it seems like it's worth the tradeoff to avoid some pretty easy to introduce app-breaking bugs

packages/viewer/tsconfig.json
```diff
@@ -1,19 +1,14 @@
...
-    "declaration": false,
+    "inlineSourceMap": true,
     "experimentalDecorators": true,
-    "lib": ["dom", "es2017"],
-    "moduleResolution": "node",
-    "module": "esnext",
-    "target": "es2017",
-    "noUnusedLocals": false,
-    "noUnusedParameters": false,
+    "strictFunctionTypes": false,
```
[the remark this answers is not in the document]
MadisonEhlers-Vertex: I was having a hard time with the `commandRegistry` always returning a Stream Api. I believe I have this addressed here in the changes I am about to push up

jdm717 (approved):
LGTM!

packages/viewer/src/types/frame.ts
danschultz: Because of how type coercion works in JS, if `camera.up.z = 0`, this will cause the returned value to be undefined. You should use `??` instead of `||` here.

PR #71, opened 2020-09-23
[SDK-962] Resolve Console Errors

MadisonEhlers-Vertex:
## What

These changes are to include some better error handling, as well as not throwing console exceptions on the event that the client is dropping frames. After investigation into the issue noted in the ticket, I noticed that when this event was happening, it happened when the frames got backed up, and then the user switches scenes. This would then close the old WS, and open a new one. This led me to find that the issue was that somewhere the client was resolving the promises, even though none of the responses were coming from the server. The root issue, pertains to setting a timeout on the frames. I am proposing in these changes that we remove this timeout so that on the event of dropped frames these errors don't show up to the client.

If there is an argument for keeping the timeout in, another change could be included to catch the exception and do a `console.warn` instead of the errors currently seen today. 

I noted another issue in the SDK where switching between scenes rapidly would cause the `Error Connecting to Vertex` message would show, in addition to a console error. This was caused by some changes late last week pertaining to the `.load` functionality. The changes included here are to _not_ show that exception if there is a frame present on the current scene. These changes are the ones in the file `viewer.tsx`.

A third issue I noticed was when changing between scenes rapidly, the call to notify on performance was failing as this call happens _last_ in a chain of events. For now I propose this exception be caught, and replaced with a `console.warn` as seen below. 

## Ticket

https://vertexvis.atlassian.net/browse/SDK-962

## Test Plan

Initial error:
- Switch around the scenes and zoom/interact with each scene. Do this until you can see even a slight delay in the frames being returned.
- Immediately switch scenes and let it sit for 10s. You should see no errors in your console. 

Switching Scenes Rapidly & performance exceptions:
- Given the example in the ticket with 5+ scenes, load the example, and rapidly click 5 distinct buttons (thus initiating 5 separate scene loads)
- You should see the final scene load in
- Wait 15s. You should not see the connection error message, nor a console exception pertaining to a promise timeout. 
- Take a look at the console. You should only see warnings pertaining to the performance calls, not console errors. 

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -402,9 +402,11 @@ export class Viewer {
 
       await this.waitNextDrawnFrame(15 * 1000);
     } catch (e) {
-      this.errorMessage = 'Unable to establish connection to Vertex.';
-      console.error('Failed to establish WS connection', e);
-      throw new WebsocketConnectionError(this.errorMessage, e);
+      if (this.lastFrame == null) {
+        this.errorMessage = 'Unable to establish connection to Vertex.';
+        console.error('Failed to establish WS connection', e);
+        throw new WebsocketConnectionError(this.errorMessage, e);
+      }
     }
   }
 
--- a/packages/viewer/src/rendering/canvas.ts
+++ b/packages/viewer/src/rendering/canvas.ts
@@ -46,7 +46,11 @@ function reportTimings(api: StreamApi, meter: TimingMeter): void {
     .map(t => ({ receiveToPaintDuration: toProtoDuration(t.duration) }));
 
   if (timings.length > 0) {
-    api.recordPerformance({ timings: timings }, false);
+    try {
+      api.recordPerformance({ timings: timings }, false);
+    } catch (e) {
+      console.warn('Unable to record performance: ', e);
+    }
   }
 }
 
--- a/packages/viewer/src/rendering/remote.ts
+++ b/packages/viewer/src/rendering/remote.ts
@@ -1,10 +1,8 @@
-import { Async, UUID } from '@vertexvis/utils';
+import { UUID } from '@vertexvis/utils';
 import { FrameCamera, Frame } from '../types';
 import { StreamApi, protoToDate } from '@vertexvis/stream-api';
 import { ifDrawFrame } from './utils';
 import { FrameRenderer } from './renderer';
-
-const DEFAULT_TIMEOUT_IN_MS = 10 * 1000; // 10 seconds
 
 export interface FrameRequest {
   correlationId?: string;
@@ -50,7 +48,6 @@ function requestFrame(api: StreamApi): RemoteRenderer {
 
   return req => {
     const corrId = req.correlationId || UUID.create();
-    const timeout = req.timeoutInMs || DEFAULT_TIMEOUT_IN_MS;
     const update = new Promise<FrameResponse>(resolve => {
       requests.set(corrId, resolve);
 
@@ -60,9 +57,9 @@ function requestFrame(api: StreamApi): RemoteRenderer {
       );
     });
 
-    return Async.timeout(timeout, update).finally(() =>
-      requests.delete(corrId)
-    );
+    return update.finally(() => {
+      requests.delete(corrId);
+    });
   };
 }
 
--- a/packages/viewer/src/rendering/remote.ts
+++ b/packages/viewer/src/rendering/remote.ts
@@ -18,9 +18,6 @@ export interface FrameResponse {
   frame: Frame.Frame;
 }
 
-let count = 0;
-let recievedCount = 0;
-
 export type RemoteRenderer = FrameRenderer<FrameRequest, FrameResponse>;
 
 function requestFrame(api: StreamApi): RemoteRenderer {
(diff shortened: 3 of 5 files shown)


packages/viewer/src/rendering/remote.ts
```diff
@@ -60,9 +57,9 @@ function requestFrame(api: StreamApi): RemoteRenderer {
       );
     });
 
-    return Async.timeout(timeout, update).finally(() =>
```
MadisonEhlers-Vertex, line 63: @danschultz I would specifically appreciate your feedback here, as I believe these changes were put in somewhat intentionally, and I am now proposing this timeout be removed.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -402,9 +402,11 @@ export class Viewer {
...
       await this.waitNextDrawnFrame(15 * 1000);
     } catch (e) {
-      this.errorMessage = 'Unable to establish connection to Vertex.';
-      console.error('Failed to establish WS connection', e);
-      throw new WebsocketConnectionError(this.errorMessage, e);
+      if (this.lastFrame == null) {
+        this.errorMessage = 'Unable to establish connection to Vertex.';
+        console.error('Failed to establish WS connection', e);
+        throw new WebsocketConnectionError(this.errorMessage, e);
+      }
```
danconner-vertex, line 409: what is the scenario where `lastFrame` is `null` here? and why do we only want to throw the exception when it is?

packages/viewer/src/rendering/canvas.ts
```diff
@@ -46,7 +46,11 @@ function reportTimings(api: StreamApi, meter: TimingMeter): void {
     .map(t => ({ receiveToPaintDuration: toProtoDuration(t.duration) }));
 
   if (timings.length > 0) {
-    api.recordPerformance({ timings: timings }, false);
+    try {
+      api.recordPerformance({ timings: timings }, false);
+    } catch (e) {
+      console.warn('Unable to record performance: ', e);
+    }
```
danconner-vertex, line 53: is there something specific about this API call that is prone to errors? or the try/catch because it's not an essential function?

packages/viewer/src/rendering/remote.ts
danconner-vertex: @MadisonEhlers-Vertex what is the case for removing the timeout?

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: last frame is null initially, and in between unloading & reloading a scene when a frame gets returned. If the user has a frame.

packages/viewer/src/rendering/remote.ts
MadisonEhlers-Vertex: Some of this was explained above. The main idea is that we don't want the promises to return at all if the client drops the frames. This was the root issue. Adding a timeout here will trigger the timeout on the event that the client dropped frames, even though the server is rightly closing the stream.

packages/viewer/src/rendering/canvas.ts
MadisonEhlers-Vertex: No, this is explained above. The request is sent on the websocket. I don't think that we should be indicating these errors to the client, but still have these tracked and logged as warnings. I think a better solution here is to not record performances on the fss/ws stream and potentially post these metrics to a restful endpoint.

packages/viewer/src/rendering/remote.ts
danschultz: This is very intentional. If removed, there's no way to complete the returned promise in the scenario that the client fails to receive a frame.

danschultz (changes requested):
This issue was easily reproducible for me without the reconnects. The timeout that was in place is intentional in cases the client never receives a draw frame request from FSS. Without it, the promise will never resolve and could hang the caller's code.

Have we taken a look in FSS to see if the frame correlation IDs are being correctly propagated to the rendered frame?

MadisonEhlers-Vertex:
> This issue was easily reproducible for me without the reconnects. The timeout that was in place is intentional in cases the client never receives a draw frame request from FSS. Without it, the promise will never resolve and could hang the caller's code.
> 
> Have we taken a look in FSS to see if the frame correlation IDs are being correctly propagated to the rendered frame?

I was not able to reproduce without switching scenes. I can revisit, but I spent a good 30-45 minutes attempting to reproduce, and did not see this without switching scenes.

danschultz:
> @danschultz I was not able to reproduce without switching scenes. I can revisit, but I spent a good 30-45 minutes attempting to reproduce, and did not see this without switching scenes.

Try spamming rotation of the model for a few seconds. I'm okay because of time permitting to silence the exception, but I'd propose we at minimum:

1. Take a peak thru FSS if we haven't already
2. Consider logging the exception with a warning
3. Complete the returned promise in some way

MadisonEhlers-Vertex:
PR has been updated to log as a warning

packages/viewer/src/components/viewer/viewer.tsx
danschultz: Why is this check necessary?

danschultz (approved):

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: In the event that you spam switching scenes, one of the failed promises will throw an exception here because a promise times out. This only happens when switching between 3+ scenes at once in rapid succession. I think cancelling connection requests or wrapping the connection promise in some disposable would be the long term resolution we'd want to have.

PR #74, opened 2020-10-07
[SDK-1015] Add configuration options for adaptive rendering

danschultz:
## What

Adding options to configure adaptive rendering.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1015

## Test Plan

Rendering without adaptive rendering should not be impacted.


jdm717 (approved):
LGTM

PR #79, opened 2020-11-09
[SDK-1047] Support reconnects when websocket loses connection unexpectedly

jdm717:
## What

Adds support for reconnecting to FSS when the `StreamApi`'s websocketClient unexpectedly closes (in the case of lost internet connection for example) and is unable to receive the graceful reconnect request.

Additionally, this PR sets up behavior in the `StreamApi` to close the underlying websocketClient in the case that there has been a minute (by default) of no requests/responses appearing over the websocket. This is intended to cover the cases where a client has lost internet connection for an extended period of time, but the browser has not detected this and closed the websocket connection, as I ran into this a handful of times while testing this behavior.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1047

## Test Plan

- Test general use of the viewer SDK
  - Reconnect messages should behave as expected and open a new connection if you have internet
  - Interacting with the model should retain the same websocket connection (shouldn't be closed due to no interactivity)
- Test putting your computer to sleep (for around ~1 min or longer to reproduce the issue)
  - The viewer should be interactive after waking the computer up
- Test disconnecting internet (for around ~1 min or longer to reproduce the issue)
  - The viewer should be interactive after reconnecting internet

Graceful reconnection, websocket connections

N/A

## Dependencies

N/A

@Vertexvis/sdk 
PTAL


packages/stream-api/src/streamApi.ts
```diff
@@ -32,9 +32,15 @@ export class StreamApi {
...
+  // Tracks a period of time with no interaction (requests or responses)
+  // indicating that we have lost connection, and the websocket client
+  // used will no longer function as expected
+  private uninteractiveTimeout?: number;
+
   public constructor(
     private websocket: WebSocketClient = new WebSocketClientImpl(),
-    private loggingEnabled = false
+    private loggingEnabled = false,
+    private uninteractiveThreshold: number = 60 * 1000
```
jdm717, line 43: Thinking it might make sense to have this value specified by FSS on `StartStream` messages if we'd want to move forward with this approach.

packages/stream-api/src/streamApi.ts
```diff
@@ -54,6 +68,8 @@ export class StreamApi {
       ...descriptor,
       url: appendSettingsToUrl(descriptor.url, settings),
     };
+    window.addEventListener('offline', this.handleOffline);
+    window.addEventListener('online', this.handleOnline);
```
danschultz, line 72: I'm not sure if this is the right place to put this logic. I'd like to leave open the possibility of using this client in a Node environment, which is possible now by passing a Node compatible `WebSocketClient`. I'd consider maybe moving this into some helper class/module in the viewer package, OR using the decorator pattern to wrap this client to provide this additional behavior.

packages/stream-api/src/webSocketClient.ts
```diff
@@ -80,28 +76,6 @@ export class WebSocketClientImpl implements WebSocketClient {
...
-    });
-
-    this.reopenAttempt += 1;
-
-    try {
-      await this.connect(descriptor);
-    } catch (e) {
-      // Failed connection attempt here will be handled, and this exception can be ignored
-    }
-  }
```
danschultz, line 103: The close logic and removal of event listeners should probably be revisted now as well. In particular, we had to do some hacky things like manually emitting a close event. I wonder if this can be improved by reordering when we close the connection and remove the listeners.

danschultz (approved):

PR #85, opened 2020-11-19
[SDK-1147] Support for double tap and long press events

jdm717:
## What

Adds support for a `doubletap` and `longpress` event. These both default to allowing a second tap to trigger a `doubletap` if it occurs within 500ms, and triggering a `longpress` if a mousedown/touchstart occur and 500ms pass before a mouseup/touchend are received. These durations are configurable through the `viewer.config` property by passing:
```javascript
viewer.config = {
  events: {
    longPressThreshold: {ms},
    doubleTapThreshold: {ms}
  }
}
```

Also made a small adjustment to the touch interaction handler for rotating the camera to allow for single pixel movements after you've initiated a rotation (by moving a distance >= 2 pixels). This allows for more fine-tuned movements of the camera during a touch-based camera interaction.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1147

## Test Plan

- Test adding a listener for a `doubletap` event
  - When a double tap occurs an event containing the position of the tap should be emitted
- Test adding a listener for a `longpress` event
  - When a long press (mouse/touch held down for 500ms) an event containing the position of the longpress should be emitted
- Test `tap` events
  - Tap events should fire for every tap/click (currently even when a doubletap fires)
- Test configuring the threshold for a `doubletap` event
  - Verify that the double tap can occur during this new threshold
- Test configuring the threshold for a `longpress` event
  - Verify that the long press can occur during this new threshold
- Test touch-based camera manipulation
  - After the initial 2 pixel movement, single pixel movements should manipulate the camera

Tap events, touch-based camera manipulation

As mentioned above, the logic for when to manipulate the camera on touch events has been slightly adjusted

## Dependencies

N/A

@Vertexvis/sdk 
PTAL

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -113,6 +113,13 @@ export class Viewer {
   @Event() public doubletap!: EventEmitter<TapEventDetails>;
 
   /**
+   * Emits an event whenever the user taps or clicks a location in the viewer and the
+   * configured amount of time passes without receiving a mouseup or touchend.
+   * The event includes the location of the tap or click.
+   */
+  @Event() public longpress!: EventEmitter<TapEventDetails>;
+
+  /**
    * Emits an event when a frame has been received by the viewer. The event
    * will include details about the drawn frame, such as the `Scene` information
    * related to the scene.
@@ -660,7 +667,8 @@ export class Viewer {
       this.stream,
       () => this.createScene(),
       this.tap,
-      this.doubletap
+      this.doubletap,
+      this.longpress
     );
   }
 
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -198,7 +198,9 @@ export class Viewer {
       this.registerInteractionHandler(new TouchInteractionHandler());
     }
 
-    this.registerInteractionHandler(new TapInteractionHandler());
+    this.registerInteractionHandler(
+      new TapInteractionHandler(() => this.getConfig())
+    );
 
     this.injectViewerApi();
   }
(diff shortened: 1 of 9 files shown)


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -106,6 +106,19 @@ export class Viewer {
    */
   @Event() public tap!: EventEmitter<TapEventDetails>;
 
+  /**
+   * Emits an event whenever the user double taps or clicks a location in the viewer.
+   * The event includes the location of the first tap or click.
+   */
+  @Event() public doubletap!: EventEmitter<TapEventDetails>;
```
brentav, line 113: Is it standard for these event types to be all lowercase vs camel? Camel feels more natural to me personally but I'm good with whatever is more standard.

danschultz (approved):

packages/viewer/src/components/viewer/viewer.tsx
danschultz: In JS/HTML yes, lowercase is the norm. https://developer.mozilla.org/en-US/docs/Web/Events/touchstart

packages/viewer/src/types/events.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * The number of milliseconds that a tap or click has to be held down
+   * before triggering the "longpress" event.
+   */
+  longPressThreshold: number;
+}
+
+export const defaultEventConfig: EventConfig = {
+  doubleTapThreshold: 500,
```
danschultz, line 16: Can't remember, do we have double click in the connect app? Is this what we use there? The value seems a bit high.

jdm717: Yep, we do have a double click gesture in Connect. That's also set to 500ms, but I agree that it might be a bit high. I think that threshold was set quite some time ago and just never revisited. Certainly open to adjusting these

brentav (approved):
Pulled and tested locally. Working well for me. +1

PR #87, opened 2020-11-23
[SDK-1171] Support Ghosting attributes

jdm717:
## What

Adds support for supplying `streamAttributes` as part of the viewer configuration to enable the Ghosting feature when loading a model.

Additionally, this adds a `updateStream` method to the viewer, which allows modification of `streamAttributes` when a stream has already been created to allow for toggling Ghosting on/off and changing opacity.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1171

## Test Plan

- Test enabling Ghosting when starting a stream
- Test toggling Ghosting through the `updateStream` method
- Test that reconnections keep the same attributes that have been set through `updateStream`
- Test that loading of new models keep the same attributes that have been set through `updateStream`

N/A

N/A

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/117

@Vertexvis/sdk 
PTAL


packages/viewer/src/config/config.ts
```diff
@@ -15,6 +15,7 @@ interface NetworkConfig {
 export interface Config {
   network: NetworkConfig;
   flags: Flags.Flags;
+  streamAttributes: StreamAttributes;
```
danschultz, line 18: If possible I think we should limit the number of places this can be set. How about if we just have a watched property `Viewer.streamAttributes`? Whenever it changes, it'll send a request over the WS to update the stream.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -391,6 +393,28 @@ export class Viewer {
...
+   *
+   * @param attributes The set of attributes to update the stream with.
+   */
+  @Method()
+  public async updateStream(attributes: StreamAttributes): Promise<void> {
+    this.streamAttributes = attributes;
+
+    this.stream.updateStream({
+      streamAttributes: attributes,
+    });
```
danschultz, line 407: Use `isStreamStarted` to check if the stream is connected and viewing a model.

packages/viewer/src/config/config.ts
jdm717: Makes sense, I'll update 👍

danschultz (approved):

PR #90, opened 2020-12-29
[SDK-1279] Defaulting to Pointer Events if supported

MadisonEhlers-Vertex:
## What

Regression test the web sdk on touch and mouse events. 

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1279

## Test Plan

Test pan/zoom, pinch, tap, etc.


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -228,13 +234,13 @@ export class Viewer {
...
+      this.usePointerEvents ? 'canvas-container-touch' : ''
+    }`;
     return (
       <Host>
         <div class="viewer-container">
-          <div
-            ref={ref => (this.containerElement = ref)}
-            class="canvas-container"
-          >
+          <div ref={ref => (this.containerElement = ref)} class={canvasClass}>
```
danconner-vertex, line 243: appears `classnames` is a dependency, but is not used, could that be used here?

packages/viewer/src/interactions/mouseInteractionHandler.ts
```diff
@@ -31,18 +37,22 @@ export class MouseInteractionHandler implements InteractionHandler {
   private primaryInteractionTypeChange = new EventDispatcher<void>();
 
   public constructor(
+    private usePointerEvents: boolean,
```
danconner-vertex, line 40: feels to me like this would be more readable as a property on `InteractionHandler`, or a new class for interaction handlers that support whatever the common theme of interactions are here.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -201,12 +203,16 @@ export class Viewer {
     }
 
     if (this.cameraControls) {
-      this.registerInteractionHandler(new MouseInteractionHandler());
-      this.registerInteractionHandler(new TouchInteractionHandler());
+      if (!this.usePointerEvents) {
+        this.registerInteractionHandler(new TouchInteractionHandler());
+      }
```
danconner-vertex, line 208: do 'pointer' and 'touch' events conflict in ways that 'touch' and 'mouse' events don't?

packages/viewer/src/interactions/interactionEvent.ts
```diff
@@ -41,3 +41,23 @@ export interface ZoomInteractionEvent extends InteractionEvent {
...
+  usePointerEvents: boolean
+): 'pointermove' | 'mousemove' => {
+  return usePointerEvents ? 'pointermove' : 'mousemove';
+};
+
+export const getUpEvent = (
+  usePointerEvents: boolean
+): 'pointerup' | 'mouseup' => {
+  return usePointerEvents ? 'pointerup' : 'mouseup';
+};
```
danconner-vertex, line 63: these could also be properties on `InteractionHandler` or a new class.

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: Yeah, if pointer events are supported, there's no need to capture touch events.

packages/viewer/src/interactions/mouseInteractionHandler.ts
MadisonEhlers-Vertex: Sounds good!

danschultz: +1. Instead of using conditionals, another option is to have a base class that has the interaction implementation, and delegates to subclasses for registering events. For example:

```ts
abstract class BaseInteractionHandler implements InteractionHandler {
  protected abstract registerEvents(element: HTMLElement): void;
  protected abstract deregisterEvent(): void;

  public initialize(...): void {
    this.registerEvents(element);
  }
}

class PointerInteractionHandler extends BaseInteractionHandler {
  protected registerEvents(element: HTMLElement): void {
    element.addEventListener('pointerdown', this.handleDownEvent);
  }
}
```

MadisonEhlers-Vertex (commented):
Upon doing some further testing, I'm going to need to handle the pinch event in a unique way for pointer events.

packages/viewer/src/components/viewer/viewer.css
```diff
@@ -27,6 +27,9 @@
   position: relative;
   background: var(--image-background, var(--viewer-background, #ffffff));
 }
+.canvas-container-touch {
```
danschultz, line 30: Consider renaming to something like `enable-pointer-events`.

packages/viewer/src/interactions/baseInteractionHandler.ts
```diff
@@ -0,0 +1,203 @@
...
+  private isDragging = false;
+
+  private computedBodyStyle?: CSSStyleDeclaration;
+
+  private primaryInteractionTypeChange = new EventDispatcher<void>();
+
+  public constructor(
+    protected downEvent: 'mousedown' | 'pointerdown',
+    protected upEvent: 'mouseup' | 'pointerup',
+    protected moveEvent: 'mousemove' | 'pointermove',
```
danschultz, line 39: If you do this, then there's probably not a need for `register` and `deregister` events. I'd say either remove those abstract methods, or remove these fields.

packages/viewer/src/interactions/pointerTapInteractionHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ConfigProvider } from '../config/config';
+
+export class PointerTapInteractionHandler extends BaseTapInteractionHandler {
+  public constructor(getConfig: ConfigProvider) {
+    super('pointerdown', 'pointerup', 'pointermove', getConfig);
+  }
+
+  protected registerEvents(element: HTMLElement): void {
+    element.addEventListener(this.downEvent, this.handleDown);
+    element.addEventListener('touchstart', this.handleTouchStart);
```
danschultz, line 11: Is `touchstart` needed?

packages/viewer/src/interactions/pointerTapInteractionHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  protected deregisterEvents(): void {
+    this.element?.removeEventListener(this.downEvent, this.handleDown);
+    this.element?.removeEventListener('touchstart', this.handleTouchStart);
+    this.element = undefined;
+
+    this.clearDoubleTapTimer();
+    this.clearLongPressTimer();
```
danschultz, line 20: I'd move things that are common to all sub-classes into the base class.

packages/viewer/src/interactions/pointerInteractionHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.downPosition = Point.create(event.screenX, event.screenY);
+    this.touchPoints = {
+      ...this.touchPoints,
+      [event.pointerId]: this.downPosition,
+    };
+
+    const keys = Object.keys(this.touchPoints);
+    if (keys.length === 2) {
+      this.disableIndividualInteractions = true;
+    }
```
danschultz, line 44: I think we can use an ES6 [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead, that just tracks the pointer IDs. We're not using the positions for anything.

packages/viewer/src/interactions/pointerInteractionHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+    window.addEventListener('pointerup', this.handlePointerUp);
+  }
+
+  private handlePointerUp(event: PointerEvent): void {
+    this.interactionApi?.endInteraction();
+    this.touchPoints = {};
+    this.disableIndividualInteractions = false;
+
+    window.removeEventListener('pointerup', this.handlePointerUp);
```
danschultz, line 53: Because of how function references work, I think there's a bug here with removing the event handler. You only want to remove the event listener for the last pointer event.

packages/viewer/src/interactions/pointerInteractionHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (keys.length === 2) {
+      this.disableIndividualInteractions = true;
+    }
+    window.addEventListener('pointerup', this.handlePointerUp);
+  }
+
+  private handlePointerUp(event: PointerEvent): void {
+    this.interactionApi?.endInteraction();
+    this.touchPoints = {};
+    this.disableIndividualInteractions = false;
```
danschultz, line 51: I suspect there's a bug here with `disableIndividualInteractions` getting set to false after a third pointer dispatches its up event. I'd recommend only setting `disableIndividualInteractions` only if the number of pointer ids in the set is 1.

packages/viewer/src/interactions/pointerInteractionHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.touchPoints = {
+      ...this.touchPoints,
+      [event.pointerId]: this.downPosition,
+    };
+
+    const keys = Object.keys(this.touchPoints);
+    if (keys.length === 2) {
+      this.disableIndividualInteractions = true;
+    }
+    window.addEventListener('pointerup', this.handlePointerUp);
```
danschultz, line 45: We probably only want to set add the event listener for the first pointer down.

packages/viewer/src/interactions/multiPointerInteractionHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+  private handlePointerDown(event: PointerEvent): void {
+    event.preventDefault();
+    const point = Point.create(event.screenX, event.screenY);
+    this.touchPoints = {
+      ...this.touchPoints,
+      [event.pointerId]: point,
+    };
+
+    window.addEventListener('pointermove', this.handlePointerMove);
+    window.addEventListener('pointerup', this.handlePointerUp);
```
danschultz, line 35: I suspect there's a similar bug with adding/removing event listeners. Should add checks so we only register event listeners on the first pointer down, and deregister on the last point up.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -201,14 +206,30 @@ export class Viewer {
     }
 
     if (this.cameraControls) {
-      this.registerInteractionHandler(new MouseInteractionHandler());
-      this.registerInteractionHandler(new TouchInteractionHandler());
+      // default to pointer events if allowed by browser.
+      if (this.usePointerEvents) {
```
danschultz, line 210: Could probably just inline `window.PointerEvent != null` here. The additional property isn't adding any value.

packages/viewer/src/interactions/tapInteractionHandler.ts
```diff
@@ -11,7 +12,7 @@ type TapEmitter = (
 ) => Promise<void> | void;
 
 export class TapInteractionHandler implements InteractionHandler {
-  private element?: HTMLElement;
+  protected element?: HTMLElement;
```
danschultz, line 15: This can still be private.

packages/viewer/src/interactions/tapInteractionHandler.ts
```diff
@@ -199,7 +204,7 @@ export class TapInteractionHandler implements InteractionHandler {
     this.clearLongPressTimer();
   }
 
-  private clearDoubleTapTimer(): void {
+  protected clearDoubleTapTimer(): void {
```
danschultz, line 207: ^

packages/viewer/src/interactions/tapInteractionHandler.ts
```diff
@@ -216,7 +221,7 @@ export class TapInteractionHandler implements InteractionHandler {
     );
   }
 
-  private clearLongPressTimer(): void {
+  protected clearLongPressTimer(): void {
```
danschultz, line 224: ^

packages/viewer/src/interactions/tapInteractionHandler.ts
```diff
@@ -107,10 +93,29 @@ export class TapInteractionHandler implements InteractionHandler {
     this.handlePointerEnd(this.pointerDownPosition);
   }
 
-  private handleMouseUp(event: MouseEvent): void {
+  protected handleDown(event: BaseEvent): void {
```
danschultz, line 96: ^

packages/viewer/src/interactions/pointerInteractionHandler.ts
MadisonEhlers-Vertex: I believe its fine, because any touch event that was triggered will set up this window listener, and it will only be removed if all of them are removed.

packages/viewer/src/interactions/pointerInteractionHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+      window.addEventListener('pointerup', this.handlePointerUp);
+    }
+
+    if (this.touchPoints.size === 2) {
+      this.disableIndividualInteractions = true;
+    }
+  }
+
+  private handlePointerUp(event: PointerEvent): void {
+    this.interactionApi?.endInteraction();
```
danschultz, line 48: I don't think this should be here. The parent class should handle this already.

packages/viewer/src/interactions/multiPointerInteractionHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+      const point2 = this.touchPoints[keys[1]];
+      this.handleTwoPointTouchMove(point1, point2);
+    }
+  }
+
+  private handlePointerUp(event: PointerEvent): void {
+    this.interactionApi?.endInteraction();
+    delete this.touchPoints[event.pointerId];
+    this.currentPosition1 = undefined;
+    this.currentPosition2 = undefined;
```
danschultz, line 62: I think we only want to end the interaction and clear the positions if the keys length is 1.

danschultz (approved):

PR #97, opened 2021-01-12
[SDK-1025] Update reference to wait for custom element to be defined

MadisonEhlers-Vertex:
## What

In some browsers the custom elements are not yet defined. We will want to wait for those prior to trying to access any custom elements on our web component. 

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1025

## Test Plan

Ensure the regression steps on SDK-1025 do not happen any more with this change. Note, you'll need to update the html in the ticket with the code referenced in the example to get this to work. 

N/A

N/a

## Dependencies

N/A

### DIFF
--- a/packages/viewer/readme.md
+++ b/packages/viewer/readme.md
@@ -43,6 +43,28 @@ file that references our published JS bundles from a CDN.
 </html>
 ```
 
+If you want to interact with the web component via JavaScript, you'll need to ensure the browser has registered the custom elements prior to use. 
+
+One option is to call this prior to any method calls on the viewer:
+```js
+  await window.customElements.whenDefined('vertex-viewer');
+```
+
+Another option is to import `defineCustomElements`, and this will register all custom elements in your DOM.
+```js
+import { defineCustomElements } from 'https://unpkg.com/@vertexvis/viewer@latest/dist/esm/loader.mjs';
+
+async function main() {
+  const viewer = document.querySelector('#viewer');
+  viewer.load(...)
+}
+
+window.addEventListener('DOMContentLoaded', () => {
+  defineCustomElements(window).then(() => main());
+});
+
+```
+
 ### NPM Dependency
 
 Our components can also be installed as an NPM dependency and imported through a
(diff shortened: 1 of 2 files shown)


packages/viewer/src/index.html
```diff
@@ -28,7 +28,8 @@
     </style>
 
     <script type="module">
-      window.addEventListener('DOMContentLoaded', () => {
+      window.addEventListener('DOMContentLoaded', async () => {
+        await window.customElements.whenDefined('vertex-viewer');
```
danschultz, line 32: I think it'd better communicate that we need to wait for the component to be hydrated if this was moved to `main`, since we're querying for the element there.

MadisonEhlers-Vertex: That sounds good. If we go with this route, I will move it.

Going to look into using `defineCustomElements` instead

packages/viewer/readme.md
```diff
@@ -43,6 +43,28 @@ file that references our published JS bundles from a CDN.
 </html>
 ```
 
+If you want to interact with the web component via JavaScript, you'll need to ensure the browser has registered the custom elements prior to use. 
+
+One option is to call this prior to any method calls on the viewer:
+```js
+  await window.customElements.whenDefined('vertex-viewer');
+```
```
danschultz, line 51: What do you think of just documenting `defineCustomElements`? It looks to be the documented/recommended approach in Stencil.

packages/viewer/readme.md
```diff
@@ -43,6 +43,28 @@ file that references our published JS bundles from a CDN.
...
+
+Another option is to import `defineCustomElements`, and this will register all custom elements in your DOM.
+```js
+import { defineCustomElements } from 'https://unpkg.com/@vertexvis/viewer@latest/dist/esm/loader.mjs';
+
+async function main() {
+  const viewer = document.querySelector('#viewer');
+  viewer.load(...)
```
danschultz, line 59: Consider using an unabbreviated value, e.g. `"urn:vertexvis:stream-key:123"`. Reduces some confusion on what is expected.

packages/viewer/readme.md
MadisonEhlers-Vertex: I'm ok with this. I do think it will be weird if we don't document it though, but then we use it in our sample index.html file for local dev.

danschultz (approved):

PR #103, opened 2021-01-13
[SDK-876] Support canary releases

danschultz:
## What

Adds support for automatically releasing a versions of the SDK when changes are merged to the main branch. Once published, you'll be able to consume the canary version by running `yarn add @vertexvis/viewer@canary`.

The release process has changed. There's no longer a need to manually enter a release version. There's a new field in `nextVersionBump` in `package.json` that tracks if the next version should bump the major, minor or patch level. Developers should run `yarn version:bump` if making a breaking change.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-876


.github/workflows/master.yml
```diff
@@ -38,23 +38,62 @@ jobs:
...
-        id: "detect_publish"
+        id: "detect-publish"
         run: |
-          result=`./scripts/detect_publish.sh`
+          result=`./scripts/detect_release.sh`
           echo $result
           echo ::set-output name=publish::$result
-  publish:
+
+  publish-canary:
```
jdm717, line 56: Will this job run for each PR merged including release PRs? Just wondering if this job should depend on `detect-publish-release`, and have a `if: needs.detect-publish-release.outputs.publish != 1` conditional

jdm717 (approved):
LGTM, had one question around when canary releases get published

.github/workflows/master.yml
danschultz: Yes, that's the intended behavior. When a release is run, I wanted to have the next canary also published that shadows the released version. This way if you're on a canary tag, and do an update, you'll effectively get the latest published version.

PR #109, opened 2021-01-14
[SDK-1249] Expose the scene view id from the viewer

danschultz:
## What

There wasn't a way to get the loaded scene view id programmatically in the viewer. This adds that + some logging about connections.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1249

### DIFF
--- a/packages/viewer/src/scenes/scene.ts
+++ b/packages/viewer/src/scenes/scene.ts
@@ -112,8 +112,7 @@ export class Scene {
     private stream: StreamApi,
     private renderer: RemoteRenderer,
     private frame: Frame.Frame,
-    private commands: CommandRegistry,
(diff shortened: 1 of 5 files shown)


packages/viewer/src/scenes/scene.ts
```diff
@@ -112,8 +111,7 @@ export class Scene {
     private stream: StreamApi,
     private renderer: RemoteRenderer,
     private frame: Frame.Frame,
-    private commands: CommandRegistry,
-    private sceneViewId: UUID.UUID
+    public readonly sceneViewId: UUID.UUID
```
danschultz, line 114: Actual change.

jdm717 (approved):
LGTM

PR #110, opened 2021-01-20
[SDK-1250] Add transform support

jdm717:
## What

Adds support for transforms through the SDK.

```javascript

// as an object
await scene
  .items((op) => [
    op
      .where((q) => q.withItemId('id'))
      .transform({
        r0: { x: 1, y: 0, z: 0, w: 0 },
        r1: { x: 0, y: 1, z: 0, w: 0 },
        r2: { x: 0, y: 0, z: 1, w: 0 },
        r3: { x: 0, y: 0, z: 0, w: 1 },
      }),
  ])
  .execute();

// as an array
await scene
  .items((op) => [
    op
      .where((q) => q.withItemId('id'))
      .transform([
        1, 0, 0, 0,
        0, 1, 0, 0,
        0, 0, 1, 0,
        0, 0, 0, 1
      ]),
  ])
  .execute();
```

Additionally, this adds a `dimensionschange` event to the viewer to support finding the dimensions of the underlying canvas in the viewer, rather than the container div.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1250

## Test Plan

- Test performing transforms through the SDK
- Test adding a `dimensionschange` listener and resizing the browser

N/A

As mentioned, a `dimensionschange` event has been added.

## Dependencies

N/A


packages/viewer/src/scenes/scene.ts
```diff
@@ -62,6 +64,53 @@ export class SceneItemOperationsBuilder
     );
   }
 
+  public transform(
+    matrix: vertexvis.protobuf.core.IMatrix4x4f | number[]
```
danconner-vertex, line 68: what are the cases for allowing either option for specifying the matrix? does something provide the transform in that way where we want to just pass through without modifying?

danconner-vertex (commented):
this looks pretty reasonable to me, just one question.

packages/viewer/src/scenes/scene.ts
danschultz: I think we should keep the API client more closely in line with what's expected in the protos to minimize additional logic, and move this conversion step a level higher into the viewer operations.

jdm717: Yeah, I suppose we don't necessarily have to support the `number[]` option here, my thinking was to have a slightly easier option for specifying this transform when using the SDK over CDN. Somewhat similar to the change we made to `materialOverride` to support a hex string (though not quite as useful here)

danschultz (approved):

PR #111, opened 2021-01-21
[SDK-1306] Exposing JWT on the viewer via emitter & accessor method

MadisonEhlers-Vertex:
## What

These changes introduce a way for consumers to get a current jwt. On the viewer instance, you can call `getJwt()`, like this:
```
   const jwt = await viewer.getJwt();
```

The jwt refreshes on every single stream reconnection, and updates the value in memory. You can also subscribe to this by adding an event listener:

```

    viewer.addEventListener('connectionChange', async (test) => {
       console.log(test.detail); // detail will be the new jwt with the connection status
       // OR
       await viewer.getJwt()
    });
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1306

## Test Plan

calling `await viewer.getJwt()` should always return an up to date jwt, and subscribing to it should allow consumers to hook into the event to always get an up to date jwt:
```
    viewer.addEventListener('connectionChange', async (test) => {
       console.log(test.detail); // detail will be the new jwt with the connection status
       // OR
       await viewer.getJwt()
    });
```

N/A

N/A

## Dependencies

N/a


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -621,14 +635,18 @@ export class Viewer {
...
         streamId: { hex: streamId },
         dimensions: this.dimensions,
         frameBackgroundColor: this.getBackgroundColor(),
         streamAttributes: this.getStreamAttributes(),
       });
       this.isStreamStarted = true;
       this.isReconnecting = false;
+
+      this.jwt = result.reconnect?.jwt || undefined;
+      this.jwtRefresh.emit(this.jwt);
```
danconner-vertex, line 648: if these are only valid for 5 minutes, are there updates on more messages that aren't start or reconnect?

is it guaranteed it is different than the previous jwt on start and reconnect messages? we wouldn't want to emit the event if it was the same

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -386,6 +392,11 @@ export class Viewer {
     return this.interactionHandlers;
   }
 
+  @Method()
+  public async getJwt(): Promise<string | undefined> {
+    return this.jwt;
+  }
```
danconner-vertex, line 398: is this jwt specifically for STS? might be helpful to specify that in the var/method name.

therockstorm: It is right now, but that might not always be the case.

packages/viewer/src/components/viewer/viewer.tsx
therockstorm: It's a new token on each reconnect. How often do reconnects happen?

danschultz: Reconnects should happen ~ every 1 minutes.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -163,6 +163,11 @@ export class Viewer {
    */
   @Event() public tokenExpired!: EventEmitter<void>;
 
+  /**
+   * Emits an event when the jwt for a session gets updated
+   */
+  @Event() public jwtRefresh!: EventEmitter<string>;
```
danschultz, line 169: How about we instead have a more generic "connected" event with an event payload that we can add additional info to later? I could see developers wanting to know when the SDK is in different connections states, e.g. `connecting | connected | connection_failed | disconnected`

packages/viewer/src/components/viewer/viewer.tsx
MadisonEhlers-Vertex: See Dan Shultz' comment. I changed the event to be a connectionChanged event and it has a jwt on connect/reconnect events.

therockstorm:
LGTM, but will yield to @danschultz

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -163,6 +168,11 @@ export class Viewer {
    */
   @Event() public tokenExpired!: EventEmitter<void>;
 
+  /**
+   * Emits an event when the connection status changes for the viewer
+   */
+  @Event() public connectionChange!: EventEmitter<ConnectionStatus>;
```
danconner-vertex, line 174: curious to get @danschultz 's thoughts on this as well, I think I like the interface of having specific events for the different connection statuses vs a single event where I need to check the status in the handler to handle the connection status I'm looking for. but probably comes down to whether we expect developers to more frequently want to run code whenever the connection status changes or more often when the connection status is a certain value.

MadisonEhlers-Vertex: My personal opinion if I were a consumer would be to subscribe to one location.

danschultz: If we're to do this, then we should turn `ConnectionStatus` into an intersection type. Some fields don't make any sense in a `connecting` or `disconnected` state, and having these additional types provides type safety within the context of our TS developers. Example:

```ts
interface ConnectedStatus {
  jwt: string;
  status: 'connected';
}

interface ConnectingStatus {
  status: 'connecting';
}

interface DisconnectedStatus {
  status: 'disconnected';
}

type ConnectionStatus = ConnectingStatus | ConnectedStatus | DisconnectedStatus;
```

danschultz: > but probably comes down to whether we expect developers to more frequently want to run code whenever the connection status changes or more often when the connection status is a certain value.

I think there'll be consumers who'll want to know the `connected` and `disconnected` state in a pair. Use-case would be to show some status to the user if the viewer is connected or not. FWIW, the browser's connection status APIs use a single event. https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API

danschultz (approved):

PR #112, opened 2021-01-23
[SDK-1313] Dispatch event when scene is ready to be interacted with

danschultz:
## What

Emit an event when the scene is ready for interaction.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1313


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -187,6 +186,11 @@ export class Viewer {
    */
   @Event() public connectionChange!: EventEmitter<ConnectionStatus>;
 
+  /**
+   * Emits an event when the scene is ready to be interacted with.
+   */
+  @Event() public sceneReady!: EventEmitter<void>;
```
jdm717, line 192: More of a general thought on the naming of these events, but we've got a bit of a mixture of camel-case and all lowercase currently (`connectionChange`/`sceneReady` versus `longpress`/`dimensionschange`). Wondering if we may want to revisit/define the pattern we use for these? Certainly not something to hold up the PR, but it feels a bit inconsistent right now

jdm717 (approved):
LGTM

packages/viewer/src/components/viewer/viewer.tsx
danschultz: Agreed. I'm sort of leaning towards use of camel case, because of how these get converted in Stencil's bindings. For example, with the React bindings, you'd get `onSceneReady` vs `onSceneready`.

PR #113, opened 2021-01-27
[SDK-1311] Support selection operations

jdm717:
## What

Adds support for selection operations.

```

scene.items(op => 
  op.where(q => q.withItemId('id').select('#ff0000'))
);

scene.items(op => 
  op.where(q => q.withItemId('id').deselect())
);
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1311

## Test Plan

- Test applying selection through the SDK
- Test deselecting the same item

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/scenes/operations.ts
+++ b/packages/viewer/src/scenes/operations.ts
@@ -9,6 +9,15 @@ interface HideItemOperation {
   type: 'hide';
 }
 
+interface SelectItemOperation {
+  type: 'select';
+  color: ColorMaterial;
+}
+
+interface DeselectItemOperation {
+  type: 'deselect';
+}
+
 interface ClearItemOperation {
   type: 'clear-override';
 }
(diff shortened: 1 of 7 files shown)


packages/viewer/src/scenes/operations.ts
```diff
@@ -74,6 +87,18 @@ export class SceneOperationBuilder
     );
   }
 
+  public select(color: ColorMaterial): SceneOperationBuilder {
```
danschultz, line 90: I think we can do this in a follow up, but I think we should have a way to set the default selection color on the viewer, then making this arg optional.

danschultz (approved):

packages/viewer/src/scenes/operations.ts
jdm717: Yeah, that makes sense to me. It's definitely a bit messy currently to try to manage that color across all cases of selection when consuming the viewer. I'll merge this change for now and follow-up with a change to support that

PR #114, opened 2021-02-02
[SDK-1332] Support bounding box on hits

jdm717:
## What

Adds support to include bounding boxes with hit results.

```
const resultsWithBounds = await scene.raycaster().hitItems(event.detail.position, ['bounding-box']);
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1332

## Test Plan

- Test retrieving a bounding box for a hit item properly returns a bounding box
- Test retrieving hit items without passing `['bounding-box']` for the `include` property
  - Should return undefined for all hit items' bounding boxes

Hit requests

N/A

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/144

### DIFF
--- a/packages/viewer/src/scenes/raycaster.ts
+++ b/packages/viewer/src/scenes/raycaster.ts
@@ -3,6 +3,11 @@ import { Point } from '@vertexvis/geometry';
 import { vertexvis } from '@vertexvis/frame-streaming-protos';
 
 /**
+ * Additional fields that can be included with hit results.
+ */
+export type HitResultInclude = 'bounding-box';
+
+/**
  * The `Raycaster` class is here.
  */
 export class Raycaster {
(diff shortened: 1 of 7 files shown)


packages/viewer/src/scenes/raycaster.ts
```diff
@@ -13,11 +18,20 @@ export class Raycaster {
...
   public async hitItems(
-    point: Point.Point
+    point: Point.Point,
+    include?: HitResultInclude[]
   ): Promise<vertexvis.protobuf.stream.IHitItemsResult | undefined> {
-    const res = await this.stream.hitItems({ point }, true);
+    const res = await this.stream.hitItems(
+      {
+        point,
+        includeBoundingBox: include && include.includes('bounding-box'),
```
danconner-vertex, line 31: could an interface to FSS here have been to pass the array of 'includes' to fetch. then this logic to indicate specifically the bounding box should be included could exist in FSS? are there advantages to making the payload to FSS some number of keys on what to include?

I'm not confident either way on this, just a question on the decision.

danschultz: You might be able to use protobuf enums to make this typesafe.

jdm717: Yeah, we certainly could still go that route if we wanted to change this. Currently the payload is mainly structured this way to follow similar structuring to the protos we have for the scene-service protos. I'd certainly be open to changing that though if FSS seems like the better place to do the mapping to those scene-service protos

jdm717:
Closing this PR for now. Created https://vertexvis.atlassian.net/browse/SDK-1354 to capture this work, and split it out since it appears to be a larger effort.

PR #116, opened 2021-02-08
[SDK-1283] Command/Control + click to fly to part

jdm717:
## What

Adds support for `KeyInteraction`s which can be registered for `tap` events (currently) which contain a predicate that returns true/false based on specific keys being pressed, along with a function to execute if the predicate returns true. The viewer will map over these `KeyInteraction`s whenever a `tap` event occurs, and invoke the functions for any where the predicate passes.

A `FlyToPartKeyInteraction` utilizing this `KeyInteraction` flow has also been added that will fetch hit results for and `flyTo` a part when the `Control` or `Meta` key are pressed.

Additionally, a couple configuration options have been added with these changes to allow customization of this behavior. A `keyboard-controls` prop (default: true) has been added to the viewer which can be set to `false` to prevent key state tracking and keyboard interactions. Along with that, a `viewer.config.animation` configuration has been added to allow configuration of the default animation duration for this `flyTo` behavior.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1283

## Test Plan

- Test initializing a viewer without specifying `keyboard-controls`
  - Holding Command/Control and clicking on a part should fly to that part
- Test initializing a viewer specifying `keyboard-controls="false"`
  - Holding Command/Control and clicking a part should not fly to that part
- Test registering a custom `KeyInteraction` by passing it to `viewer.registerTapKeyInteraction`
  - The function should be invoked when the predicate passes and a tap event occurs
  - See this example for a simple way to test this:
```javascript
class TestKeyInteraction {
  predicate(keyState) {
    return keyState['a'];
  }

  fn(event) {
    console.log(event);
  }
}

// Holding the 'a' key and clicking should log the event
viewer.registerTapKeyInteraction(new TestKeyInteraction());
```

N/A

N/A

## Dependencies

N/A


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -136,6 +140,13 @@ export class Viewer {
    */
   @Prop() public cameraControls = true;
 
+  /**
+   * Enables or disables the default keyboard shortcut interactions provided by
+   * the viewer. Enabled by default, requires `cameraControls` being enabled.
+   *
+   */
+  @Prop() public keyboardControls = true;
```
MadisonEhlers-Vertex, line 148: Wondering if we'll want a customizable payload here to specify which ones you might want? Just a thought. Or do we want just a kill switch to shut them all off?

jdm717: Yeah, that's a good question. These changes follow the pattern we have for camera controls, but maybe we'd want more control over these. Trying to think through a way to avoid having to add another "Config" type prop, curious what you would think about something like:
```typescript
type KeyboardControls = 'default' | 'custom' | 'none';
```
and then exporting the individual `KeyInteraction`s, so in the case that you want some interactions but not others, you could do something like:
```typescript
<vertex-viewer keyboard-controls="custom" />

viewer.registerTapKeyInteraction(
  new FlyToPartKeyInteraction(),
  new ShiftSelectKeyInteraction(),
  // etc
)
```
might require a bit of refactoring, but that would allow customization of the interactions without passing an object description of key interactions, and without having to register the key state interaction handler

MadisonEhlers-Vertex: This is likely something we could defer and prioritize when we add additional controls. Probably nothing needed now for this.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -267,27 +280,41 @@ export class Viewer {
...
         // fallback to touch events and mouse events as a default
         this.registerInteractionHandler(new MouseInteractionHandler());
         this.registerInteractionHandler(new TouchInteractionHandler());
-        this.registerInteractionHandler(
-          new TapInteractionHandler('mousedown', 'mouseup', 'mousemove', () =>
-            this.getConfig()
-          )
+        this.registerInteractionHandler(tapInteractionHandler);
+        this.registerTapKeyInteraction(
+          new FlyToPartKeyInteraction(this.stream, () => this.getConfig())
```
MadisonEhlers-Vertex, line 317: does this need to be duplicated in both the `if & else`? 

I think this could live inside the `if (this.cameraControls) {...}` conditional.

packages/viewer/src/interactions/flyToPartKeyInteraction.ts
```diff
@@ -0,0 +1,37 @@
...
+      hitResult.hitItems?.hits != null &&
+      hitResult.hitItems.hits.length > 0
+    ) {
+      await this.stream.flyTo({
+        itemId: hitResult.hitItems.hits[0].itemId,
+        animation: {
+          duration: toProtoDuration(this.configProvider().animation.durationMs),
+        },
+      });
+    }
```
MadisonEhlers-Vertex, line 35: Since the user will be expecting an action here from the sdk given that this interaction is turned on, should we add a warning or log or something that no parts were found? Not necessary, just a thought.

packages/viewer/src/interactions/keyStateInteractionHandler.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  private keyStateChange = new EventDispatcher<Record<string, boolean>>();
+
+  public constructor() {
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleKeyUp = this.handleKeyUp.bind(this);
+  }
+
+  public initialize(): void {
+    window.addEventListener('keydown', this.handleKeyDown);
```
MadisonEhlers-Vertex, line 15: Any benefit here to using pointer events? I don't think it'd matter much if this is a specific key-binding event.

packages/viewer/src/components/viewer/viewer.tsx
jdm717: Yep, you're right -- this is leftover from some changes I made. I'll update this

packages/viewer/src/interactions/flyToPartKeyInteraction.ts
jdm717: I guess my thought on that would be that I'd expect a click in empty space to do nothing with this interaction, and the warning might get a bit noisy (granted I don't expect people would be Command+clicking in white space often). Maybe a debug log would serve the same purpose and communicate it's expected behavior?

packages/viewer/src/interactions/keyStateInteractionHandler.ts
jdm717: I don't think the keyboard would dispatch pointer events, and we'll have to use `keydown`/`keyup`, if that's what you were thinking of replacing here?

MadisonEhlers-Vertex: Haha, yeah. Disregard. 🤦  It's not a mouse down :D

MadisonEhlers-Vertex (approved):

PR #119, opened 2021-02-11
[SDK-1321] Support starting from scene view state

jdm717:
## What

Adds support for starting a stream using a scene view state. This is provided using a query param on the `src` urn:

```
urn:vertexvis:stream-key:<key>?scene-view-state=<state-id>
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1321

## Test Plan

- Test starting a stream with a scene view state id provided

Loading of stream keys

N/A

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/153

### DIFF
--- a/packages/viewer/src/types/loadableResource.ts
+++ b/packages/viewer/src/types/loadableResource.ts
@@ -9,7 +9,7 @@ export type LoadableResource = StreamKeyResource;
 
 export interface Resource {
   resource: LoadableResource;
-  query?: QueryResource;
+  queries?: QueryResource[];
 }
 
 export function fromUrn(urn: string): Resource {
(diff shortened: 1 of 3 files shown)


packages/viewer/src/types/loadableResource.ts
```diff
@@ -22,8 +27,31 @@ export function fromUrn(urn: string): LoadableResource {
...
+    const [resourceType, resourceId] = query.split('=');
+
+    switch (resourceType) {
+      case 'scene-view-state':
+        return { type: 'scene-view-state', id: resourceId };
+      default:
+        throw new Error('Invalid URN. Unknown query resource type');
+    }
+  }
+}
```
danschultz, line 57: Should we have something that's a bit more generic here? It's possible to have multiple query params.

jdm717: Yeah, I suppose we can add it here and just use the first query parameter during load (until we add support for starting from multiple scene-view-states if we want that in the future). I'll update this

danschultz (approved):

PR #120, opened 2021-02-12
[SDK-1123] viewer sdk docs

danconner-vertex:
## What

Generating viewer sdk api docs with typedoc

Looks like we'll need to figure out some options to identify the right format for the generated docs. Github Pages for the vertex-web-sdk repo is serving two versions of the docs.

there are docs generated for these packages in the sdk:
https://vertexvis.github.io/vertex-web-sdk/packages/viewer/index.html
https://vertexvis.github.io/vertex-web-sdk/packages/utils/index.html
https://vertexvis.github.io/vertex-web-sdk/packages/geometry/index.html
https://vertexvis.github.io/vertex-web-sdk/packages/stream-api/index.html

for now, these would be linked to from the dev portal's sidebar navigation

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1123


danschultz:
> do we want to generate these docs for every package in this repo?

I think we start with the packages that are commonly used with working with the SDK. This'd include `@vertexvis/geometry`, `@vertexvis/viewer`, `@vertexvis/utils`.

> the defaults side navigation for the package generated docs seems a bit funny. I'll be looking at what's possible to clean that up. not sure why JSX appears.

It does seem a bit over complicated. I kinda like the simplicity of ESDoc ([example](https://doc.esdoc.org/github.com/nimiq/core-js/)). They do have a [TS plugin](https://www.npmjs.com/package/esdoc-typescript-plugin) that maybe we can take a look at.

> still working out what the best process would be for generating and publishing them. was looking at the post commit hook or 'lint-staged' style approach, but feel like the command takes a bit too long for that, would be annoying.

What do you think of doing this via Github Actions after successful publish?

packages/viewer/src/components.d.ts
```diff
@@ -57,17 +57,20 @@ export namespace Components {
...
+          * @example ``` class CustomInteractionHandler extends InteractionHandler {   private element: HTMLElement;   private api: InteractionApi;    public dispose(): void {     this.element.removeEventListener('click', this.handleElementClick);   }    public initialize(element: HTMLElement, api: InteractionApi): void {     this.api = api;     this.element = element;     this.element.addEventListener('click', this.handleElementClick);   }    private handleElementClick = (event: MouseEvent) => {     api.tap({ x: event.clientX, y: event.clientY });   } }  const viewer = document.querySelector("vertex-viewer"); viewer.registerInteractionHandler(new CustomInteractionHandler); ```
```
danconner-vertex, line 60: the stencil build reformats this to a single line, removing newlines (from what I can tell so far). this results in not so great code blocks: https://vertexvis.github.io/vertex-web-sdk/packages/viewer/interfaces/viewer.components.vertexviewer.html#registerinteractionhandler

compared to: https://vertexvis.github.io/vertex-web-sdk/packages/viewer/interfaces/interactions.interactionhandler.html

not sure what we can do about this, same behavior exists in stencil 2

@danschultz

packages/viewer/src/commands/index.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * @module Commands
+ */
+export * from './commandRegistry';
+
+export * from './streamCommands';
```
danconner-vertex, line 6: added to define the module. without this, and including in the `typedoc.json`, the navigation ends up looking like the attached image. see comment on `src/index.ts` for more context.

Screen Shot 2021-02-22 at 8 49 33 AM

packages/viewer/src/interactions/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export * from './interactionApi';
+
+export * from './interactionEvent';
+
+export * from './interactionHandler';
+
+export * from './keyInteraction';
+
+export * from './keyStateInteractionHandler';
```
danconner-vertex, line 16: see https://github.com/Vertexvis/vertex-web-sdk/pull/120/files#r580383341

these should be all classes/interfaces used by the viewer class.

packages/viewer/src/scenes/index.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @module Scenes
+ */
 export * from './camera';
 
 export * from './raycaster';
```
danconner-vertex, line 6: see https://github.com/Vertexvis/vertex-web-sdk/pull/120/files#r580383341

packages/viewer/typedoc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "entryPoints": [
+    "./src/index.ts",
+    "./src/commands/index.ts",
+    "./src/interactions/index.ts",
+    "./src/scenes/index.ts",
+    "./src/types/index.ts"
```
danconner-vertex, line 7: including these as `entryPoints` tells typedoc to generate documentation for the classes/interfaces, then link to them in the viewer docs. see `Scene` here https://vertexvis.github.io/vertex-web-sdk/packages/viewer/interfaces/viewer.components.vertexviewer.html#scene

another approach would be to define these directories as modules and export them in `src/index.ts` as well, if it's valuable to keep a single entry point.

@danschultz

danschultz (commented):
Overall, I think this is a good start. There's something about the default settings of TypeDoc that IMO makes it difficult to grok the generated documentation. But, I think that it exposes enough knobs that we'll be able to configure it to our needs.

Some general configuration changes that I'd recommend making in this first pass:

* Can we rename the directory that the documentation gets output into? Instead of `packages/*` I propose `reference/*`.
* Can we use the `minimal` theme? Playing around with the output between `default` and `minimal`, I generally find the `minimal` setting easier to grok, as reduces one layer of navigation.

danschultz (commented):
Also, I'm wondering if it might be better to generate TypeDoc as a monorepo, vs generating docs individually for each project. By doing this, we might be able to link to types across packages. There's a bit of a disjoint right now because you won't be able to click through to the `Color` type from `ColorMaterial` because color lives in the utils package.

package.json
```diff
@@ -6,12 +6,13 @@
...
     "lerna": "^3.22.1",
     "semver": "^7.3.4",
+    "typedoc": "^0.20.25",
     "typescript": "^4.2.2"
   },
   "scripts": {
     "postinstall": "yarn generate:vscode-workspace && yarn generate:npmrc && yarn bootstrap",
     "generate:vscode-workspace": "./scripts/generate_vscode_workspace.sh",
-    "generate:docs": "lerna run generate:docs",
+    "generate:docs": "lerna run generate:docs && npx typedoc --tsconfig tsconfig.docs.json",
```
danschultz, line 15: You shouldn't need `npx` here.

danschultz (approved):

PR #121, opened 2021-02-15
[SDK-1162] Increase max image size

jdm717:
## What

Increases the maximum image size possible, and modifies the way we scale dimensions to use an area-based approach rather than fitting to a specific set of dimensions. Currently this is set to scale dimensions to have an area less than or equal to 2,073,600 pixels (1920x1080/1080p).

Additionally, a feature-flag to enable scaling images up to the host dimensions has been added, and can be enabled using:
```typescript
viewer.config {
  flags: {
    scaleFramesToHost: true
  }
}
```
This flag will cause all images to be scaled up to the host's dimensions, while still being bound to the pixel count listed above.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1162

## Test Plan

- Test rendering images in environments with less than the max pixel count
  - Images should show no letterboxing
- Test rendering images in environments with more than the max pixel count (`scaleFramesToHost: false`)
  - Images should be letterboxed uniformly around the image, and maintain the aspect ratio of the container
- Test rendering images in environments with more than the max pixel count (`scaleFramesToHost: true`)
  - Images should show no letterboxing
  - Images should show more visual artifacts due to being scaled up in size

Viewer sizing

N/A

## Dependencies

N/A


packages/geometry/src/rectangle.ts
```diff
@@ -87,6 +87,23 @@ export const cropFit = (to: Rectangle, rect: Rectangle): Rectangle => {
...
+ *
+ * @param to - the maximum area this rectangle can have
+ * @param rect - the rectangle to scale to fit the specified area
+ */
+export const scaleFit = (to: number, rect: Rectangle): Rectangle => {
+  const scale = Math.min(Math.sqrt(to / area(rect)), 1);
+  const dimensions = Dimensions.floor(
+    Dimensions.proportionalScale(scale, rect)
+  );
+  const position = Point.subtract(center(rect), Dimensions.center(dimensions));
```
danconner-vertex, line 103: is there potential for alternate behaviors for the resulting position after the `scaleFit`? like keeping one of the same points? or is the scaling down and moving the behavior we always want for now?

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -849,10 +854,13 @@ export class Viewer {
   private async handleFrame(
     payload: vertexvis.protobuf.stream.IDrawFramePayload
   ): Promise<void> {
-    if (this.canvasElement != null && this.dimensions != null) {
+    const dimensions = this.getConfig().flags.scaleFramesToHost
+      ? this.hostDimensions
+      : this.dimensions;
```
danconner-vertex, line 859: looks like this could be a useful private method

packages/geometry/src/rectangle.ts
jdm717: Good question. As far as I know right now the re-centering behavior is always what we want, but that's just based on the existing behavior also re-centering, so maybe it's worth revisiting that

danconner-vertex: sounds like something we can revisit if we need different behavior.

danconner-vertex (approved):

therockstorm (approved):

therockstorm (approved):

danschultz (approved):

PR #133, opened 2021-02-18
[SDK-1360] Support waiting for animations to complete

jdm717:
## What

Adds support for waiting for an animation to complete when performing a `flyTo`. When `.render()` is called, it returns a promise that resolves with a mapping of possible stream events to promises that can be awaited. Specifically in the case of animations, an `animationCompleted` promise will be available that can be awaited as shown here:

```

const events = await viewer.camera().flyTo({
  suppliedId: 'supplied-id'
}).render({
  animation: {
    milliseconds: 500
  }
});

await events.animationCompleted;
```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1360

## Test Plan

- Test fly to
- Test waiting for animations to complete after performing a fly to
- Test performing fly to without animations

Fly to/animations

N/A

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/165


packages/viewer/src/stream/events.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  return (id: string) => {
+    const promise = new Promise<void>(resolve => {
+      requests.set(id, () => {
+        resolve();
+      });
+    });
+
+    return Async.timeout(10 * 1000, promise).finally(() => requests.delete(id));
```
danschultz, line 27: This timeout will need to be a function of the animation duration.

packages/viewer/src/scenes/camera.ts
```diff
@@ -158,7 +160,9 @@ export class Camera implements FrameCamera.FrameCamera {
    * Queues the rendering for a new frame using this camera. The returned
    * promise will resolve when a frame is received that contains this camera.
    */
-  public async render(renderOptions?: CameraRenderOptions): Promise<void> {
+  public async render(
+    renderOptions?: CameraRenderOptions
+  ): Promise<StreamEvents | undefined> {
```
danschultz, line 165: I think we need to think about these events a bit differently and perhaps a bit more abstractly. It's possible that we add additional events in the future, and as a developer, I'd want to subscribe to certain ones. We actually have two types of events right now, one for the first frame being received and this new one for animation end. What if we had an API and convention like this:

```ts
const camera = (await viewer.scene()).camera();
const resp = await camera
  .update({ position: {x: 1, y: 1, z: 1 })
  .render({ animation: {milliseconds: 1000} }); // Completes when FSS sends a response.

// Any data that we want to expose from the FSS response.
console.log(resp.data);

// Invokes the callback once and removes the listener. Need to think about 
// how event listeners are easily cleaned up to prevent a memory leak.
resp.onAnimationCompleted.once(() => console.log('animation done'));

// Invokes the callback N times for each animationCompleted event.
const handler = () => console.log('animation done')
resp.onAnimationCompleted.on(handler);

// Remove the event listener.
resp.onAnimationCompleted.off(handler);
```

BTW, there's a shell for a `EventDispatcher` class that is we may be able to leverage.

packages/viewer/src/stream/result.ts
```diff
@@ -0,0 +1,5 @@
+import { StreamApi } from '@vertexvis/stream-api';
+
+export class Result<T = undefined> {
+  public constructor(public data: T, private stream: StreamApi) {}
+}
```
danschultz, line 5: I'd maybe make `stream` be part of subclasses to keep this more generic. Could also probably turn this into an interface.

packages/viewer/src/scenes/__tests__/camera.spec.ts
```diff
@@ -1,21 +1,30 @@
...
 
 describe(Camera, () => {
-  const renderer = jest.fn();
+  const stream = new StreamApi();
   const data = FrameCamera.create({ position: Vector3.create(1, 2, 3) });
   const boundingBox = BoundingBox.create(Vector3.create(), Vector3.create());
 
+  beforeAll(() => {
+    stream.flyTo = jest.fn(async () => ({ flyTo: {} }));
+    (toProtoDuration as any).mockImplementation(realApi.toProtoDuration);
```
danschultz, line 17: Lets add a lint ignore.

packages/viewer/src/stream/dispatcher.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public off(listener: Listener<T>): void {
+    this.listeners = this.listeners.filter((l) => l !== listener);
+    if (this.listeners.length === 0) {
+      this.removeListeners();
+    }
+  }
+
+  public once(listener?: Listener<T>): Promise<T> {
```
danschultz, line 37: What's the purpose of this listener? Can't the caller just do `.then(listener)`?

packages/viewer/src/scenes/renderResult.ts
```diff
@@ -0,0 +1,40 @@
...
+import { StreamApi } from '@vertexvis/stream-api';
+import { Result } from '../stream/result';
+import { StreamApiEventDispatcher } from '../stream/dispatcher';
+
+export interface RenderResultIds {
+  animationId?: string;
+  correlationId?: string;
+}
+
+export class RenderResult implements Result {
```
danschultz, line 11: nit, but how about we rename this to `CameraRenderResult` to prevent a future naming collision if we have another render method that has different events.

packages/viewer/src/scenes/renderResult.ts
```diff
@@ -0,0 +1,40 @@
...
+  public data = undefined;
+  public onAnimationCompleted: StreamApiEventDispatcher<string | undefined>;
+  public onFrameReceived: StreamApiEventDispatcher<Frame.Frame | undefined>;
```
danschultz, line 14: It'd be better if this was a type `StreamApiEventDispatcher<Frame.Frame>`. This way the caller doesn't have to do a null check.

packages/viewer/src/scenes/renderResult.ts
```diff
@@ -0,0 +1,40 @@
...
+import { StreamApiEventDispatcher } from '../stream/dispatcher';
+
+export interface RenderResultIds {
+  animationId?: string;
+  correlationId?: string;
+}
+
+export class RenderResult implements Result {
+  public data = undefined;
+  public onAnimationCompleted: StreamApiEventDispatcher<string | undefined>;
```
danschultz, line 13: ^

danschultz (commented):
Just some minor things. Overall LGTM!

packages/viewer/src/stream/dispatcher.ts
jdm717: Yeah, this probably isn't necessary -- I'll remove

danschultz (approved):

packages/viewer/src/scenes/cameraRenderResult.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface RenderResultIds {
+  animationId?: string;
+  correlationId?: string;
+}
+
+export class RenderResult implements Result {
+  public data = undefined;
+  public onAnimationCompleted: StreamApiEventDispatcher<string | undefined>;
+  public onFrameReceived: StreamApiEventDispatcher<Frame.Frame | undefined>;
```
danschultz, line 14: These signatures still have `undefined` in them.

packages/viewer/src/scenes/cameraRenderResult.ts
```diff
@@ -0,0 +1,40 @@
...
+import { StreamApi } from '@vertexvis/stream-api';
+import { Result } from '../stream/result';
+import { StreamApiEventDispatcher } from '../stream/dispatcher';
+
+export interface RenderResultIds {
+  animationId?: string;
+  correlationId?: string;
+}
+
+export class RenderResult implements Result {
```
danschultz, line 11: `CameraRenderResult`

packages/viewer/src/scenes/cameraRenderResult.ts
jdm717: Hmm, apparently I wound up with an outdated version of that file being pushed up. I'll correct that

danschultz (approved):

PR #136, opened 2021-02-23
[SDK-989] Add components to render toolbars in the Viewer SDK

danschultz:
## What

This adds components to add toolbars to the Viewer SDK. The toolbar contains a collection of components that a developer can use to build their own very custom toolbar, and a "default" toolbar that provides out of the box functionality for the viewer. These components consist of:

* `<vertex-viewer-toolbar>`: A component that performs the layout of elements in a toolbar. It's intended that a developer would supply `vertex-viewer-toolbar-group` elements, but isn't required. Exposes a property `position` to position the toolbar vertically and horizontally, and `direction` to flow elements vertically or horizontally.
* `<vertex-viewer-toolbar-group>`: A component that performs the layout of a group of items in a toolbar. It's intended that a developer provide a set of `vertex-viewer-button` elements, but isn't required. Exposes a property `direction` to flow elements vertically or horizontally.
* `<vertex-viewer-button>`: A dumb button. Currently this component acts as a CSS reset, nullifying any default styles that the browser applies to buttons. Maybe a question if this component is even needed.
* `<vertex-viewer-icon>`: A component containing the SVGs for our default icon set.
* `<vertex-viewer-default-toolbar>`: A component that provides a toolbar with some default behavior and styling (currently a fit-all button, with old styling from the POC SDK).

All the components except for `<vertex-viewer-default-toolbar>` don't have any out of the box styling, with the exception of the necessary CSS to layout their children. This means developers who want to use these lower-level components will need to provide their own styling to customize colors, padding, borders, etc.

I've also moved `./packages/viewer/src/index.html -> ./packages/viewer/index.html`. This was causing Stencil to not properly create separate bundles for each component. This may mean that livereload of `index.html` to not work.

Screen Shot 2021-02-23 at 3 36 55 PM

## Ticket

https://vertexvis.atlassian.net/browse/SDK-989


packages/viewer/src/components/viewer-button/viewer-button.tsx
```diff
@@ -0,0 +1,18 @@
+import { Component, h, Host } from '@stencil/core';
+
+@Component({
+  tag: 'vertex-viewer-button',
+  styleUrl: 'viewer-button.css',
+  shadow: true,
+})
+export class ViewerButton {
```
danschultz, line 8: I'm leaning towards not including this component for now.

danconner-vertex: I do see the usefulness though. hard to imagine a scenario where a toolbar button should have the default button styles.

packages/viewer/src/components/viewer-icon/viewer-icon.tsx
```diff
@@ -0,0 +1,67 @@
+import { Component, h, Host, Prop } from '@stencil/core';
+import classname from 'classnames';
+
+/**
+ * The names of the icons that the `<vertex-viewer-icon>` element can display.
+ */
+export type ViewerIconName = 'fit-all';
```
danconner-vertex, line 7: what are the advantages to a single icon component driven by a property? easier to manage a single file with a big if/switch statement over multiple files?

danconner-vertex (commented):
feels like we should start adding/updating guides with any new feature to the SDK. but since we already have work slotted for that this sprint, can easily be addressed there.

packages/viewer/src/components/viewer-button/viewer-button.tsx
danschultz: Ok, will keep it around. I think we'll need it in the future anyway for buttons that have richer logic, like selected buttons.

packages/viewer/src/components/viewer-icon/viewer-icon.tsx
danschultz: That, and there'll be a lot of boilerplate in Stencil to have separate components for each icon, e.g. each icon would need its own directory + tests.

danconner-vertex (approved):

PR #141, opened 2021-02-24
[SDK-1407] Adding Twist Operation

MadisonEhlers-Vertex:
## What

Adding a twist operation that defaults to shift+alt + rotate
 out of the box you now have twist bound to the keys listed above. You can turn it off by turning the keyboard controls off:
 
```
<vertex-viewer keyboard-controls="false" />
 ```

You can create your own key binding to interaction type like this:
```
export class TwistInteractionHandler implements KeyInteractionWithReset {
  public constructor(private baseInteractionHandler: BaseInteractionHandler) {}

  public predicate(keyState: KeyState): boolean {
    return keyState['Alt'] === true && keyState['Shift'] === true;
  }

  public async fn(): Promise<void> {
    this.baseInteractionHandler.setCurrentInteractionType('twist');
  }

  public async reset(): Promise<void> {
    this.baseInteractionHandler.setCurrentInteractionType(undefined);
  }
}

```

And then registering it with the viewer:
```

```
## Ticket

https://vertexvis.atlassian.net/browse/SDK-1407

## Test Plan

Test the twist is working with the key bindings

### CODE REVIEW

**Comment by danschultz** [comment]
Seeing some weird jumping when testing these changes.


https://user-images.githubusercontent.com/37429/109188826-f10e3680-7758-11eb-83c4-59e2acab1e9c.mov

File: packages/viewer/src/interactions/__tests__/mouseInteractionHandler.spec.ts [hunk 1]
```diff
@@ -164,6 +167,29 @@ describe(MouseInteractionHandler, () => {
...
+  //     buttons: 1,
+  //     bubbles: true,
+  //     shiftKey: true,
+  //     altKey: true,
+  //   });
+  //   div.dispatchEvent(mouseDown);
+  //   window.dispatchEvent(shiftAltPrimaryMove1);
+  //   window.dispatchEvent(shiftAltPrimaryMove2);
+  //   window.dispatchEvent(mouseUp);
+  // }
```
danschultz, line 191: Let's remove this if it's not needed.

packages/viewer/src/interactions/baseInteractionHandler.ts
```diff
@@ -80,6 +82,15 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
       case 'pan':
         this.primaryInteraction = this.panInteraction;
         break;
+      case 'twist':
+        this.primaryInteraction = this.twistInteraction;
+        break;
```
danschultz, line 87: In the future, this class will become responsible for driving the behavior for these UI controls. One thing it'll probably handle is emitting of events that the UI control will listen to visually update the selected tool. By having twist in here and doing things like resets, the UI control is going to flicker between selected tools.

Screen Shot 2021-02-25 at 11 07 22 AM

I don't think this is the right place to handle twist interactions based on keyboard modifiers.

MadisonEhlers-Vertex: What do you suggest then?

MadisonEhlers-Vertex: Just thinking here: I could create a specific key down handler that then adds pointer event listeners. I don't know how we would do that and allow for seamless interactions between interactions though. Which is why this lives in the base interaction handler:

one interaction

MadisonEhlers-Vertex:
@danschultz 
>Seeing some weird jumping when testing these changes.

Yeah, I saw it too. I pushed up a fix for this. Was not reseting the angle & position correctly. Should be fixed.

danschultz (approved):

PR #142, opened 2021-02-24
[SDK-1408] Distance threshold scaling

jdm717:
## What

Adds support for configuring the scaling of our interaction thresholds (rotate/pan/zoom/tap) through `viewer.config`. These changes also introduce some device-dependent scaling of those thresholds for [coarse](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer) pointer-based devices that will scale the threshold based on the device's pixel ratio. 

An example for this on an iPad (where touch events are considered coarse pointer events), is that the movement threshold of `>= 2 pixels` would become (base * pixelRatio * configValue) or (2 * 2 * 1.5), resulting in a threshold of `>= 6 pixels`.

Currently this is configurable for coarse (touch) pointer events and fine (mouse) pointer events:
```typescript

viewer.config = {
  interactions: {
    finePointerThresholdScale: 2, // default is 1
    coarsePointerThresholdScale: 5 // default is 1.5
  }
}

```

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1408

## Test Plan

- Test mouse interactions on various mouse and touch devices (should require more movement on touch devices)
- Test tap interactions on various mouse and touch devices (should tolerate more movement on touch devices)

Mouse/touch interactions

N/A

## Dependencies

N/A


packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -191,6 +195,33 @@ export class InteractionApi {
...
+    isTouch?: boolean
+  ): number {
+    if (this.isCoarseInputDevice(isTouch)) {
+      return (
+        pixelThreshold *
+        window.devicePixelRatio *
+        this.getConfig().coarsePointerThresholdScale
+      );
+    } else {
+      return pixelThreshold * this.getConfig().finePointerThresholdScale;
```
danschultz, line 221: I think we want to take into account the device pixel ratio even for fine pointers. My guess is that a stylus on a iPad or windows desktop would be categorized as a fine pointer device.

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -191,6 +195,33 @@ export class InteractionApi {
...
+   */
+  public scalePixelThreshold(
+    pixelThreshold: number,
+    isTouch?: boolean
+  ): number {
+    if (this.isCoarseInputDevice(isTouch)) {
+      return (
+        pixelThreshold *
+        window.devicePixelRatio *
+        this.getConfig().coarsePointerThresholdScale
```
danschultz, line 218: Thoughts on simplifying this to simply `course|fineThreshold * window.devicePixelRatio`? I think this additional scaling factor may be more difficult for external devs to understand vs a simpler setting that is just `pixel threshold * DPI`.

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -210,4 +241,8 @@ export class InteractionApi {
       shiftKey,
     });
   }
+
+  private isCoarseInputDevice(isTouch?: boolean): boolean {
+    return isTouch || window.matchMedia('(pointer: coarse)').matches;
```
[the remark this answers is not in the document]
jdm717: Yeah, I had no idea it existed prior to doing some research for this ticket. Definitely pretty cool! I also didn't expect it to have as much support as it does. https://caniuse.com/matchmedia (for the api) and https://caniuse.com/css-media-interaction (for the pointer types)

packages/viewer/src/interactions/interactionApi.ts
jdm717: I believe a stylus actually registers as a `coarse` input type, and from what I've read only a mouse or trackpad would register as a `fine` input type. Certainly not opposed to changing this, but it would mean that on macbooks with retina display we'd be scaling up the threshold (which maybe we want anyway)

packages/viewer/src/interactions/interactionApi.ts
jdm717: Yeah, I think that makes sense. I suppose in the case of a higher `window.devicePixelRatio` on a laptop/desktop, we still may want to increase the threshold, so I can update this

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -191,6 +195,27 @@ export class InteractionApi {
...
+   */
+  public scalePixelThreshold(
+    pixelThreshold: number,
+    isTouch?: boolean
+  ): number {
+    const scaleFactor = this.isCoarseInputDevice(isTouch)
+      ? this.getConfig().coarsePointerThresholdScale
+      : this.getConfig().finePointerThresholdScale;
+
+    return pixelThreshold * window.devicePixelRatio * scaleFactor;
```
danschultz, line 216: Sorry, to clarify, I think it'd be easier to understand if there was just a simple threshold for fine and course pointers and no scale factor. In other words, I can configure a threshold and and it scales by the devices DPI. Having an additional scale factor on top of that adds an extra layer of complexity and doesn't give me a way to simply adjust the pixel threshold.

jdm717: I see, I think in the ticket we specifically wanted the ability to configure mouse/touch thresholds separately. Are you thinking that'd no longer be needed?

danschultz: How about something like this?

```ts
    const pixelThreshold = this.isCoarseInputDevice(isTouch)
      ? this.getConfig().coarsePointerThreshold
      : this.getConfig().finePointerThreshold;

    return pixelThreshold * window.devicePixelRatio;
```

jdm717: So just to make sure I'm following, with this change, we'd essentially remove all the places where previously we had something like `Point.distance(first, second) >= 2` and replace those with `Point.distance(first, second) >= api. pixelThreshold()`? Essentially to get rid of the idea of "scaling" the thresholds we already have, and instead define your own?

danschultz: Yep

danschultz (approved):

PR #143, opened 2021-02-25
[SDK-1406] Enabling passing correlation IDs on scene alterations

MadisonEhlers-Vertex:
## What

These changes introduce a way for a client to specify what correlation ID they would want to set on a frame, as well as enabling them to prevent drawing a frame until a frame with that correlation id comes back.

Here is an example in platdev:

```
      <style>
             .viewer-hide {
               visibility: hidden;
             }
      </style>

        viewer.classList.add("viewer-hide");
        await viewer.load('urn:vertexvis:stream-key:YourStreamKey'); // DM for stream key
        const suppliedCorrelationId = 'madisons-62f4708c-77e7-11eb-9439-0242ac130002';
        await scene.items(op => [
            op.where(q => q.all()).hide(),
            op.where(q => q.withItemId("53a2c961-e630-4619-97bb-d50d321492c5")).show()
          ]).execute(
            {
            suppliedCorrelationId
          }
        );
        viewer.addEventListener('frameDrawn', (frame) => {
          if (frame.detail.correlationIds.includes(suppliedCorrelationId)) {
            viewer.classList.remove("viewer-hide");
          }
        });
```

Waiting for scene alteration

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1406

## Test Plan

See above. Should wait for a frame with a correlation id. 

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -239,6 +239,7 @@ export class Viewer {
   private streamSessionId?: UUID.UUID;
   private streamId?: UUID.UUID;
   private streamDisposable?: Disposable;
+  private firstFrameCorrelationId?: string;
   private jwt?: string;
   private isStreamStarted = false;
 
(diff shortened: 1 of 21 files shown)


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -562,6 +563,22 @@ export class Viewer {
...
 
+  /**
+   * Waits for a certain duration for a frame with a given correlation ID to be present on a frame.
+   * Default timeout for waiting is 2 seconds, but can be configured.
+   */
+  @Method()
+  public async setWaitForFrameWithCorrelationId(
+    suppliedCorrelationId: string,
+    waitTimeoutMs = 2000
+  ): Promise<void> {
```
danschultz, line 574: Developers can either use `onFrameDrawn` or `onFrameReceived` event handlers and check if the frame has a correlation ID. I don't think we need to expose anything additional for this.

danschultz (approved):

PR #145, opened 2021-03-01
[SDK-1429] Stencil 2.0 support for Edge <= 18

jdm717:
## What

Adds `extras` corresponding to necessary polyfills for Edge <= 18 to allow for usage in the Xamarin SDK.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1429

## Test Plan

- Test on Windows with EdgeHTML <= 18 replacing the `DOMContentLoaded` listener with the following
```typescript
import { applyPolyfills } from '/dist/esm/polyfills/index.js';
import { defineCustomElements } from '/dist/esm/loader.js';

window.addEventListener('DOMContentLoaded', () => {
  applyPolyfills().then(() => defineCustomElements(window).then(() => main()))
});
```

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/stencil.config.ts
+++ b/packages/viewer/stencil.config.ts
@@ -34,4 +34,9 @@ export const config: Config = {
     },
     roots: ['<rootDir>/src/'],
   },
+  extras: {
+    cssVarsShim: true,
+    dynamicImportShim: true,
+    shadowDomShim: true,
+  },
 };
--- a/packages/viewer/stencil.config.ts
+++ b/packages/viewer/stencil.config.ts
@@ -35,7 +35,6 @@ export const config: Config = {
     roots: ['<rootDir>/src/'],
   },
   extras: {
-    cssVarsShim: true,
     dynamicImportShim: true,
     shadowDomShim: true,
   },


jdm717:
One note here, this doesn't restore all the defaults for Stencil 1, only those for EdgeHTML <= 18. See the [changelog](https://github.com/ionic-team/stencil/blob/master/CHANGELOG.md#restore-stencil-1-defaults) for more info on restoring those defaults. Curious if anyone has thoughts on whether we should restore all those defaults?

packages/viewer/stencil.config.ts
```diff
@@ -34,4 +34,9 @@ export const config: Config = {
     },
     roots: ['<rootDir>/src/'],
   },
+  extras: {
+    cssVarsShim: true,
```
danschultz, line 38: Edge >= 16 supports CSS variables. https://caniuse.com/css-variables

jdm717: Removed

MadisonEhlers-Vertex (approved):

danschultz:
> One note here, this doesn't restore all the defaults for Stencil 1, only those for EdgeHTML <= 18. See the [changelog](https://github.com/ionic-team/stencil/blob/master/CHANGELOG.md#restore-stencil-1-defaults) for more info on restoring those defaults. Curious if anyone has thoughts on whether we should restore all those defaults?

My preference would be to just include support for Edge. IE11 is ancient, and Safari 10 was release in 2016.

danschultz (approved):

PR #146, opened 2021-03-04
[SDK-1422] Twist touch interaction

jdm717:
## What

Adds support for the twist interaction as a mobile gesture.

https://user-images.githubusercontent.com/5252480/109891552-e8cd6400-7c4e-11eb-90f3-09e72444a226.mov

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1422

## Test Plan

- Test twist gesture on mobile devices
- Test twist on desktop devices

Twist on desktop

N/A

## Dependencies

N/A


packages/viewer/src/components.d.ts
```diff
@@ -4,424 +4,345 @@
  * This is an autogenerated file created by the Stencil compiler.
```
jdm717, line 4: This file appears to update for me running the dev server. Unsure if we've just made changes + not run the dev server prior to this change, so I've left it in for now

packages/geometry/src/matrix2.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * A 2x2 matrix
+ */
+export interface Matrix {
+  a: number;
+  b: number;
+  c: number;
+  d: number;
+}
```
danschultz, line 11: Please add some documentation about the where each value is positioned in the matrix. See https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/geometry/src/matrix.ts#L10-L14

danschultz (approved):
LGTM after doc update.

packages/viewer/src/components.d.ts
danschultz: Must be a change with the Stencil 2 update.

danschultz (approved):

packages/geometry/src/matrix2.ts
jdm717: Updated. Also updated that documentation at https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/geometry/src/matrix.ts#L10-L14. Looked like it referenced `b` twice in the doc

PR #151, opened 2021-03-25
[SDK-1513] Perform drag when beginning interaction

jdm717:
## What

Corrects a couple behaviors from the SDK that were resulting in seemingly lost `tap` events:
- Performs a drag immediately after beginning an interaction since we will have moved the threshold amount
- Emits a tap when a longpress occurred, and adds both a `longPress` and `doubleTap` property to the tap event info to indicate whether it was part of either interaction

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1513

## Test Plan

- Test tap events (longpress, doubletap, tap)
- Test interactions, there should no longer be a "dead zone" where a tap is not emitted and an interaction doesn't begin visually

Tap events, interactions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/interactions/baseInteractionHandler.ts
+++ b/packages/viewer/src/interactions/baseInteractionHandler.ts
@@ -130,7 +130,6 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
     }
 
     const position = Point.create(event.screenX, event.screenY);
-    let didBeginDrag = false;
     const pixelThreshold =
       this.interactionApi != null
         ? this.interactionApi.pixelThreshold(this.isTouch(event))
@@ -141,13 +140,10 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
       !this.isDragging
     ) {
       this.beginDrag(event);
-      didBeginDrag = true;
       this.isDragging = true;
     }
 
-    // We only invoke drag interactions for mouse events after a beginDrag has
-    // been invoked.
-    if (!didBeginDrag && this.isDragging) {
+    if (this.isDragging) {
       this.drag(event);
     }
   }
--- a/packages/viewer/src/interactions/tapInteractionHandler.ts
+++ b/packages/viewer/src/interactions/tapInteractionHandler.ts
@@ -152,6 +152,10 @@ export class TapInteractionHandler implements InteractionHandler {
     isTouch = false
   ): void {
     if (position != null && this.pointerDownPosition != null) {
+      if (this.longPressTimer != null) {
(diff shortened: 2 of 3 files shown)


packages/viewer/src/interactions/baseInteractionHandler.ts
```diff
@@ -141,13 +140,12 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
       !this.isDragging
     ) {
       this.beginDrag(event);
-      didBeginDrag = true;
       this.isDragging = true;
     }
 
     // We only invoke drag interactions for mouse events after a beginDrag has
     // been invoked.
```
danconner-vertex, line 147: is this comment inaccurate now?

jdm717: Yeah, it looks like it will be. I'll remove it

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -46,23 +46,36 @@ export class InteractionApi {
...
-    keyDetails: Partial<TapEventKeys> = {}
+    keyDetails: Partial<TapEventKeys> = {},
+    eventInfo: Partial<TapEventInfo> = {}
   ): Promise<void> {
-    this.emitTapEvent(this.doubleTapEmitter.emit, position, keyDetails);
+    this.emitTapEvent(
+      this.doubleTapEmitter.emit,
+      position,
+      keyDetails,
+      eventInfo
```
danschultz, line 64: What's the need for `eventInfo`? Shouldn't the event name already communicate that this is a double tap or long press?

packages/viewer/src/interactions/tapInteractionHandler.ts
```diff
@@ -148,25 +145,27 @@ export class TapInteractionHandler implements InteractionHandler {
...
         this.clearDoubleTapTimer();
       }
 
-      if (this.longPressTimer != null) {
-        this.emit(this.interactionApi?.tap)(position, keyDetails);
-      }
+      this.emit(this.interactionApi?.tap)(position, keyDetails, {
+        doubleTap: isDoubleTap,
+        longPress: isLongPress,
+      });
```
danschultz, line 168: My experience from using most platforms is that click events are emitted before double click events.

packages/viewer/src/interactions/interactionApi.ts
jdm717: Yeah, I suppose in the case of a double tap this might not need to be populated. I can remove it

danschultz (approved):

PR #157, opened 2021-04-13
[SDK-1578] Scene tree init fix

danschultz:
## What

Should address a bug where the scene tree would not get initialized. This can happen if the viewer has already been loaded.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1578


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -50,6 +50,9 @@ export class SceneTree {
   @Prop({ reflect: true, mutable: true })
   public controller: SceneTreeController | undefined;
 
+  @Prop({ reflect: true, mutable: true })
+  public client!: SceneTreeAPIClient;
```
jdm717, line 54: Do we want this API client to be `reflect`ed here? Not sure it makes sense as a DOM attribute, unless I'm missing something here?

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -238,70 +241,85 @@ export class SceneTree {
...
+      );
 
-      if (newViewer != null) {
-        newViewer.addEventListener('sceneReady', this.handleViewerSceneReady);
-        newViewer.addEventListener(
-          'connectionChange',
-          this.handleViewerConnectionStatusChange
-        );
+      const isSceneReady = await newViewer.isSceneReady();
+      console.log('is scene ready', isSceneReady);
```
jdm717, line 266: Can we move this to `console.debug` or remove?

jdm717 (approved):
LGTM aside from a couple nits

packages/viewer/src/components/scene-tree/scene-tree.tsx
danschultz: Will remove.

packages/viewer/src/components/scene-tree/scene-tree.tsx
danschultz: I have a follow up PR to address this.

PR #158, opened 2021-04-13
[SDK-1579] Add method to invalidate scene tree

danschultz:
## What

Adds a method `SceneTree.invalidateRows()` to trigger a rerender of the rendered rows. This is useful if data that's returned by `rowData` changes and you need to refresh the UI.

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1579


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -104,8 +103,34 @@ export class SceneTree {
...
+      this.cleanupController();
+    }
+
+    this._controller = value;
+
+    if (this.controller != null) {
+      this.connectController(this.controller);
+    }
+  }
+  /* eslint-enable lines-between-class-members */
```
danschultz, line 133: Addresses https://github.com/Vertexvis/vertex-web-sdk/pull/157#discussion_r612583014

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -263,28 +294,13 @@ export class SceneTree {
       );
 
       const isSceneReady = await newViewer.isSceneReady();
-      console.log('is scene ready', isSceneReady);
```
danschultz, line 266: Addresses https://github.com/Vertexvis/vertex-web-sdk/pull/157#discussion_r612583384

packages/viewer/src/components/scene-tree/lib/dom.ts
```diff
@@ -0,0 +1,3 @@
+export function getSceneTreeViewportHeight(el: HTMLElement): number {
+  return el.clientHeight;
+}
```
jdm717, line 3: nit: Seems like this could just be a top level `lib` rather than specific to the scene tree

jdm717 (approved):
LGTM, had one nit, but it's pretty minor/probably doesn't need to be updated for this PR

PR #160, opened 2021-04-14
[SDK-1542] Fix invalid key states in SDK

MadisonEhlers-Vertex:
## What

These changes introduce removing the key state and relying on mouse/pointer events for getting the key state for those interactions. 

The two main features that changed in the web-sdk would be 
- Fly To Part (cmd + click)
- Twist Interaction (Alt + shift + rotate)

These features can still be optional by the consumer passing in keyboard-controls=false to the vertex-viewer:

```html
  <vertex-viewer
      id="viewer"
      client-id="client-id"
      config-env="platdev"
      keyboard-controls=false
      src="urn:vertexvis:stream-key:stream-key"
  ></vertex-viewer>
```

Additional context around the twist interaction handler:
_I removed the twist interaction handler, as we don't have a super clean way of getting a mouse event to pass into the predicate to determine turning off/on the twist operation. A solution like that would require adding pointerdown/move/up event listeners on the web component, similar to how we have tap events registered there, and I'm not sure we want to add a mouse listener there yet. Either direction will likely be revisited with [SDK-1515](https://vertexvis.atlassian.net/browse/SDK-1515)._ 

## Ticket

https://vertexvis.atlassian.net/browse/SDK-1542

## Test Plan

Test doing a twist op, or fly to part with the keyboard-controls on, and leave the page with buttons still pressed. Return to your viewer, and discontinue pressing on the buttons. You should not see a twist or fly to part operation by just clicking the model.


packages/viewer/src/interactions/baseInteractionHandler.ts
```diff
@@ -195,6 +200,15 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
   protected drag(event: BaseEvent): void {
+    if (this.keyboardControls && event.altKey && event.shiftKey) {
+      this.currentInteraction = this.twistInteraction;
+    } else {
+      this.currentInteraction = undefined;
+    }
+    this.draggingInteraction =
+      this.currentInteraction ||
+      this.draggingInteraction ||
+      this.primaryInteraction;
```
jdm717, line 211: Just curious, any reason this couldn't continue to live in the `twistInteractionHandler`? Seems like either way is probably fine right now though if this resolves the bug since https://vertexvis.atlassian.net/browse/SDK-1515 will hopefully address some of the challenges around these interactions

MadisonEhlers-Vertex: Chatted on zoom about this. Going to add some description that encapsulates the conversation around some of these things.

jdm717 (approved):
LGTM for the fix. My opinion would be to move forward with this change, and address some of the interaction type setting behavior/refactoring in https://vertexvis.atlassian.net/browse/SDK-1515 as mentioned in the description

PR #163, opened 2021-04-19
[API-1769] Add ability to expand and scroll to item

danschultz:
## What

Adds ability to expand and scroll to an item by ID. Introduces additions and changes to the the scene tree:

* `SceneTree.scrollToItem(id: string, options: ScrollToOptions)`: Fetches the index of the item and scrolls to the item's index. If the node for the item is collapsed, will expand its parents.
* `SceneTree.scrollToIndex(index: number, options: ScrollToOptions)`: Added options to specify if the row should be positioned at the start, middle or end or the viewport. Defaults to `middle`.

## Ticket

https://vertexvis.atlassian.net/browse/API-1769


.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 .github/CODEOWNERS @danschultz
-* @vertexvis/sdk
+* @vertexvis/platform
```
MadisonEhlers-Vertex, line 2: :(

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -194,6 +196,38 @@ export class SceneTreeController {
...
   }
 
+  /**
+   * Invokes a network request that will return the index of the node with the
+   * given ID. If the node is collapsed, all parents will be expanded.
+   *
+   * @param id An ID of an item.
+   * @returns A promise that resolves with the index of the node.
+   */
+  public async locateNode(id: string): Promise<number> {
```
MadisonEhlers-Vertex, line 206: The documentation here is good, but I wonder if this should be called `locateNodeIndex`? I could see a case where `locateNode` would maybe be used for returning more information about the specific item.

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -59,6 +60,22 @@ type OperationHandler = (data: {
...
 }) => void;
 
+/**
+ * A set of options to configure the scroll to index behavior.
+ */
+export interface ScrollToOptions {
+  /**
+   * Indicates if this operation will be animated. Defaults to `false`.
+   */
+  animate?: boolean;
```
MadisonEhlers-Vertex, line 70: Should this be of type `Animation` instead of a boolean type? That way you can include ms to indicate the speed of the scroll rate.

.github/CODEOWNERS
danschultz: APP+API are both in there :)

packages/viewer/src/components/scene-tree/scene-tree.tsx
danschultz: Unfortunately the browser doesn't give you any control over this using the native [`Element.scrollTo`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) behavior.

MadisonEhlers-Vertex (approved):

PR #165, opened 2021-04-20
Sdk 1605/spin center

jdm717:


packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -194,6 +202,70 @@ export class InteractionApi {
...
+          this.startingCameraInverseProjection,
+          normalizedScreenPoint
+        );
+
+        const scaledProjPoint = Vector3.scale(1.0 / point.w, point);
+
+        const viewPoint = Matrix4.multiplyVector3(
+          this.startingCameraInverseView,
+          scaledProjPoint
+        );
```
jdm717, line 252: Conversion of screen click to world position. This calculation follows examples from http://bookofhook.com/mousepick.pdf and loosely from https://www.khronos.org/opengl/wiki/GluProject_and_gluUnProject_code (this uses the combined view/projection matrix, and results in the same computed point).

packages/viewer/src/rendering/canvas.ts
```diff
@@ -134,3 +138,21 @@ export function createCanvasRenderer(): CanvasRenderer {
...
+): CanvasDepthProvider {
+  return (point) => {
+    if (canvas != null) {
+      const scale = imageScaleProvider();
+      const scaled = Point.scale(point, scale?.x || 1, scale?.y || 1);
+      const data = canvas
+        .getImageData(scaled.x, scaled.y, 1, 1)
+        .data.slice(0, 1);
+
+      return data[0] / 255.0;
```
jdm717, line 154: Depth value calculation. `data[0]` will be `0` at the near clipping plane and `255` at the far clipping plane

packages/viewer/src/interactions/interactionApi.ts
jdm717: One additional note, a couple of these values are adjusted based on testing -- the `y` component of the normalized device coordinate is flipped to go [-1, 1] from [bottom, top]. Additionally, the `z` component of the normalized device coordinate is in the range [0, 1] instead of [-1, 1] based on experimenting with the range, since it resulted in values that were closer to the expected world position

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -280,4 +330,28 @@ export class InteractionApi {
...
+      normalizedDeviceCoordinate
+    );
+
+    const scaledProjPoint = Vector3.scale(
+      1.0 / inverseProjPoint.w,
+      inverseProjPoint
+    );
+
+    return Matrix4.multiplyVector3(camera.inverseViewMatrix(), scaledProjPoint);
+  }
```
jdm717, line 356: Conversion of screen click to world position. This calculation follows examples from http://bookofhook.com/mousepick.pdf and loosely from https://www.khronos.org/opengl/wiki/GluProject_and_gluUnProject_code (this uses the combined view/projection matrix, and results in the same computed point).

jdm717: One additional note, a couple of these values are adjusted based on testing -- the `y` component of the normalized device coordinate is flipped to go [-1, 1] from [bottom, top]. Additionally, the `z` component of the normalized device coordinate is in the range [0, 1] instead of [-1, 1] based on experimenting with the range, since it resulted in values that were closer to the expected world position

danconner-vertex (approved):

packages/viewer/src/scenes/camera.ts
```diff
@@ -282,20 +375,44 @@ export class Camera implements FrameCamera.FrameCamera {
     return { ...this.data.position };
   }
 
+  /**
+   * The position vector for the camera when fit to the visible bounding box,
+   * in world space coordinates.
+   */
+  public get positionFitAll(): Vector3.Vector3 {
+    return { ...this.fitToBoundingBox(this.boundingBox).position };
+  }
```
danschultz, line 384: These fit all getters seem unnecessary to me. With this API, you start to go down the path of having getters for each of the camera fields for each of the different states a camera can be in. Just have the caller call `fitToBoundingBox` once and they should have access to the attributes of the fit camera.

packages/viewer/src/types/flags.ts
```diff
@@ -28,7 +28,14 @@ type Flag =
...
    */
-  | 'letterboxFrames';
+  | 'letterboxFrames'
+
+  /**
+   * Enables or disables the default rotation interaction being changed to
+   * rotate around the mouse down location. This requires the enabling of
+   * depth buffers through the viewer's `streamAttributes`.
+   */
+  | 'defaultToSpinCenter';
```
danschultz, line 38: I think this deserves to be a top-level prop on the viewer. AFAIK, this is an option that is enabled via a toolbar button in Creo, which seems more like a prop you want to set on the viewer vs resetting the whole config.

packages/viewer/src/types/frame.ts
```diff
@@ -29,15 +30,19 @@ export interface SceneAttributes {
 }
 
 export const fromProto = (
-  payload: vertexvis.protobuf.stream.IDrawFramePayload
+  payload: vertexvis.protobuf.stream.IDrawFramePayload,
+  isDepthBuffer = false
```
danschultz, line 34: Why is this a flag vs just including it if the depth buffer is set?

packages/viewer/src/rendering/coordinates.ts
```diff
@@ -0,0 +1,91 @@
...
+ * distance to the center of the bounding box compared to the distance to
+ * the center when fit to the bounding box. This corrects the normalized
+ * values to be relative to the original viewport.
+ * @param depth - The depth from [0, 1] between the near and far planes.
+ */
+export function computeNormalizedDeviceCoordinates(
+  viewport: Dimensions.Dimensions,
+  point: Point.Point,
+  distanceToCenterRatio: number,
+  depth?: number
```
danschultz, line 80: Doesn't look like `depth` is ever `undefined`.

packages/viewer/src/interactions/interactionApi.ts
```diff
@@ -194,6 +203,49 @@ export class InteractionApi {
     });
   }
 
+  public async rotateCameraAtPoint(
+    delta: Point.Point,
+    point: Point.Point
+  ): Promise<void> {
```
danschultz, line 209: There's code duplication with `rotateCamera` that suggests `point` could just be an arg to the existing function.

packages/viewer/src/types/frame.ts
jdm717: Yeah, probably doesn't need to be a flag. Was using this due to the canvas renderer expecting `image` to be populated with the data to draw, but I'll update to do that when passing the depth buffer instead

packages/viewer/src/rendering/coordinates.ts
jdm717: I think currently this actually is never defined (outside of tests). [Here](https://github.com/Vertexvis/vertex-web-sdk/pull/165/files#diff-90bfcbc92bff48dfcfc3dc462a906c0c8d738c64999d5851b8e2f3b6e3d42bb3R19-R23) is where it's used, and since we just override the `z` value, the depth isn't passed

packages/viewer/src/interactions/interactionApi.ts
jdm717: Yeah, I moved this to a separate method to prevent breaking changes to anyone using `rotateAroundAxis` currently. I could certainly update that to call `rotateAroundAxisAtPoint` with the lookat point at a minimum. Open to removing `rotateAroundAxisAtPoint` and just adding the arg to `rotateAroundAxis` though if that seems like the better option

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -157,6 +157,13 @@ export class Viewer {
    */
   @Prop() public keyboardControls = true;
 
+  /**
+   * Enables or disables the default rotation interaction being changed to
+   * rotate around the mouse down location. This requires the enabling of
+   * depth buffers through the viewer's `streamAttributes`.
```
danschultz, line 163: Can do this in a follow up, but IMO enabling this prop should enable depth buffers automatically.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -152,11 +157,18 @@ export class Viewer {
...
+   */
+  @Prop() public defaultSpinCenter = false;
```
danschultz, line 165: Discussed offline to rename.

danschultz (approved):
LGTM after feedback.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -362,6 +382,14 @@ export class Viewer {
...
             })}
           >
+            {this.getStreamAttributes().depthBuffers?.enabled && (
+              <canvas
+                ref={(ref) => (this.depthBufferCanvasElement = ref)}
+                class="depth-buffer-canvas"
+                width={1}
+                height={1}
+              ></canvas>
+            )}
```
danschultz, line 392: This isn't really required either :) You should be able to create a new canvas element and draw into it without it actually adding it to the DOM.

danschultz (approved):

PR #168, opened 2021-04-29
[API-1774] Virtualized scene tree list

danschultz:
## What

Updates the scene tree to use a custom virtualized list strategy that is more performant than VDOM and supports more customization of rows.

Virtualization is performed through `<template>` elements and a managed pool of created DOM elements. As a user scrolls the tree, elements are popped off the head or tail of the pool and shifted to the other end of the list depending on the direction of scrolling. Only these elements that are shifted need to be updated and results in less work for the browser to do.

```html
<html>
  <head>
    <style>
      .tree {
        --scene-tree-row-background: white;
        --scene-tree-row-background-hover: light-grey;
        --scene-tree-row-background-selected: linear-gradient(0deg, rgba(34,193,195,1) 0%, rgba(253,187,45,1) 100%);
      }

      .indicator {
        width: 8px;
        height: 8px;
        border-radius: 4px;
      }
    </style>
  </head>
  <body>
    <vertex-scene-tree class="tree">
      <template>
        <vertex-scene-tree-row prop:node={{row.node}}>
          <div slot="left-gutter" class="indicator" attr:style="background-color: {{row.data.color}};"></div>
        </vertex-scene-tree-row>
      </template>
    </vertex-scene-tree>

    <script type="module">
      const tree = document.querySelector('vertex-scene-tree');

      // This callback is invoked for each row that is presented in the tree.
      // It allows you to define custom data that can be passed to the row
      // through bindings. In this case, we specify a color that can be bound
      // through {{row.data.color}}.
      tree.rowData = (row) => {
        const color = `#${Math.round(Math.random() * (255 * 255 * 255))
          .toString(16)
          .padStart(6, '0')}`;
        return { color };
      }
    </script>
  </body>
</html>
```

## Ticket

https://vertexvis.atlassian.net/browse/API-1774


packages/viewer/src/components/scene-tree/lib/row.ts
```diff
@@ -1,38 +1,33 @@
...
+      node: nodes.toObject(),
       data: {},
     };
   }
 }
 /* eslint-enable padding-line-between-statements */
+
+export function isLoadedRow(obj: unknown): obj is LoadedRow {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return obj != null && (obj as any).hasOwnProperty('node');
```
therockstorm, line 32: You might be able to avoid the `eslint-disable` if you use `unknown` instead of `any`.

therockstorm (approved):

packages/viewer/src/components/scene-tree/lib/row.ts
danschultz: `unknown` is pretty restrictive. You need to cast it or run a type assertion over it before accessing anything in it. There's a bit more info here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type.

PR #175, opened 2021-05-05
[APP-2376] Spin-center math adjustment

jdm717:
## What

Adjusts the calculation of the world position when using spin-center. This approach uses similar logic to compute a ray to our rendering pipeline, and uses that in combination with the depth buffer value to determine a world position instead of using the camera matrices.

This resolves a couple issues related to camera zoom levels and click position.

## Ticket

https://vertexvis.atlassian.net/browse/APP-2376

## Test Plan

- Test spin-center behavior
  - Test at different zoom levels, and picking parts at the edge/corner of the screen
  - Test when clicking in whitespace

N/A

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/rendering/__tests__/coordinates.spec.ts
+++ b/packages/viewer/src/rendering/__tests__/coordinates.spec.ts
@@ -22,8 +22,6 @@ describe(computeWorldPosition, () => {
   it('should return correct world position for near-plane depth', () => {
     const depth = 0;
 
-    console.log(camera.near, camera.far);
-
     expect(computeWorldPosition(camera, viewport, point, depth).z).toBe(4);
   });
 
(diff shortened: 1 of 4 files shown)


packages/viewer/src/rendering/__tests__/coordinates.spec.ts
```diff
@@ -1,125 +1,41 @@
...
-        inverseView,
-        viewport,
-        point,
-        depth,
-        near,
-        far,
-        1
-      ).z
-    ).toBe(0);
+    console.log(camera.near, camera.far);
```
danschultz, line 25: Should remove.

danschultz (approved):
Aside nit, LGTM.

packages/viewer/src/rendering/__tests__/coordinates.spec.ts
jdm717: Missed that one, removed in https://github.com/Vertexvis/vertex-web-sdk/pull/175/commits/263d4f616a453378aa2525d63a186a39a0b6e7d3

PR #179, opened 2021-05-11
[API-1852] Fix bug with binding to camel cased events and props

danschultz:
## What

Fixes a bug where binding to a rows `selectionToggled` event wasn't working. This is because the browser treats HTML as case insensitive and will lowercase any attributes you set. To address this, properties and events can use dash case and it'll automatically be converted to camel case.

```html
<vertex-scene-tree class="tree">
  <template>
    <vertex-scene-tree-row prop:node={{row.node}} event:selection-toggled={{row.data.myHandler}}>
    </vertex-scene-tree-row>
  </template>
</vertex-scene-tree>
```

## Ticket

https://vertexvis.atlassian.net/browse/API-1852

### DIFF
--- a/packages/viewer/package.json
+++ b/packages/viewer/package.json
@@ -53,6 +53,7 @@
     "@vertexvis/scene-tree-protos": "^0.1.0",
     "@vertexvis/stream-api": "0.9.26",
     "@vertexvis/utils": "0.9.26",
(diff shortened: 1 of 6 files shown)


jdm717 (approved):
LGTM!

PR #191, opened 2021-06-07
[API-1978] Fix RTE when setting stream attributes and changing scene

danschultz:
## What

This is a potential fix for https://vertexvis.atlassian.net/browse/API-1878, which may be caused when quickly assigning stream attributes and changing the src of the stream, as might happen in React applications. This will eagerly disable the connection flag when `unload` is called which should prevent sending a message over a disconnecting websocket when `streamAttributes` is assigned.

## Ticket

https://vertexvis.atlassian.net/browse/API-1878

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -583,9 +583,12 @@ export class Viewer {
           frameType: 'final',
         },
       };
-      this.stream.updateStream({
-        streamAttributes: toProtoStreamAttributes(this.getStreamAttributes()),
-      });
+
+      if (this.isStreamStarted) {
+        this.stream.updateStream({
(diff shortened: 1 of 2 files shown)


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -701,6 +705,7 @@ export class Viewer {
    */
   public async handleWebSocketClose(): Promise<void> {
     if (this.isStreamStarted) {
+      console.log('handle ws closed');
```
danconner-vertex, line 708: do you want to leave this for continued visibility?

danschultz: Nope, removed.

danconner-vertex (approved):

PR #202, opened 2021-07-01
[API-1939] Add support for creating measurements

danschultz:
## What

Adds a new `<vertex-viewer-measurement-tool>` to support adding new measurements. You can use this component stand alone, but it is intended to be nested inside a `<vertex-viewer-measurements>` component. The `<vertex-viewer-measurements>` will listen for new measurements and add them as children to the component.

**Example:** Components for creating measurements.

```html
<html>
  <head>
    <style>
      .measurement {
        --viewer-distance-measurement-accent-color: green;
      }
    </style>
  </head>
  <body>
    <!-- This specifies the template for new measurements. It's optional. -->
    <template id="distance-measurement">
      <vertex-viewer-distance-measurement
        class="measurement"
      >
        <div slot="start-label">1</div>
        <div slot="end-label">2</div>
      </vertex-viewer-distance-measurement>
    </template>

    <vertex-viewer>
      <!-- This is the container of measurements. You can add measurements programmatically or it'll listen for user interaction  -->
      <vertex-viewer-measurements distance-template-id="my-distance-measurement">
        <!-- This is the tool that'll handle user interactions for creating measurements -->
        <vertex-viewer-measurement-tool></vertex-viewer-measurement-tool>
      </vertex-viewer-measurements>
    </vertex-viewer>
  </body>
</html>
```

https://user-images.githubusercontent.com/37429/124061706-99c0f100-d9ec-11eb-98b5-f518998083ca.mov

## Other Notable Changes

* I've had to change the interaction target of the viewer to be compatible with measurements. Instead of adding listeners to the canvas, listeners are now added to the root viewer element. Without this, the measurement elements would capture events and not give an opportunity for the viewer to handle them.
* You can use `Event.stopPropagation()` to tell the viewer to not handle your event.
* You can also wrap an element in `<vertex-viewer-layer>` and specify the `viewer-interactions-off` attribute to automatically stop propagation of events.

## Ticket

https://vertexvis.atlassian.net/browse/API-1939


packages/viewer/index.html
```diff
@@ -84,10 +94,18 @@
       <div class="workarea">
         <vertex-viewer
           id="viewer"
-          client-id="client-id"
           config-env="platdev"
-          src="urn:vertexvis:stream-key:key"
```
therockstorm, line 89: Intentional?

packages/viewer/src/components/viewer-dom-element/viewer-dom-element.css
```diff
@@ -1,6 +1,6 @@
 :host {
   position: absolute;
-  pointer-events: none;
+  /* pointer-events: none; */
```
therockstorm, line 3: Remove?

therockstorm (approved):

PR #212, opened 2021-07-13
[API-1982] Add feature line support to viewer

jareddellitt:
## What
Adds ability to set feature line width and color.

## Test Plan

```javascript
const v = document.getElementById('viewer');
viewer.featureLines = { width: 3, color: { r: 0, b: 255, g: 255 }};
```

https://user-images.githubusercontent.com/477684/125373217-5484b380-e34a-11eb-9f37-6bc19ca84515.mov

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/204/files


packages/viewer/src/lib/stream/streamAttributes.ts
```diff
@@ -46,6 +53,16 @@ export const toProtoStreamAttributes = (
           ? { value: attributes.experimentalGhosting?.opacity }
           : attributes.experimentalGhosting?.opacity,
     },
+    featureLines: {
+      lineWidth: attributes.featureLines?.width,
+      lineColor: attributes.featureLines?.color
+        ? {
```
therockstorm, line 59: Do we have a default color if only a width is defined?

jareddellitt: it defaults to black

therockstorm (approved):

PR #220, opened 2021-07-28
[API-2073] Support selection of measurements added as children

danschultz:
## What

Selection wasn't supported for measurements added as children.

## Ticket

https://vertexvis.atlassian.net/browse/API-2073


jdm717 (approved):
💯  LGTM. Thanks for looking into this!

PR #228, opened 2021-08-14
[APP-2536] Only run web workers in the browser for nextjs apps

MadisonEhlers-Vertex:

### DIFF
--- a/packages/viewer/src/workers/png-decoder-pool.ts
+++ b/packages/viewer/src/workers/png-decoder-pool.ts
@@ -7,6 +7,8 @@ type DecodePngPool = Pool<DecodePngFn>;
 let workerLoader: Promise<DecodePngModule> | undefined;
 let poolLoader: Promise<DecodePngPool> | undefined;
 
+const DEFAULT_POOL_SIZE = 1;
+
 function loadWorkerModule(): Promise<DecodePngModule> {
   if (workerLoader == null) {
     workerLoader = loadWorker();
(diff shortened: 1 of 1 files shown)


packages/viewer/src/workers/png-decoder-pool.ts
```diff
@@ -33,4 +33,6 @@ export const decodePng: DecodePngFn = async (bytes) => {
 };
 
 // Prefetch the worker and initialize the pool.
-getPool();
+if (typeof window !== 'undefined') {
+  getPool();
+}
```
danschultz, line 38: Can we do something like this? Technically a code path in Node could still call `decodePng` which would throw an exception.

```ts
import type { DecodePngFn } from './png-decoder';
import { loadWorker, WorkerModule, Pool } from 'worker:./png-decoder';

type DecodePngModule = WorkerModule<DecodePngFn>;
type DecodePngPool = Pool<DecodePngFn>;

let workerLoader: Promise<DecodePngModule> | undefined;
let poolLoader: Promise<DecodePngPool> | undefined;

function loadWorkerModule(): Promise<DecodePngModule> {
  if (workerLoader == null) {
    workerLoader = loadWorker();
  }
  return workerLoader;
}

async function getPool(): Promise<DecodePngPool> {
  if (poolLoader == null) {
    poolLoader = loadWorkerModule().then(({ spawnPool }) =>
      spawnPool({ size: getPoolSize() })
    );
  }
  return poolLoader;
}

function getPoolSize(): number {
  if (typeof window !== 'undefined') {
    return Math.ceil(window.navigator.hardwareConcurrency / 4);
  } else {
    return 1;
  }
}

export const decodePng: DecodePngFn = async (bytes) => {
  const pool = await getPool();
  return pool.queue((decode) => {
    return decode(bytes);
  });
};

// Prefetch the worker and initialize the pool.
getPool();
```

danschultz (approved):

PR #242, opened 2021-09-02
[APP-2619] Markup support

jdm717:
## Summary

Adds support for `circle` and `arrow` markup using a similar approach to measurement.

## Test Plan

- Test arrow markup
  - Test adding programmatically
  - Test adding using the markup tool
  - Test resizing
  - Test selection
  - Test providing a template
- Test circle markup
  - Test adding programmatically
  - Test adding using the markup tool
  - Test resizing
  - Test selection
  - Test providing a template

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


packages/viewer/src/components/viewer-markup-arrow/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+    Angle.toDegrees(Angle.fromPoints(start, end)) - 270
+  );
+
+  const height = Math.max(4, Math.min(16, distance * 0.25));
+  const sideLength = height / Math.cos(Angle.toRadians(90 - arrowAngle));
+
+  const rotation = Matrix.rotation(angle);
+  const arrowLeft = Point.polar(sideLength, Angle.toRadians(arrowAngle * 2));
+  const arrowRight = Point.polar(sideLength, Angle.toRadians(arrowAngle));
+  const arrowBase = Point.polar(height, Angle.toRadians(90)); //  * 0.8
```
MadisonEhlers-Vertex, line 26: is this comment needed?

packages/viewer/src/components/viewer-markup-arrow/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  start: Point.Point,
+  end: Point.Point,
+  arrowAngle = 60
+): ArrowheadPoints {
+  const distance = Point.distance(start, end);
+  const angle = Angle.normalize(
+    Angle.toDegrees(Angle.fromPoints(start, end)) - 270
+  );
+
+  const height = Math.max(4, Math.min(16, distance * 0.25));
```
MadisonEhlers-Vertex, line 20: What is this height accounting for? the `sideLength`? etc. Some notes/comments on this might not hurt to add to add above the method to help our future self. 😄

packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.tsx
```diff
@@ -0,0 +1,338 @@
...
+   * `<vertex-viewer-markup>` or `<vertex-viewer>` element.
+   */
+  @Prop()
+  public viewer?: HTMLVertexViewerElement;
+
+  /**
+   * An event that is dispatched anytime the user begins editing the
+   * markup.
+   */
+  @Event({ bubbles: true })
```
MadisonEhlers-Vertex, line 88: IIRC the `bubbles` value defaults to true, so this is redundant.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.spec.tsx
```diff
@@ -0,0 +1,53 @@
+jest.mock('../viewer/utils');
+jest.mock('../viewer-markup/dom');
+
+import '../../testing/domMocks';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
MadisonEhlers-Vertex, line 6: nit: Can this be removed?

MadisonEhlers-Vertex (commented):
This likely won't be anything we need to worry about until we actually start investigating free form markup, but was wondering if there are any things we need to do in this pr to consider the pressure on pointer events? Could this be something that we could leverage for arrows/circles?

https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pressure

packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.tsx
danschultz: Not sure if this is the case with Stencil. Would be good to 2x check.

packages/viewer/src/components/viewer-markup/viewer-markup-components.tsx
```diff
@@ -0,0 +1,236 @@
...
+      transform={angle ? `rotate(${angle},${center.x},${center.y})` : undefined}
+      onPointerDown={onGrab}
+    />
+  );
+};
+
+export interface BoundingBox1dProps {
+  start: Point.Point;
+  end: Point.Point;
+  deviceSize?: DeviceSize;
```
danschultz, line 60: This would be better customized from the applicationn through CSS variables.

packages/viewer/src/components/viewer-markup/viewer-markup-components.tsx
```diff
@@ -0,0 +1,236 @@
...
+        />
+        <BoundingBoxAnchor
+          id="bounding-box-1d-start-anchor"
+          center={start}
+          angle={angle}
+          width={anchorWidth}
+          height={anchorHeight}
+          onGrab={onStartAnchorPointerDown}
+        />
+        <BoundingBoxAnchor
```
danschultz, line 103: Can we make the selection bounding box and anchors be HTML elements vs SVG? SVG is difficult to customize using CSS variables. I'm thinking it'd be nice for folks to specify circular vs square handles, which can be done if we exposed a variable for border radius.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.spec.tsx
danschultz: Not without producing a lint warning. Unfortunately, it's a side-effect of how the Stencil compiler works.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
```diff
@@ -0,0 +1,355 @@
...
+  translatePointToScreen,
+} from '../viewer-markup/utils';
+import { getMarkupBoundingClientRect } from '../viewer-markup/dom';
+
+/**
+ * The supported arrow markup modes.
+ *
+ * @see {@link ViewerMarkupArrowMode.mode} - For more details about modes.
+ */
+export type ViewerMarkupArrowMode = 'edit' | 'replace' | '';
```
danschultz, line 36: `replace` is improperly named for the behavior it's doing.

packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.tsx
```diff
@@ -0,0 +1,338 @@
...
+} from '../viewer-markup/viewer-markup-components';
+import { parseBounds, transformCircle } from './utils';
+import { getMarkupBoundingClientRect } from '../viewer-markup/dom';
+
+/**
+ * The supported markup modes.
+ *
+ * @see {@link ViewerMarkupCircleMode.mode} - For more details about modes.
+ */
+export type ViewerMarkupCircleMode = 'edit' | 'replace' | '';
```
danschultz, line 33: ^

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -0,0 +1,385 @@
...
+    const tool = this.getMarkupTool();
+    if (tool != null) {
+      tool.disabled = this.disabled;
+      tool.arrowTemplateId = this.arrowTemplateId;
+      tool.circleTemplateId = this.circleTemplateId;
+      tool.tool = this.tool;
+      tool.viewer = this.viewer;
+
+      if (this.viewer != null) {
+        this.viewer.cameraControls = tool.cameraControls;
```
danschultz, line 375: Looks like this behavior is already handled by the markup tool.

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -0,0 +1,385 @@
...
+    >
+  > {
+    const circleMarkup = Array.from(this.hostEl.children).filter(
+      isVertexViewerCircleMarkup
+    );
+    const arrowMarkup = Array.from(this.hostEl.children).filter(
+      isVertexViewerArrowMarkup
+    );
+
+    return [...arrowMarkup, ...circleMarkup];
```
danschultz, line 178: Can you use a `||` in your filter function. This way element order is preserved.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.css
```diff
@@ -0,0 +1,138 @@
+:host {
+  font-size: 0.75rem;
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
+  left: 0;
+  top: 0;
+  pointer-events: none;
```
danschultz, line 9: Please review these properties. On first look, `font-size`, `overflow`, `left`, `top` and `pointer-events` are not needed.

packages/viewer/src/components/viewer-markup-arrow/utils.ts
danschultz: It adjusts the size of the arrow head, based on how long the arrow is.

packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.css
```diff
@@ -0,0 +1,127 @@
+:host {
+  font-size: 0.75rem;
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  overflow: hidden;
+  left: 0;
+  top: 0;
+  pointer-events: none;
```
danschultz, line 9: Please review props that are actually needed.

danschultz (commented):
Overall, I think the direction looks good. An enhancement would be to try to make the editing bounding boxes and handles be HTML, so they can be customized through CSS variables.

packages/viewer/src/components/viewer-markup-arrow/utils.ts
jdm717: Nope! This is pulled over from mfg-ui. I'll remove this

packages/viewer/src/components/viewer-markup/viewer-markup-components.tsx
jdm717: For sure, this was related to challenges with styling the SVG anchors. This should go away with swapping to HTML anchors

packages/viewer/src/components/viewer-markup/viewer-markup-components.tsx
jdm717: Yeah, I'll take a look at swapping over to that approach. I had run into some challenges with the SVG anchors as I tested with and used these changes, but hadn't done too much investigation into the HTML anchor approach. I'll take another look at that

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
jdm717: For sure. I had originally run into some typing issues when using the `||` here, but that was probably because I had the return typing here incorrect. I'll update

packages/viewer/src/components/viewer-markup/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Returns the scale factor to convert screen coordinates to canvas coordinates.
+ *
+ * Scale factor is determined by taking the shortest side of the given
+ * dimension. This behavior ensures that relative units are consistent in
+ * non-square aspect ratios.
+ *
+ * @see toRelativeScaleFactor
+ */
+export const toScreenScaleFactor = (
```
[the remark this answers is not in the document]
jdm717: Pulled this over from mfg-ui, so this is some leftover from that -- I'll update

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
jdm717: Actually looks like this is required to disable the interactions, but I'm not certain why. I've verify that the markup tool is setting `cameraControls` on the viewer it receives to false, but they remain enabled. Curious if you've seen that behavior at all with some of the other components?

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.css
jdm717: Looks like `font-size` and `overflow` are not needed, but `left` and `top` are needed to prevent the arrow from appearing below the viewer, and `pointer-events` is necessary to allow the `pointerdown` from the viewer to trigger the listener in the arrow/circle markup

packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.tsx
MadisonEhlers-Vertex: Sounds good. I was wrong :)

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
```diff
@@ -0,0 +1,352 @@
...
+              {this.mode === 'edit' && (
+                <line
+                  id="bounding-box-1d-line"
+                  class="bounds-line"
+                  x1={screenStart.x}
+                  y1={screenStart.y}
+                  x2={screenEnd.x}
+                  y2={screenEnd.y}
+                />
+              )}
```
jdm717, line 237: @danschultz converted everything bounding box related to HTML over SVG except for this `<line />` element. Didn't have much luck consistently centering that on the arrow with a div. Had some difficulty with the `rotateZ` and positioning it with `translate`. Open to suggestions here, but since this isn't possible to interact with, I left it as an SVG for now

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -0,0 +1,382 @@
...
+      return el;
+    } else if (markup instanceof CircleMarkup) {
+      const { bounds, id } = markup;
+
+      const el = this.createCircleMarkupElement();
+      el.id = id;
+      el.bounds = bounds;
+
+      this.updatePropsOnMarkup(el);
+      this.hostEl.appendChild(el);
```
danschultz, line 125: Missing a `this.markupAdded.emit(el);`. Might be a good opportunity to pull this shared code into a function.

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -0,0 +1,382 @@
...
+  @Method()
+  public async removeMarkup(
+    id: string
+  ): Promise<
+    | HTMLVertexViewerMarkupArrowElement
+    | HTMLVertexViewerMarkupCircleElement
+    | undefined
+  > {
+    const markups = await this.getMarkupElements();
+    const markup = markups.find((m) => m.id === id);
```
danschultz, line 149: We can reuse `getMarkupElement()` here.

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
danschultz: I suspect it's because `viewer.cameraControls` doesn't respond to being changed. It's working because of the timing of how components are initialized. This may cause problems if you add then remove the tool.

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
danschultz, line 125: Missing a `this.markupAdded.emit(el);`. Might be a good opportunity to pull this shared code into a function.

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
danschultz, line 149: We can reuse `getMarkupElement()` here.

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
danschultz: I suspect it's because `viewer.cameraControls` doesn't respond to being changed. It's working because of the timing of how components are initialized. This may cause problems if you add then remove the tool.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.css
```diff
@@ -0,0 +1,170 @@
...
+  width: 100%;
+  height: 100%;
+}
+
+.line {
+  pointer-events: auto;
+  stroke: var(--viewer-markup-arrow-line-stroke-color);
+  stroke-width: var(--viewer-markup-arrow-line-stroke-width);
+  fill: var(--viewer-markup-arrow-line-fill-color);
+  fill-opacity: var(--viewer-markup-arrow-line-fill-opacity);
```
danschultz, line 111: Is there ever a case of filling a line?

danschultz (approved):
Overall, LGTM. Might want to check to see if `viewer.cameraControls` is behaving the way you'd expect when adding/removing markup elements. As an alternative approach, I think you could also an overlay element that's managed by your arrow/circle component and listen to its events.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.css
jdm717: Yeah, I suppose this probably isn't needed. We set these in mfg-ui, but I don't know of a case where the `line` element could display that fill. I'll remove this

PR #255, opened 2021-10-04
[PLAT-88] Implement smoother pan behavior for mouse interactions

danschultz:
## Summary

This improves the pan behavior for mouse interactions. The behavior will try to keep any geometry under the mouse during pan, which makes panning feel like you're interacting directly with the surface of the model. If there is not geometry under the mouse, then will fallback to using the plane of the look at point.

https://vertexvis.atlassian.net/browse/PLAT-88

Old:

https://user-images.githubusercontent.com/37429/135931988-3e006754-c9c9-4875-a3b8-ea49e21d1ad2.mov

New:

https://user-images.githubusercontent.com/37429/135931994-afdb6acd-974e-4299-8ab1-85a1df47e72e.mov


gbiv (approved):
Tested on different models. Definitely better behavior with these changes. 
:shipit:

mikemccarty-vertex (approved):
LGTM

MadisonEhlers-Vertex (approved):

PR #270, opened 2021-10-29
[APP-2566] Additional interaction shortcuts

amvertex:
## Summary
We want keyboard shortcuts to rotate, pan, and zoom when the option is not selected. We also want to be able to perform a fit all by pressing the "F" key.

## Test Plan
Test the shortcuts in the viewer.

## Release Notes
Keyboard shortcuts for rotate, pan, zoom, and fit all

## Possible Regressions
None

## Dependencies
None


packages/viewer/src/lib/interactions/keyInteractionHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  public constructor(private getScene: SceneProvider) {
+    this.fitAllWithFKey = this.fitAllWithFKey.bind(this);
+  }
+
+  public initialize(element: HTMLElement, api: InteractionApi): void {
+    this.element = element;
+    this.interactionApi = api;
+
+    window.addEventListener('keypress', this.fitAllWithFKey);
```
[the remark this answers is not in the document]
amvertex: How do we connect the keyboard to the viewer? I tried a few different ways, but I wasn't able to get the viewer to register the keyboard event.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -201,8 +201,19 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
   }
 
   protected beginDrag(event: BaseEvent): void {
+    const isMac = /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform);
```
danconner-vertex, line 204: looking at what we're doing in man-ui, it doesn't appear that we specifically distinguish based on the user agent. would be interested to pursue if there are ways we can continue, but also reviewing that code more to understand it better.

danconner-vertex: looking at https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/viewer/views/MonoscopicViewer.tsx#L779

which is using https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/utils/platform.ts#L82

amvertex: We do for the pan tool (the same operation we are doing here). I got this code from https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/utils/platform.ts#L121 and https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/viewer_tools/views/common/PanModelTool.tsx#L14.

danconner-vertex: I _think_ those are only used to decide whether to display the unicode icon in the tooltip or not, but isn't used when deciding which interaction mode to use.

packages/viewer/src/lib/interactions/keyInteractionHandler.ts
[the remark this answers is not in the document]
danschultz: Keyboard events are based on focus. So you'd need to click or tab to the viewer element, and then it should receive keyboard events. Which is kind of why having some form of user feedback is important.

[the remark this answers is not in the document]
danconner-vertex: pretty interesting question here. the key events in man-ui are attached to window:
https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/viewer/views/MonoscopicViewer.tsx#L157

I don't think we want a user to be concerned about whether the viewer itself has focus before they use the 'f' shortcut key, or any others. and I don't think we want to try to maintain focus on the viewer element at all points the user expects it.

so is that an argument for handling these key events outside the viewer? the interaction handlers can be defined outside the SDK and passed in right? maybe that direction would make more sense vs handling window events in the viewer component.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
amvertex: I looked into it some more and in legacy Connect, we actually pan when using either the command or control key on macs. https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/viewer/views/MonoscopicViewer.tsx#L794 and https://github.com/Vertexvis/manufacturing-ui/blob/master/src/js/modules/utils/platform.ts#L28-L29. Would we rather do that?

danconner-vertex: I would be ok with maintaining current behavior and using either of those keys for pan.

amvertex: Sounds good. I updated the code accordingly.

packages/viewer/src/lib/interactions/keyInteractionHandler.ts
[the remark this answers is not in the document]
amvertex: Are you suggesting that the fit all shortcut should be implemented in connect-app instead of here?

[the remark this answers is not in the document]
amvertex: I removed the fit all shortcut and we will implement it in connect-app instead.

danconner-vertex (approved):

danschultz (approved):

PR #275, opened 2021-11-10
[PLAT-26] Scene tree table support

jdm717:
## Summary

Adds support for a table-based scene-tree layout to represent additional metadata alongside the existing name representation.

By default, the `<vertex-scene-tree />` element will create a basic table structure that is visually the same as the row-based layout, but additional columns can be defined as shown below:

```typescript
// Setup to enable desired metadata availability
function main() {
  const tree = document.getElementById('scene-tree');

  tree.metadataKeys = ['MetadataValue1', 'MetadataValue2'];
}
```
```html
<vertex-scene-tree id="scene-tree" viewer-selector="#viewer">
  <vertex-scene-tree-table>
    <vertex-scene-tree-table-column initial-width="150">
      <template slot="header">
        <vertex-scene-tree-table-header label="Name" />
      </template>
      <template>
        <vertex-scene-tree-table-cell prop:value="{{row.node.name}}" expand-toggle />
      </template>
    </vertex-scene-tree-table-column>

    <vertex-scene-tree-table-column initial-width="100">
      <template slot="header">
        <vertex-scene-tree-table-header label="Metadata Label 1" />
      </template>
      <template>
        <vertex-scene-tree-table-cell prop:value="{{row.metadata.MetadataValue1}}">
        </vertex-scene-tree-table-cell>
      </template>
    </vertex-scene-tree-table-column>

    <vertex-scene-tree-table-column initial-width="100">
      <template slot="header">
        <vertex-scene-tree-table-header label="Metadata Label 2" />
      </template>
      <template>
        <vertex-scene-tree-table-cell prop:value="{{row.metadata.MetadataValue2}}" visibility-toggle>
        </vertex-scene-tree-table-cell>
      </template>
    </vertex-scene-tree-table-column>
  </vertex-scene-tree-table>
</vertex-scene-tree>
```

## Test Plan

- Test existing scene-tree methods (`getRowAtClientY`, `getRowAtIndex`, etc)
- Test rendering the scene-tree without providing a custom table element
- Test rendering the scene-tree providing custom column headers
- Test rendering the scene-tree providing custom column cells

## Release Notes

N/A

## Possible Regressions

This PR causes a breaking change to the scene-tree element, removing the existing `<vertex-scene-tree-row>` element and the associated default template, instead using this new `<vertex-scene-tree-table>` layout.

## Dependencies

N/A


packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+  >;
+  headerInstances?: Array<InstancedTemplate<HTMLElement>>;
+
+  startIndex: number;
+  endIndex: number;
+  viewportRows: Row[];
+}
+
+@Component({
+  tag: 'vertex-scene-tree-table',
```
danschultz, line 37: How about renaming this to `vertex-scene-tree-table-layout`? The current name implies to me this can be used in place of `vertex-scene-tree`.

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+  @Prop({ mutable: true })
+  public layoutHeight?: number;
+
+  /**
+   * A flag that disables the default interactions of this component. If
+   * disabled, you can use the event handlers to be notified when certain
+   * operations are performed by the user.
+   */
+  @Prop()
+  public interactionsDisabled = false;
```
danschultz, line 140: I'd recommend keeping this out of the layout component. It's an implementation detail of the `vertex-scene-tree-table-cell` component.

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+
+  /**
+   * A flag that disables selection of the node's parent if the user selects
+   * the row multiple times. When enabled, selection of the same row multiple
+   * times will recursively select the next unselected parent until the root
+   * node is selected.
+   */
+  @Prop()
+  public recurseParentSelectionDisabled = false;
```
danschultz, line 149: Same here.

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+      }
+
+      col.style.minHeight = `${this.rowHeight * this.totalRows}px`;
+
+      pool.iterateElements((el, binding, rowIndex) => {
+        const row = this.stateMap.viewportRows[rowIndex];
+        if (row != null) {
+          this.updateCell(
+            row,
+            el as HTMLVertexSceneTreeTableCellElement,
```
danschultz, line 307: I think we should assume that a user can swap out the cell to something other than a `vertex-scene-tree-table-cell`. This way we have an escape hatch if devs don't want to use our component.

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+  private resizeObserver?: ResizeObserver;
+
+  private tableElement?: HTMLDivElement;
+  private headerElement?: HTMLDivElement;
+  private columnElements: HTMLVertexSceneTreeTableColumnElement[] = [];
+
+  public componentWillLoad(): void {
+    this.updateColumnElements();
+    this.computeColumnGridLayout();
+    this.createPools();
```
danschultz, line 185: I think we'll also need to call some variation of these methods if the column counts change. Can probably handle that in a follow up.

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+      }
+      this.rowHeight = height ?? this.rowHeight;
+      element.remove();
+      this.isComputingCellHeight = false;
+    }
+  };
+
+  private computeColumnGridLayout = (): void => {
+    this.columnGridLayout = `${this.columnElements
+      .map((c) => c.initialWidth)
```
danschultz, line 454: This won't allow for a developer to programmatically change the width of a column, or query for the current width of the column. I think we can restructure a couple things to allow for this:

1. Move the header to be a child of a column.
1. Use `position: sticky` for the header element, and offset the cells by the header height.
1. Instead of grid, use flexbox to just layout columns horizontally. Resizing a column will just push over the rest of the columns.

Don't think this is a blocker for this PR. But, if it's not a huge lift would be nice to support.

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
```diff
@@ -0,0 +1,511 @@
...
+  }
+
+  @Method()
+  public async getViewportStartIndex(): Promise<number> {
+    return this.stateMap.startIndex;
+  }
+
+  @Method()
+  public async getViewportEndIndex(): Promise<number> {
+    return this.stateMap.endIndex;
```
danschultz, line 236: These seem like props to me, in the same way that `layoutOffset` and `scrollOffset` are props.

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -0,0 +1,236 @@
...
+          </div>
+          {!this.visibilityToggle && <div class="column-spacer" />}
+        </div>
+      </Host>
+    );
+  }
+
+  private displayValue = (): string => {
+    return this.value == null || this.value.replace(' ', '') === ''
+      ? '--'
```
danschultz, line 190: Would be nice to make this configurable via a prop.

packages/viewer/src/components/scene-tree-table-header/scene-tree-table-header.css
```diff
@@ -0,0 +1,12 @@
...
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  box-sizing: border-box;
+  font-size: 0.875rem;
+  height: 2rem;
```
danschultz, line 11: You should let the content of the element dictate its height. Otherwise you have scenarios where the content can overflow its container.

packages/viewer/src/components/scene-tree-table-header/scene-tree-table-header.tsx
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   * The label to display in this header.
+   */
+  @Prop()
+  public label?: string;
+
+  public render(): h.JSX.IntrinsicElements {
+    return (
+      <Host>
+        <div class="header">{this.label}</div>
```
danschultz, line 18: This could also be a slot. Gives a bit more flexibility.

```html
<vertex-scene-tree-table-header>
  <button>my label</button>
</vertex-scene-tree-table-header>
```

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -195,6 +195,10 @@ export class SceneTreeController {
     }
 
     this.startIdleReconnectTimer();
+
+    if (this.metadataKeys.length > 0) {
+      this.setMetadataKeys(this.metadataKeys);
```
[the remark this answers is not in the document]
jdm717: In most cases the call to `setMetadataKeys` occurs prior to the connection status of the scene-tree moving to `connected`, which causes the keys to never be updated without this setup, so this is just to ensure that if we have metadata keys specified at connection time that the tree is updated

packages/viewer/src/components/scene-tree/lib/row.ts
```diff
@@ -43,14 +43,14 @@ export function fromNodeProto(
...
 function makeMetadataMap(keys: MetadataKey[], values: string[]): MetadataMap {
-  if (keys.length !== values.length) {
-    throw new Error('Column key length and column value length mismatch');
-  } else {
-    return keys.reduce((map, key, i) => {
-      map[key] = values[i];
-      return map;
-    }, {} as MetadataMap);
-  }
+  // if (keys.length !== values.length) {
```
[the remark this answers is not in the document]
jdm717: Yeah, from testing there are a number of models where certain nodes do not have values matching specified keys. If this is left in place, an exception would be thrown any time one of those "optional" metadata keys is specified as a column to display. This commented out block should be removed however

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
jdm717: Sounds good, this was mainly for convenience in the case that all interactions for the scene tree should be disabled, but it's not much more effort to specify on a per cell template basis, so I'll remove this

packages/viewer/src/components/scene-tree-table/scene-tree-table.tsx
jdm717: I'll take a look at this approach. Some of the column resize behavior in PLAT-27 is dependent on the grid based layout, but ideally that's not too big of a change to support

danschultz (approved):

danschultz (approved):

PR #277, opened 2021-11-15
[PLAT-254] Fix mispositioned snap

danschultz:
## Summary

Should address an issue with snapping where the mouse would not snap to a feature edge. This was happening because the stencil buffer is generated without antialiasing, and the depth buffer is. This makes a few changes to address this issue:

1. We now generate a depth buffer _with_ the stencil buffer. When a snap point is found, we use the stencil's depth buffer to compute the world coordinate.
2. We now snap to the center of the pixel.

https://vertexvis.atlassian.net/browse/PLAT-254


mikemccarty-vertex (approved):
LGTM

PR #281, opened 2021-11-16
[PLAT-372] make the viewer mutable in precise measurement

MadisonEhlers-Vertex:
## Summary

The viewer is by default `immutable` which my hunch is that the `PreciseMeasurement` is getting initialized prior to the viewer ref being present, and always returning `undefined`

The `JWT Provider` that is returning the jwt for precise measurement likely was getting a viewer that was null, which resulted in the error. https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/viewer-measurement-precise/viewer-measurement-precise.tsx#L81

### DIFF
--- a/packages/viewer/src/components/viewer-measurement-precise/viewer-measurement-precise.tsx
+++ b/packages/viewer/src/components/viewer-measurement-precise/viewer-measurement-precise.tsx
@@ -23,7 +23,7 @@ export class ViewerMeasurementPrecise {
   @Prop({ mutable: true })
   public measurementController?: MeasurementController;
 
-  @Prop()
+  @Prop({ mutable: true })
   public viewer?: HTMLVertexViewerElement;
 
   @Prop()


packages/viewer/src/components/viewer-measurement-precise/viewer-measurement-precise.tsx
```diff
@@ -23,7 +23,7 @@ export class ViewerMeasurementPrecise {
   @Prop({ mutable: true })
   public measurementController?: MeasurementController;
 
-  @Prop()
+  @Prop({ mutable: true })
```
danschultz, line 26: This only affects changes made internally by the component. Consumers can still set this property and Stencil will rerender, invoke callbacks, etc.

MadisonEhlers-Vertex: Are you suggesting I should look at the where this is being used instead of changing something here?

danschultz: Ya, I think the issue is elsewhere. Will ping ya.

PR #282, opened 2021-11-17
[PLAT-232] Use hits to improve accuracy of pt-to-pt measurements

danschultz:
## Summary

Refactoring of the `vertex-viewer-measurement-distance` component to request a hit when performing point-to-point measurements. Before we solely relied on the depth buffer to determine the world space position. Because the depth buffer is lossy, we lose accuracy.

The interaction and state management of the component was also getting pretty complicated. I moved a lot of this into an `PointToPointInteractionController` that is now responsible for maintaining and mutating this state. This class exposes a couple methods `PointToPointInteractionController.newInteraction` and `PointToPointInteractionController.editInteraction` that both return a `PointToPointInteraction` interface. On this interface there's an `update` method that is called by the component to update the interaction, and a `finish` method to complete the interaction.

https://vertexvis.atlassian.net/browse/PLAT-232


jdm717 (approved):
LGTM! +11

PR #283, opened 2021-11-18
[PLAT-19] Fix clipping bug with measurement lines when point behind camera

danschultz:
## Summary

Fixes an issue where measurement lines were mispositioned if one of its anchors were behind the camera's near plane. Math should be fixed now.

https://vertexvis.atlassian.net/browse/PLAT-19

### DIFF
--- a/packages/geometry/src/plane.ts
+++ b/packages/geometry/src/plane.ts
@@ -75,7 +75,7 @@ export function intersectLine(
   if (t < 0 || t > 1) {
     return undefined;
   } else {
-    return Vector3.add(line.start, Vector3.scale(t, direction));
+    return Vector3.add(Vector3.scale(t, direction), line.start);
   }
 }
 
--- a/packages/viewer/src/lib/types/frame.ts
+++ b/packages/viewer/src/lib/types/frame.ts
@@ -4,6 +4,8 @@ import {
   Rectangle,
   BoundingBox,
   Matrix4,
+  Line3,
+  Plane,
 } from '@vertexvis/geometry';
 import * as FrameCamera from './frameCamera';
 import * as CrossSectioning from './crossSectioning';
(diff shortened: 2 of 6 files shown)


packages/geometry/src/plane.ts
```diff
@@ -75,7 +75,7 @@ export function intersectLine(
   if (t < 0 || t > 1) {
     return undefined;
   } else {
-    return Vector3.add(line.start, Vector3.scale(t, direction));
+    return Vector3.add(Vector3.scale(t, direction), line.start);
```
mikemccarty-vertex, line 78: is there an important difference here (I hope not)?

packages/viewer/src/lib/types/frame.ts
```diff
@@ -171,27 +201,38 @@ export class FramePerspectiveCamera
...
+   *
+   * @param line The line to intersect.
+   * @returns A point in world space, or `undefined` if the line does not
+   * intersect with the near plane.
+   */
+  public intersectLineWithNear(line: Line3.Line3): Vector3.Vector3 | undefined {
+    const { position, lookAt, direction, near } = this;
+
+    const vs = Vector3.subtract(line.start, lookAt);
+    const ve = Vector3.subtract(line.end, lookAt);
```
mikemccarty-vertex, line 229: Shouldn't these be subtracting `position`?

packages/viewer/src/lib/types/frame.ts
```diff
@@ -171,27 +201,38 @@ export class FramePerspectiveCamera
...
+    const vl = Line3.create({ start: vs, end: ve });
+
+    const dist = Vector3.distance(lookAt, position) - near;
+    const nearP = Plane.create({ normal: direction, constant: dist });
```
mikemccarty-vertex, line 233: I'm thinking this should be `constant: -dist` since in the `Plane` functions, the `constant` is always added (hint: the plane is formed by all of the points where the sum of all the terms is zero so the `constant` needs to pull the sum in the opposite direction of everything else to make that happen).  This might explain why you were subtracting `lookAt` above (one sorta reverses the effect of the other), but I can also sometimes get myself confused.

Given that `dist` is a function of `position` (which intuitively makes total sense to me), it follows the rest of the vectors here should also be a function of `position` and not `lookAt`.

packages/viewer/src/lib/types/frame.ts
danschultz: Resolving to consolidate conversation in other thread.

packages/viewer/src/lib/types/frame.ts
danschultz: My understanding of `constant` is that it's a distance from an origin. I guess origin could either be `position` or `lookAt`, but `lookAt` is a better mental model for me, as it's the center of rotation for the camera.

mikemccarty-vertex: Except `near` is defined relative to `position`, not `lookAt` so you're kinda mixing two coordinate systems together which happens to sorta work but I wouldn't depend on it.

mikemccarty-vertex (approved):
:shipit:

PR #289, opened 2021-12-03
[PLAT-385] Refresh tokens

danschultz:
## Summary

Uses the new refresh tokens endpoint in FSS to fix a bug where a token would become invalid if a user was consistently using the viewer. This happened because our old refresh token strategy would only get a new token when reconnecting the WS. Since we only reconnect if the WS is idle, it'd cause tokens to expire.

As part of this work, wanted to move the connection logic out of `<vertex-viewer>`. This helps make the component just be responsible for presentation, and makes testing easier. Also removed some legacy code paths around `Commands` that was being used for connection behavior. This is all centralized now within `ViewerStream` implementation.

https://vertexvis.atlassian.net/browse/PLAT-385


packages/utils/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json.schemastore.org/tsconfig",
-  "extends": "@vertexwebsdk/build/tsconfig",
+  "extends": "@vertexwebsdk/build/tsconfig-web",
```
jdm717, line 3: Any concern that this config would allow for methods that are unsafe for NodeJS? Not sure if we're using it in any node projects, but our rollup config specifies `isMultiplatform: true`, and this would allow me to use `window` functions that'd cause issues there

jdm717 (approved):
LGTM! Just had one thought on the updated utils `tsconfig`

therockstorm (approved):

PR #291, opened 2021-12-04
[PLAT-175] Twist interaction updates

jdm717:
## Summary

Makes a couple updates to the twist (Alt+Shift) interaction:
- Moves to using client X/Y instead of screen X/Y. This corrects some odd behavior when the viewport is scaled, since the `twistCamera` method uses the center of the viewport as a reference point for angles.
- Adds a check when ending an interaction to set the `currentInteraction` back to undefined in the case of twist to prevent the current case where twist applies to the interaction immediately following it 

## Test Plan

- Test scaling a browser window to 50% -- twist should work as expected
- Test performing a twist, then a second interaction with no modifier keys -- this should default back to the primary interaction

## Release Notes

N/A

## Possible Regressions

Twist

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
+++ b/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
@@ -252,6 +252,13 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
   }
 
   protected endDrag(event: BaseEvent): void {
+    if (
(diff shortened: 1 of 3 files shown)


amvertex (approved):
Looks great! I couldn't reproduce the bug when I ran this locally.

MadisonEhlers-Vertex (approved):

PR #295, opened 2021-12-07
[PLAT-385] Set min size for viewer

danschultz:
## Summary

Fixes a couple bugs during testing with Connect+ where the viewer would fail to connect because the dimensions were `(0,0)`. We now set a min width/height on the viewer of 1px and check if the args to update stream have empty dimensions.

https://vertexvis.atlassian.net/browse/PLAT-385


chrisgray-vertex (approved):
LGTM

PR #297, opened 2021-12-09
[PLAT-285] Select multiple scene tree items at a time

amvertex:
## Summary
We want to be able to select adjacent parts in the scene tree when the user is holding down the shift key.

## Test Plan
Test holding down the shift key and selecting parts in the scene tree and verify that the entire range becomes selected. Also, holding down the meta key will also select the part you clicked on, without deselecting currently selected parts, though will not select the entire range. 

## Release Notes
Shift+click to multiselect in the BOM

## Possible Regressions
Scene tree selection

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/scene-tree/scene-tree.tsx
+++ b/packages/viewer/src/components/scene-tree/scene-tree.tsx
@@ -183,6 +183,9 @@ export class SceneTree {
   @Prop({ mutable: true })
   public controller?: SceneTreeController;
 
+  @Prop({ mutable: true })
+  public multiSelectedRows?: Row[];
+
   /**
    * A list of part metadata keys that will be made available to each row. This
    * metadata can be used for data binding inside the scene tree's template.
@@ -387,6 +390,33 @@ export class SceneTree {
   }
 
   /**
+   * Performs an API call that will select the items with indexes between the given bounds.
+   *
+   * @param lowerBound The smaller index of the set of rows to select.
+   * @param upperBound The larger index of the set of rows to select.
+   * @param viewer The viewer to select in.
+   */
+  @Method()
+  public async performMultiSelection(
+    lowerBound: number,
+    upperBound: number,
+    viewer: HTMLVertexViewerElement
+  ): Promise<void> {
+    console.log(
+      'Updating multiselection between ' + lowerBound + ' and ' + upperBound
+    );
+    const rowsToMultiSelect = this.rows.filter(
+      (row) => row.index && row.index <= upperBound && row.index >= lowerBound
+    );
+    this.multiSelectedRows = rowsToMultiSelect;
+    if (viewer && rowsToMultiSelect) {
+      await rowsToMultiSelect.forEach((row) => {
+        selectItem(viewer, row?.node?.id?.hex, { append: true });
+      });
+    }
+  }
+
+  /**
    * Performs an API call that will select the item associated to the given row
    * or row index.
    *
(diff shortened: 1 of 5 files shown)


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -386,6 +389,33 @@ export class SceneTree {
...
+      (row) =>
+        row?.index && row?.index <= upperBound && row?.index >= lowerBound
+    );
+    this.multiSelectedRows = rowsToMultiSelect;
+    if (viewer && rowsToMultiSelect) {
+      await rowsToMultiSelect.forEach((row) => {
+        if (row?.node?.id?.hex) {
+          selectItem(viewer, row?.node?.id?.hex, { append: true });
+        }
+      });
```
danschultz, line 415: Hypothetically, a user could select thousands of items, so this'd be a lot of requests going over the wire. @jareddellitt can we expose an RPC in STS to handle range selection?

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -183,6 +183,9 @@ export class SceneTree {
   @Prop({ mutable: true })
   public controller?: SceneTreeController;
 
+  @Prop({ mutable: true })
+  public multiSelectedRows?: Row[];
```
danschultz, line 187: This prop will need to be kept up-to-date when selections happen outside the tree. If not needed, maybe we exclude it for now?

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -386,6 +389,33 @@ export class SceneTree {
...
 
+  /**
+   * Performs an API call that will select the items with indexes between the given bounds.
+   *
+   * @param lowerBound The smaller index of the set of rows to select.
+   * @param upperBound The larger index of the set of rows to select.
+   * @param viewer The viewer to select in.
+   */
+  @Method()
+  public async performMultiSelection(
```
danschultz, line 400: Naming nit, but maybe call this `selectRange` so it shows up in docs near our other select methods.

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -386,6 +389,33 @@ export class SceneTree {
...
+    lowerBound: number,
+    upperBound: number,
+    viewer: HTMLVertexViewerElement
```
danschultz, line 403: The scene tree already has a reference to the viewer. Don't think this is needed as an arg.

packages/viewer/src/components/scene-tree/scene-tree.tsx
amvertex: Connect-app watches this variable to see get all the items that were included in the multi-selection: https://github.com/Vertexvis/connect-app/pull/265/files#diff-73d237b734a2c7c0749a9d783e73634e1fdba135954dcb73ad4511eb7902cb88R128. 
Connect-app doesn't know the index of the scene tree items, so it wasn't able to get this information on its own, and without it, _show/hide only selected_ was only changing the state of the scene tree items that were actually clicked on instead of the entire range.

danschultz: I think there's a few issues here:

* The `rows` property that you're filtering on doesn't include the whole list of items. The tree will lazy-load items and purge items from this list to keep memory pressure low. The result is that users may see gaps in there selection if they perform a range selection over multiple pages of data.
* There could be some large data sets that we're sending over the wire when performing scene item overrides. I think we may need some additional help on the backend here to resolve this. IIRC, @jareddellitt had some thoughts here.
* By adding a property to the component, I'd expect this field to return all the rows that are currently selected. Selection can happen in other methods besides `performMultiSelection`, e.g. `selectItem` and through the viewer. If you don't want to maintain this bookkeeping, I'd recommend instead to return the rows from `selectRange`.

danschultz:
Will be superseded by https://github.com/Vertexvis/vertex-web-sdk/pull/309. (reopen if closed by mistake)

PR #298, opened 2021-12-09
[PLAT-533] Fix scene tree scrollTo methods

jdm717:
## Summary

Corrects an issue with the `scrollToItem` and `scrollToIndex` methods where the incorrect element was being scrolled.

## Test Plan

- Test `scrollToItem` with an item that has not been expanded to (changing the scroll height of the layout)
- Test `scrollToItem` with an item that has been expanded to (not changing the scroll height of the layout)
- Test `scrollToIndex` 

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


packages/viewer/src/components/scene-tree-table-layout/scene-tree-table-layout.tsx
```diff
@@ -244,6 +256,7 @@ export class SceneTreeTableLayout {
 
   public componentDidRender(): void {
     this.layoutColumns();
+    this.updateScrollTop();
```
danschultz, line 259: Do we need to do this on every render? Seems like this'd cause problems if a user scrolls the tree and the layout rerenders.

jdm717: Currently this is conditionally updating every render only if `scrollToPosition` was called to cover for the case that the column heights have not been updated when we call `scrollTo` on the table element, but checking every render. Essentially it seems like `scrollToPosition` would need to know when `totalRows` has been updated + a render has happened before performing the `scrollTo` to avoid checking this per render. Open to other ideas though if it seems like there's a better alternative

danschultz: But once it's set, it's always going to scroll back to that position. So after `scrollToPosition` is called and the user scrolls, the next rerender will snap the scroll position back to `scrollToTop`.

Can we just run this once after the component has been initialized and the cell heights have been computed?

jdm717: Discussed offline with @danschultz some alternatives here to having the layout managing these values. Ended up going with an approach here where the `scene-tree` element will manage whether the `scene-tree-table-layout` has been re-rendered after a change to the `totalRows` property prior to telling it to scroll to a position - eecc8bd.

@danschultz let me know if you have any thoughts on this updated approach

danschultz (approved):

PR #299, opened 2021-12-10
[PLAT-571] add deviceId to measurement metadata

chrisgray-vertex:
## Summary

Add device-id to grpc metadata for measurement calls when it's available.


gbiv (approved):
LGTM
:shipit:

PR #301, opened 2021-12-15
[PLAT-581] Scene tree connection fixes

danschultz:
## Summary

Fixes an issue where the scene tree becomes unresponsive if you remove and add a new tree to the DOM. This was happening because STS only allows a single subscription per scene view, and we were not terminating the connection of the tree that was removed.

https://vertexvis.atlassian.net/browse/PLAT-581

### DIFF
--- a/packages/viewer/src/components/scene-tree/scene-tree.tsx
+++ b/packages/viewer/src/components/scene-tree/scene-tree.tsx
@@ -527,7 +527,10 @@ export class SceneTree {
     return this.controller?.fetchMetadataKeys() ?? [];
   }
 
-  protected componentWillLoad(): void {
+  /**
+   * @ignore
+   */
+  protected connectedCallback(): void {
     if (this.viewerSelector != null) {
       this.viewer = document.querySelector(this.viewerSelector) as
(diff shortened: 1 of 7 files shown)


jdm717 (approved):
LGTM

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -570,14 +578,14 @@ export class SceneTree {
 
     this.controller?.setMetadataKeys(this.metadataKeys);
 
-    this.updateLayoutElement();
+    // this.updateLayoutElement();
```
jdm717, line 581: Can this be removed?

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -570,14 +578,14 @@ export class SceneTree {
...
   }
 
   /**
    * @ignore
    */
   protected componentWillRender(): void {
-    this.updateLayoutElement();
+    // this.updateLayoutElement();
```
jdm717, line 588: Same as above

PR #302, opened 2021-12-15
[PLAT-472] Double clicking middle mouse button performs fit all

amvertex:
## Summary
We want double clicking the middle mouse button to perform a fit all.

## Test Plan
Double click the middle mouse button

## Release Notes
Double clicking the middle mouse button performs a fit all

## Possible Regressions
None

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
+++ b/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
@@ -160,7 +160,7 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
         this.handleDoubleClick(event);
       }
       this.clickCount = 0;
-    }, 250);
+    }, this.getConfig().interactions.doubleClickDelay);
 
     window.addEventListener(this.moveEvent, this.handleWindowMove);
     window.addEventListener(this.upEvent, this.handleWindowUp);
--- a/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
+++ b/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
@@ -45,7 +45,6 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
   private computedBodyStyle?: CSSStyleDeclaration;
 
   private primaryInteractionTypeChange = new EventDispatcher<void>();
-  private clickCount = 0;
 
   public constructor(
     protected downEvent: 'mousedown' | 'pointerdown',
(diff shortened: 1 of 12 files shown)


packages/viewer/src/lib/types/interactions.ts
```diff
@@ -18,10 +18,16 @@ export interface InteractionConfig {
...
+   * The amount of time between two clicks to be considered a double click.
+   */
+  doubleClickDelay: number;
 }
 
 export const defaultInteractionConfig: InteractionConfig = {
   finePointerThreshold: 1,
   coarsePointerThreshold: 3,
   interactionDelay: 75,
+  doubleClickDelay: 250,
```
amvertex, line 32: I took this number from the current implementation in manufacturing-ui. The _interactionDelay_ listed above was too short for the the double click to be triggered.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -152,6 +155,13 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
       }
     }, this.getConfig().interactions.interactionDelay);
 
+    window.setTimeout(() => {
+      if (this.clickCount === 2) {
+        this.handleDoubleClick(event);
+      }
+      this.clickCount = 0;
+    }, this.getConfig().interactions.doubleClickDelay);
```
danschultz, line 163: Can we try with the native `dblclick` event before rolling something custom? https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event

packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -310,6 +310,16 @@ export class InteractionApi {
...
   }
 
+  /**
+   * Performs a view all operation for the scene's bounding box, and requests a new image
+   * for the updated scene.
+   */
+  public async viewAll(): Promise<void> {
+    return this.transformCamera(({ camera, viewport }) => {
+      return camera.viewAll();
+    });
```
danschultz, line 320: You can simplify this to `return this.getScene().camera().viewAll().render();`. Going through a camera transform will put issue additional renders for intermediate frames, which probably isn't required here.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -200,6 +210,13 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
     this.lastMoveEvent = undefined;
   }
 
+  protected async handleDoubleClick(event: BaseEvent): Promise<void> {
+    if (event.buttons === 4 && this.interactionApi != null) {
+      this.interactionApi.beginInteraction();
```
danschultz, line 215: This can prob be removed if with the comment above. If you use this in the future, you also need to invoke `this.interactionApi.endInteraction()` to get a final frame. Otherwise, rendering might be stuck sending back intermediate frames.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
amvertex: We can't use the native dblclick event as it is only for the primary mouse button (not the middle mouse button). However, we could use the mousedown event. Do you like this approach better?

danschultz (approved):

PR #304, opened 2021-12-16
[Plat 179] Surface area with 2 surfaces selected

chrisgray-vertex:
## Summary

This adds total surface area to the measurement details when 2 surfaces are selected.  Single surface selection will come in a following PR

## Test Plan

Select 2 surfaces, ensure that the total area is displayed and is accurate, change to different length units and ensure that the labels and conversions are correct.


packages/viewer/src/lib/types/measurementUnits.ts
```diff
@@ -46,32 +46,32 @@ class MillimeterUnitConverter extends ScaledUnitConverter {
 }
 
 class CentimeterUnitConverter extends ScaledUnitConverter {
-  public constructor() {
-    super(1 / 100);
+  public constructor(power = 1) {
+    super(1 / Math.pow(10, power));
```
chrisgray-vertex, line 50: The conversion factor for centimeters wasn't correct, changed it to 10 here to fix the mm -> cm conversion for area as well as length.

gbiv (approved):
LGTM

PR #306, opened 2021-12-20
[PLAT-419] Support slots for cell content

MadisonEhlers-Vertex:
## Summary

These changes introduce support for adding a slottable content. And also a separate slot for the placeholder value if the cell value is undefined


jdm717 (approved):

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -164,7 +164,17 @@ export class SceneTreeTableCell {
               )}
             </button>
           )}
-          <div class="content">{this.displayValue()}</div>
+          {this.value && (
+            <slot>
+              <div class="content">{this.displayValue()}</div>
+            </slot>
```
danschultz, line 170: Would this be more appropriate? `<div class="content"><slot>{this.displayValue()}</slot></div>`

danschultz:
Tested locally. This is what I see:

Screen Shot 2021-12-20 at 12 16 40 PM

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -164,7 +164,17 @@ export class SceneTreeTableCell {
               )}
             </button>
           )}
-          <div class="content">{this.displayValue()}</div>
+          {this.value && (
```
danschultz, line 167: You should use `this.value != null` here. An empty string will evaluate to `false` otherwise.

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -164,7 +164,17 @@ export class SceneTreeTableCell {
...
+            <slot>
+              <div class="content">{this.displayValue()}</div>
+            </slot>
+          )}
+
+          {!this.value && (
```
danschultz, line 173: ^

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -777,7 +777,9 @@ export class SceneTree {
       layout.innerHTML = `
       <vertex-scene-tree-table-column>
         <template>
-          <vertex-scene-tree-table-cell prop:value="{{row.node.name}}" expand-toggle visibility-toggle />
+          <vertex-scene-tree-table-cell prop:value="{{row.node.name}}" expand-toggle visibility-toggle>
+            <div style="width:100%;">{{row.node.name}}</div>
+          </vertex-scene-tree-table-cell>
```
danschultz, line 782: I think we can revert this now. However, it should probably be:

```html
<vertex-scene-tree-table-cell prop:value="{{row.node.name}}" expand-toggle visibility-toggle></vertex-scene-tree-table-cell>
```

Depends on the DOCTYPE, but defaults doesn't actually support XHTML style closing tags `</>`.

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -195,9 +202,12 @@ export class SceneTreeTableCell {
   }
 
   private displayValue = (): string => {
-    return this.value == null || this.value.replace(' ', '') === ''
-      ? this.placeholder
-      : this.value;
+    const resp =
+      this.value == null || this.value.replace(' ', '') === ''
```
danschultz, line 206: Not your change, but while we're here... `this.value.replace(' ', '')` should prob be `.trim()`.

danschultz (approved):

PR #308, opened 2021-12-21
[PLAT-634] do measurement on all surface picks

chrisgray-vertex:
## Summary

Do measurement requests on each surface pick.

## Test Plan

After measurement and sdk/connect are updated, the user should be able to select a single surface and get area back for it. Then when a second surface is selected the other measurements along with total surface area should be returned.


gbiv (approved):
LGTM

PR #311, opened 2021-12-22
[plat-651] Add setEntities method on measurement controller

chrisgray-vertex:
## Summary

Currently, when a client needs to add and remove entities on the measurement model in one logical operation, a clear or remove and then an add needs to be done.  This can result in inconsistent face highlighting and extra roundtrips.  This new`setEntities` method allows removing and adding entities in a single call for more consistent and efficient results.

## Test Plan

Enable precise measurement and select 2 faces, then select a 3rd face.  The first 2 faces should clear and the 3rd should highlight.  Previous behavior would often leave no faces visibly highlighted after the 3rd pick.


mikemccarty-vertex (approved):
LGTM

PR #314, opened 2022-01-03
[PLAT-641] Add ability to set column keys on filter scene tree request; implement create alteration with a metadata query

jareddellitt:
## Summary
Add ability to set column keys on filter scene tree request; implement create alteration with a metadata query


danschultz:
BTW, any files named `[name].local.html` will be git ignored :)

danschultz (approved):

therockstorm (approved):

PR #315, opened 2022-01-04
[PLAT-645 Add a flag that indicates whether metadata should be used for filtering and pass along columns in the filter call if so

jareddellitt:
## Summary
Add a flag that indicates whether metadata should be used for filtering and pass along columns in the filter call if so

## Test Plan
When creating the scene-tree element, set the `filterOnMetadata` prop to `true` and try searching.


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -184,6 +184,13 @@ export class SceneTree {
   @Prop({ mutable: true })
   public controller?: SceneTreeController;
 
+  /**
+   * Temporary flag to indicate whether metadata should be used when filtering
+   * as opposed to just the name of the item.
+   */
+  @Prop({ mutable: true })
```
danschultz, line 191: `mutable` only needs to be set if the property is modified within the component.

danschultz (approved):
Field is temp, LGTM.

packages/viewer/src/components/scene-tree/scene-tree.tsx
jareddellitt: Noted, thanks!

PR #316, opened 2022-01-04
[PLAT-656] Support withSelected Queries

MadisonEhlers-Vertex:
## Summary

These changes introduce support for operating on selected items, like this:

```typescript
            await scene
              .items((op) => [
                op.where((q) => q.withSelected()).hide(),
              ])
              .execute();
```

You can optionally pass a color material to `withSelected` that will only include selected items with that specific color. 

## Test Plan

- Do a range query, then do the operation specified above
operate on selected items


packages/viewer/src/lib/scenes/scene.ts
```diff
@@ -188,6 +188,8 @@ export class ItemsOperationExecutor {
     const pbOperations = this.queryOperations.map((op) =>
       buildSceneOperation(op.query, op.operations)
     );
+
+    console.log('pbOperations: ', pbOperations);
```
danschultz, line 192: Remove?

packages/viewer/src/lib/scenes/queries.ts
```diff
@@ -94,6 +100,10 @@ export class RootQuery implements ItemQuery<SingleQuery> {
   public withMetadata(filter: string, keys: string[]): MetadataQuery {
     return new MetadataQuery(filter, keys);
   }
+
+  public withSelected(color?: ColorMaterial): AllSelectedQuery {
```
danschultz, line 104: Accepting a `color` here is mixing the concept of queries and operations.

MadisonEhlers-Vertex: Not really - the user could select multiple items with different materials - This is a conditional paramter that says, only show the selected items with this color.

MadisonEhlers-Vertex:
@danschultz and I discussed and for now we'd like to keep the material on selected off.

danschultz (approved):

PR #320, opened 2022-01-11
[PLAT-542] Add method to query entity types in feature map

danschultz:
## Summary

Adds a method `FeatureMap.getEntityType` to query the feature map for the type of entity at a given point.

https://vertexvis.atlassian.net/browse/PLAT-542

### DIFF
--- a/packages/viewer/src/lib/types/entities.ts
+++ b/packages/viewer/src/lib/types/entities.ts
@@ -1,0 +1,37 @@
+// The source of these values are from:
+// https://github.com/Vertexvis/rendering-worker-service/blob/fc1c864d1da4b5eb12794478dc720fa046412ee2/src/embree.isph#L163
+
+/**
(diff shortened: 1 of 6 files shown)


packages/viewer/src/lib/types/entities.ts
```diff
@@ -0,0 +1,8 @@
+export enum EntityType {
```
mikemccarty-vertex, line 1: Might be worth adding a comment with the permalink to the code that generates these values:  https://github.com/Vertexvis/rendering-worker-service/blob/fc1c864d1da4b5eb12794478dc720fa046412ee2/src/embree.isph#L163

danschultz: Updated

mikemccarty-vertex (approved):
LGTM

PR #321, opened 2022-01-14
[PLAT-763] Markup improvements

jdm717:
## Summary

Makes a handful of changes to the `<vertex-viewer-markup>` element:
- Modifies the way selection of markup is handled. Previously we would select markup on a `pointerdown` event, which could cause difficult when trying to place a markup close to another. This is updated to select on `pointerup` only if `pointermove`s have not gone more than 2 pixels away from the down position
- Modifies how we start new markup. Previously this watched for a `pointerdown` on a `create-overlay` div, and wouldn't allow starting markup when the down occurred on another markup. This is updated to watch for `pointerdown` events on the window level, and start a markup as long as the `pointerdown` event occurred on a markup not in `edit` mode.
- Modifies the flow after completing the creation of a markup. Previously when an `editEnd` event fired from the markup, we would immediately reset the element within the `<vertex-viewer-markup-tool>`, then add it to the DOM underneath the `<vertex-viewer-markup>` element rather than the tool element which would cause a flicker. This is updated to watch for a new `viewRendered` event from the markup before resetting the tool's markup element to prevent the flicker
- Adds event listeners to `preventDefault` on `touchstart` events to prevent zooming/dragging the window on mobile
- Makes some small adjustments to when a markup is considered "valid" to prevent `NaN` values on arrows and "single point" arrows

## Test Plan

- Test creating markup on desktop (all three types)
- Test creating markup on mobile (all three types)
- Test markup selection
- Test markup editing

## Release Notes

N/A

## Possible Regressions

Markup, particularly selection + editing

## Dependencies

N/A


packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.tsx
```diff
@@ -95,11 +96,11 @@ export class ViewerMarkupCircle {
...
 
   /**
-   * An event that is dispatched when the user cancels editing of the
-   * markup.
+   * An event that is dispatched when this markup element is in view
+   * mode (`this.mode === ""`), and it completes a rerender.
    */
   @Event({ bubbles: true })
-  public editCancel!: EventEmitter<void>;
+  public viewRendered!: EventEmitter<void>;
```
danschultz, line 103: How will this be used?

amvertex (commented):
LGTM!

packages/viewer/src/components/viewer-markup-circle/viewer-markup-circle.tsx
jdm717: This is being used to know when the newly created markup has been appended to the `<vertex-viewer-markup>` element and completed a render, so we can reset the "draft" markup in the `<vertex-viewer-markup-tool>`. This fixes the "flicker" you see when creating a markup currently

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
```diff
@@ -206,55 +232,59 @@ export class ViewerMarkupArrow {
...
+                    x1={screenStart.x}
+                    y1={screenStart.y}
+                    x2={screenEnd.x}
+                    y2={screenEnd.y}
+                  />
+                )}
+              </g>
+            </svg>
+            {this.mode === 'edit' && (
+              <BoundingBox1d
```
danschultz, line 269: `BoundingBox2d`?

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -110,6 +112,17 @@ export class ViewerMarkup {
...
   private hostEl!: HTMLElement;
 
+  @State()
+  private toSelectMarkupId?: string;
+
+  @State()
+  private pointerDownPosition?: Point.Point;
+
+  protected disconnectedCallback(): void {
+    window.removeEventListener('pointermove', this.handlePointerUp);
```
danschultz, line 122: `'pointerup'`

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -349,6 +369,46 @@ export class ViewerMarkup {
...
+          Point.create(event.clientX, event.clientY)
+        ) <= 2
+      ) {
+        this.selectedMarkupId = undefined;
+      }
+      this.toSelectMarkupId = undefined;
+      this.pointerDownPosition = undefined;
+    }
+
+    window.removeEventListener('pointermove', this.handlePointerUp);
```
danschultz, line 408: ^

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
jdm717: This is just represented as a line along the arrow, so it is technically 1D - though "BoundingBox" does feel like a bit of an incorrect term for that. Nothing immediately comes to mind as a good replacement though

danschultz (approved):

PR #322, opened 2022-01-14
[PLAT-688] Scene tree cursor in Safari

amvertex:
## Summary
In Safari, the text in the scene tree was selectable and had the text selection cursor.  After this PR, the behavior in Safari matches other browsers.

Before:

SafariSceneTreeBefore

After:
SafariSceneTreeAfter

## Test Plan
Look at the scene tree in Safari

## Release Notes
None

## Possible Regressions
Scene tree

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/scene-tree/scene-tree.css
+++ b/packages/viewer/src/components/scene-tree/scene-tree.css
@@ -10,6 +10,8 @@
   width: 300px;
   height: 100%;
   user-select: none;
+  -webkit-user-select: none;
+  cursor: default;
   display: flex;
   flex-direction: column;
 }


packages/viewer/src/components/scene-tree/scene-tree.css
```diff
@@ -10,6 +10,8 @@
   width: 300px;
   height: 100%;
   user-select: none;
+  -webkit-user-select: none;
```
danschultz, line 13: We use `user-select` in a few places. I think we should investigate adding [postcss/autoprefixer](https://www.npmjs.com/package/@stencil/postcss) to our stencil builds, so we don't have to manually manage prefixes.

MadisonEhlers-Vertex (approved):

PR #331, opened 2022-01-31
[PLAT-846] Point-to-point axis reference lines

danschultz:
## Summary

Adds option to display axis reference lines for point-to-point measurements.

Screen Shot 2022-01-31 at 11 35 24 AM

https://vertexvis.atlassian.net/browse/PLAT-846


packages/viewer/src/components/viewer-measurement-distance/viewer-measurement-distance.tsx
```diff
@@ -132,6 +136,13 @@ export class ViewerMeasurementDistance {
   @Prop({ mutable: true })
   public distance?: number;
 
+  /**
+   * Enables the display of axis reference lines between the start and end
+   * point.
+   */
+  @Prop()
+  public showAxisReferenceLines = false;
```
jdm717, line 144: Would there be any value in being able to specify the order the reference lines are displayed? Just testing around I've noticed that it can be a bit hard to follow the reference lines since they're always X <-> Y <-> Z from the second point. Wondering if something like Y <-> X <-> Z might be easier to follow (or if I'm just not used to this representation)

jdm717 (approved):
This looks great! Left one thought related to configuring the order in which the reference lines appear, but that also seems like it'd make sense as follow-up if we wanted to support it

PR #332, opened 2022-02-01
[PLAT-847] Single distance measurement

danschultz:
## Summary

Update the distance measurement component to better handle single distance measurements. In particular, it allows for starting a new measurement and editing of an existing measurement, if the component is in `replace` mode. It also exposes a `MeasurementModel` which can be passed to the details component to display results.

https://user-images.githubusercontent.com/37429/152048487-0c4b8fcf-d5e0-406d-bd79-038ffc14bf96.mov

https://vertexvis.atlassian.net/browse/PLAT-847


jdm717 (approved):
LGTM!

PR #344, opened 2022-02-18
[PLAT-579] Alternate resize renderer

jdm717:
## Summary

Updates the approach for handling resizes of the canvas element to remove the "flashing" effect.

Before:

https://user-images.githubusercontent.com/5252480/154594285-3375b027-abc6-4882-91ea-b0016a45ad93.mov

After:

https://user-images.githubusercontent.com/5252480/154594297-28f8c457-b239-44b3-9aa8-540e743cc4fa.mov

## Test Plan

- Test general rendering
- Test resizing

## Release Notes

N/A

## Possible Regressions

Rendering, particularly around resizing since this removes the existing flow updating dimensions of the canvas

## Dependencies

N/A


amvertex (approved):

packages/viewer/src/lib/rendering/canvas.ts
```diff
@@ -121,6 +129,38 @@ export function createCanvasRenderer(): CanvasRenderer {
...
+      drawImage(image, { ...data, canvas: ctx });
+      data.beforeDraw?.();
+      data.canvas.drawImage(hiddenCanvas, 0, 0);
+    }
+
+    image.dispose();
+    return data.frame;
+  };
+}
+
```
danschultz, line 163: I think part of the jumping has to do with us using a flex layout on the canvas container. This is also causing an issue where the viewer wouldn't resize to a smaller dimension when placed in a flex container. Anyways, seems like this hidden canvas isn't needed (only did some quick testing) when I replace the CSS for `.canvas-container` with: 

```css
.canvas-container {
  position: absolute;
  top: 0;
  left: 0;
  right: 0;
  bottom: 0;
  background: var(--image-background, var(--viewer-background, #ffffff));
}
```

PR #349, opened 2022-03-15
[PLAT-269] Basic orthographic camera

jdm717:
## Summary

Adds basic support for orthographic cameras. Currently, this is done by changing the `cameraType` on the `<vertex-viewer>` element from `'perspective'` to `'orthographic'`. This converts the current perspective camera to an orthographic representation.

Future work:
- Update interactions to support orthographic cameras
- Update the view cube to show an accurate representation of the orthographic camera

## Test Plan

- Test general perspective camera interactions (a lot of the logic was changed to include orthographic support)
- Test changing the camera from `'perspective'` to `'orthographic'` and vice-versa

## Release Notes

N/A

## Possible Regressions

Camera interactions

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/272


packages/viewer/src/lib/types/frameCamera.ts
```diff
@@ -1,15 +1,113 @@
...
 }
+
+export function createOrthographic(
+  data: Partial<OrthographicFrameCamera> = {}
+): OrthographicFrameCamera {
+  return {
+    viewVector: data.viewVector || Vector3.back(),
+    lookAt: data.lookAt || Vector3.origin(),
+    up: data.up || Vector3.up(),
+    fovHeight: data.fovHeight ?? 45,
```
mikemccarty-vertex, line 49: The default should probably be `1.0` just so it isn't confused with "FOV angle".

packages/viewer/src/lib/types/frameCamera.ts
```diff
@@ -1,15 +1,113 @@
...
     lookAt: data.lookAt || Vector3.origin(),
     up: data.up || Vector3.up(),
   };
 }
+
+export function createOrthographic(
+  data: Partial<OrthographicFrameCamera> = {}
+): OrthographicFrameCamera {
+  return {
+    viewVector: data.viewVector || Vector3.back(),
```
mikemccarty-vertex, line 46: Why is this `back()` rather than `forward()`?

(Also, I don't understand this notation -- what does `||` mean in this context?  Related to `??`?)

packages/viewer/src/lib/types/frameCamera.ts
```diff
@@ -1,15 +1,113 @@
...
+  data: PerspectiveFrameCamera
+): OrthographicFrameCamera {
+  const viewVector = Vector3.subtract(data.lookAt, data.position);
+
+  return {
+    viewVector,
+    up: data.up,
+    lookAt: data.lookAt,
+    fovHeight:
+      2 * Vector3.magnitude(viewVector) * Math.tan(Angle.toRadians(22.5)),
```
mikemccarty-vertex, line 63: Can we somehow parameterize the FOV angle so it isn't hardcoded everywhere?  Someday we may want to change the FOV angle dynamically and hard-coding it everywhere (especially in coded form like `22.5`) is going to make that difficult.

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -9,14 +9,41 @@ import {
...
+export function distanceToVectorAlongViewVec(
+  vector: Vector3.Vector3,
+  camera: FrameCamera.FrameCamera
+): number {
+  const viewVector = isOrthographicFrameCamera(camera)
+    ? camera.viewVector
+    : Vector3.subtract(camera.lookAt, camera.position);
+
+  return (
+    Math.abs(Vector3.dot(viewVector, vector)) / Vector3.magnitude(viewVector)
```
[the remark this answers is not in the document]
jdm717: Yeah, that's fair. This was copied from [here](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/lib/scenes/camera.ts#L148-L154) and just pulled the `Math.abs()` that we were already using, but I think it makes more sense for the `rotateCameraAtPoint(...)` method that uses the `distanceToBoundingBoxCenter()` result to take the absolute value since it's the one that cares. I'll update this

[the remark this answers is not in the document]
mikemccarty-vertex: you might consider renaming it to `signedDist...` to highlight the fact that negative values are a possibliity.

packages/viewer/src/lib/types/clippingPlanes.ts
```diff
@@ -1,29 +1,26 @@
...
-        cameraToCenter
-      )
-    ) / Vector3.magnitude(Vector3.subtract(camera.lookAt, camera.position));
+  const distanceToCenterAlongViewVec = distanceToVectorAlongViewVec(
+    cameraToCenter,
+    camera
+  );
   const radius = 1.1 * Vector3.magnitude(centerToBoundingPlane);
   let far = distanceToCenterAlongViewVec + radius;
   let near = far * 0.01;
```
mikemccarty-vertex, line 26: What are these used for?  I'm a little concerned that this logic does not match up with RCL's version....

jdm717: Yeah, I don't know that this logic has been revisited any time recently. I think the only logic around the `near` and `far` clipping planes is depth buffer related. We do have https://vertexvis.atlassian.net/browse/PLAT-1151 which appears to be a regression in the "rotate-point" interaction to rotate around a point computed from the depth buffer, so I wonder if that's related? I can take a look at updating the logic with this PR at least.

This logic was added to the client thinking along the same lines as the camera matrices - essentially that we can compute the `near` and `far` plane values on the client, so omit it from the payload to reduce size. Given the amount of reliance we have on those values now with depth buffers and the potential for them to change though, maybe worth revisiting + creating a follow-up ticket to remove this calculation

mikemccarty-vertex: If this is being used to map values into the depth buffer, then the logic absolutely _must_ match.  

Overall, we need to decide how these values get calculated -- ideally, it is done in exactly one place to avoid mismatches but maybe that's not practical.

mikemccarty-vertex: If this is being used to map values into the depth buffer, then the logic absolutely _must_ match.  

Overall, we need to decide how these values get calculated -- ideally, it is done in exactly one place to avoid mismatches but maybe that's not practical.

jdm717: Created https://vertexvis.atlassian.net/browse/PLAT-1187 and put it in grooming to discuss this, but also updated the near/far computation with https://github.com/Vertexvis/vertex-web-sdk/pull/349/commits/dcc664815a0376579629810fa88a7e8cfc6fc026

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -162,14 +138,20 @@ export abstract class Camera {
...
+    const { position, viewVector } = withPositionAndViewVector(this.data);
+
+    const boundingBoxCenter = BoundingBox.center(
+      boundingBox ?? this.boundingBox
+    );
+    const cameraToCenter = Vector3.subtract(position, boundingBoxCenter);
 
-    return distanceToVectorAlongViewVec(cameraToCenter, this.data);
+    return (
+      Math.abs(Vector3.dot(viewVector, cameraToCenter)) /
```
mikemccarty-vertex, line 152: The `abs()` is still here!  :)

packages/viewer/src/lib/types/frameCamera.ts
```diff
@@ -60,25 +79,24 @@ export function toOrthographic(
...
 
 export function toPerspective(
   data: OrthographicFrameCamera
 ): PerspectiveFrameCamera {
-  const expectedMagnitude =
-    data.fovHeight / (2 * Math.tan(Angle.toRadians(22.5)));
-  const receivedMagnitude = Vector3.magnitude(data.viewVector);
-  const magnitudeScale = expectedMagnitude / receivedMagnitude;
+  const fovY = Angle.toDegrees(
+    Math.atan(data.fovHeight / (2 * Vector3.magnitude(data.viewVector))) * 2
```
mikemccarty-vertex, line 92: I think you want to leave this the way you had it before (`fovY==45`, solve for view-vector length) -- is there a reason you changed it?

Maybe `45.0` should be passed in as a parameter...

mikemccarty-vertex (commented):
I think we're close...

packages/viewer/src/lib/scenes/camera.ts
jdm717: Shoot, I missed removing it here and added it to the interaction hander... I'll fix that

packages/viewer/src/lib/types/frameCamera.ts
jdm717: Yeah, based on my testing the `magnitudeScale` was always 1 (or extremely close - like 1.000034), so it wasn't clear if we needed to keep that around since we maintain the magnitude of the `viewVector` when we convert to orthographic. As for passing in the `fovY`, it seems like we may run into problems if we start from an orthographic camera (since that fovY value would be unknown at that point). Would there be a way to solve for both the magnitude of the `viewVector` and the `fovY` given a starting orthographic camera with our current representation?

mikemccarty-vertex: I see what you're saying... There might be something clever you could do here but I can't see it at the moment.

In general, you would not expect `fovY` to change often, but a client may have a specific idea of what it ought to be (e.g. at PLC, the standard was `56`).  That's the level of flexibility we should be supporting -- once set, all individual cameras are very likely going to use the same setting.  I don't know how the SDK handles the concept of "application defaults" but this feels like one of those situations.

jdm717: Yeah, I think that makes sense. For the purposes of this addition, I may lean toward not including a configuration option or property (since it won't do anything until we wire up the backend). For now I'll add an optional `fovY` argument for that `toPerspective()` method that just defaults to `45` with a TODO to revisit once we allow changing that value

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -148,7 +148,7 @@ export abstract class Camera {
     );
     const cameraToCenter = Vector3.subtract(position, boundingBoxCenter);
 
-    return (
+    return -(
       Vector3.dot(viewVector, cameraToCenter) / Vector3.magnitude(viewVector)
     );
```
jdm717, line 153: Ended up flipping this value based on some testing. Basically what I found is that a camera defined as:
```typescript
{
  lookAt: { x: 0, y: 0, z: 0 },
  up: { x: 0, y: 1, z: 0 },
  position: { x: 0, y: 0, z: 5 },
}
```
looking at a bounding box centered at the origin, would return a distance of `-5` for a distance of 5 toward the `lookAt`. Moving the `lookAt` to be `{ x: 0, y: 0, z: 10 }` (effectively 180 turn), the distance was `5`, but away from the `lookAt`. There may be a better alternative than just flipping the value however, so I'm open to other ideas

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -148,9 +148,8 @@ export abstract class Camera {
...
     const cameraToCenter = Vector3.subtract(position, boundingBoxCenter);
```
mikemccarty-vertex, line 149: If this line were:
```
const cameraToCenter = Vector3.subtract(boundingBoxCenter, position);
```

... then I think you could get rid of the negation below.

PR #351, opened 2022-03-23
[PLAT-270] Basic orthographic interactions

jdm717:
## Summary

Adds some changes to support panning and zooming for orthographic cameras, as well as fixing a few issues with rotation, standard views, and fit all.

https://user-images.githubusercontent.com/5252480/159811507-7ae5deb7-4dd0-4380-b541-40ed4c1eb4d9.mov

## Test Plan

- Test interactions with orthographic cameras

## Release Notes

N/A

## Possible Regressions

N/A (interactions for orthographic were not expected to work previously)

## Dependencies

N/A


packages/geometry/src/matrix4.ts
```diff
@@ -384,7 +384,7 @@ export function makeOrthographic(
 
   const x = (right + left) * w;
   const y = (top + bottom) * h;
-  const z = (far + near) / d;
+  const z = (far + near) * d;
```
mikemccarty-vertex, line 387: That would make a difference...

mikemccarty-vertex (approved):
LGTM

packages/geometry/src/matrix4.ts
jdm717: Yeah... definitely explained a lot of issues I had been seeing initially...

PR #353, opened 2022-03-24
[PLAT-267] Move camera type checks

jdm717:
## Summary

Adds an `isOrthographic` and `isPerspective` method to the `FrameCameraBase` to avoid having to do `instanceof` checks. Additionally, this corrects an issue where the connect-app was unable to use the helpers for checking the type of a camera.

## Test Plan

- Test changing between perspective and orthographic cameras

## Release Notes

N/A

## Possible Regressions

Changing between orthographic and perspective cameras

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -1131,8 +1131,7 @@ export class Viewer {
       );
     }
 
-    return this.frame == null ||
-      FrameCamera.isPerspectiveFrameCamera(this.frame.scene.camera)
+    return this.frame == null || this.frame.scene.camera.isPerspective()
       ? new InteractionApiPerspective(
           this.stream,
           this.stateMap.cursorManager,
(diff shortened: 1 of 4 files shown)


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -1246,11 +1245,11 @@ export class Viewer {
   private updateInteractionApi(previousFrame?: Frame): void {
     if (previousFrame != null && this.frame != null) {
       const hasChangedFromPerspective =
-        FrameCamera.isPerspectiveFrameCamera(previousFrame.scene.camera) &&
-        FrameCamera.isPerspectiveFrameCamera(this.frame.scene.camera);
```
MadisonEhlers-Vertex, line 1250: Was the previous check wrong? It was looking for perspective in both cases?

MadisonEhlers-Vertex (approved):

packages/viewer/src/components/viewer/viewer.tsx
jdm717: That it was, fixed this in #351 but figured it made sense to fix with this change as well

MadisonEhlers-Vertex (approved):

PR #354, opened 2022-03-25
[PLAT-271] Orthographic depth buffer support

jdm717:
## Summary

Adds support for depth buffer interaction in orthographic mode. Specifically, this impacts a handful of features:
- Rotate around tap point in orthographic should work as expected now
- Zoom to point is now using the value from the depth buffer to determine the lookAt changes
- Point to point measurement is now properly supported
- `vertex-viewer-dom-element`s in `3d` mode will now be displayed (perspective based still, created https://vertexvis.atlassian.net/browse/PLAT-1244 to capture rendering those pins orthographically)

## Test Plan

- Test rotate around tap point in orthographic
- Test zoom to point in orthographic
- Test point to point measurement (perspective/orthographic)
- Test pins in (perspective/orthographic)

## Release Notes

N/A

## Possible Regressions

P2P measurement, pins, orthographic interactions

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-dom-renderer/renderer3d.tsx
+++ b/packages/viewer/src/components/viewer-dom-renderer/renderer3d.tsx
@@ -17,7 +17,9 @@ export const Renderer3d: FunctionalComponent<Props> = (
   children
 ) => {
   const pMatrix = Matrix4.toObject(camera.projectionMatrix);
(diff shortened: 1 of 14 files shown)


MadisonEhlers-Vertex (approved):
looks good to me - we can follow up with any other issues that come up

PR #356, opened 2022-03-29
[PLAT-1173] Improve mobile scene tree interactions

amvertex:
## Summary
Previously, users were only able to hide a visible item in the scene tree if they were first hovering over the item, making it nearly impossible to do so on mobile. Thus, this changes the visibility button to always be present and uses a blank icon when the scene item is visible and the user isn't hovering over it.

Further, users were also unable to scroll the scene tree on mobile without selecting parts. Thus, this changes the part selection logic to watch for a large change in the y position of the pointer, and does not select the item in this case.

## Test Plan
Interact with the scene tree on mobile: change the visibility of items, select items, and scroll the scene tree

## Release Notes
None

## Possible Regressions
Mobile scene tree interactions

## Dependencies
None


packages/viewer/src/components/viewer-measurement-details/viewer-measurement-details-results.tsx
```diff
@@ -30,7 +30,7 @@ export const MinimumDistanceResultEntry: FunctionalComponent<
   return (
     <Fragment>
       <MeasurementDetailsEntry
-        label="Min Dist"
+        label="Dist"
```
amvertex, line 33: https://vertexvis.atlassian.net/browse/PLAT-1042

packages/viewer/src/lib/interactions/multiTouchInteractionHandler.ts
```diff
@@ -53,7 +53,7 @@ export abstract class MultiTouchInteractionHandler
       this.interactionApi?.panCameraByDelta(delta);
 
       // Setting a minimum angle to prevent wobbling
-      if (Math.abs(angle) > 0.2) {
+      if (Math.abs(angle) > 0.5) {
```
amvertex, line 56: https://vertexvis.atlassian.net/browse/PLAT-1174

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.css
```diff
@@ -117,3 +110,6 @@
 .icon-partial {
   background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16'%3E%3Cpath d='M4.12 11.46A5.62 5.62 0 011.52 8 5.57 5.57 0 013.9 4.67 7.52 7.52 0 018 3.5a7.7 7.7 0 013.33.75l.74-.74A8.67 8.67 0 008 2.5a8.53 8.53 0 00-4.65 1.33A6.57 6.57 0 00.51 7.89v.22a6.54 6.54 0 002.88 4.08z' fill='%23444'/%3E%3Cpath d='M8 5a3 3 0 00-3 3 3 3 0 00.69 1.89l4.2-4.2A3 3 0 008 5zM5.88 5.88zM9.2 12.41a.51.51 0 00-.42.57.5.5 0 00.56.42.5.5 0 00.43-.57.51.51 0 00-.57-.42zM6.77 12.41a.5.5 0 00-.57.42.49.49 0 00.41.57.5.5 0 10.15-1zM14.48 5.61a.5.5 0 00-.7-.12.49.49 0 00-.12.69.5.5 0 00.82-.57zM11.5 11.68a.5.5 0 00-.22.67.51.51 0 00.68.22.5.5 0 00-.46-.89zM15.07 7.77a.52.52 0 00-.62.35.51.51 0 00.35.62.5.5 0 00.61-.35.52.52 0 00-.34-.62zM13.36 10.21a.5.5 0 10.76.64.49.49 0 00-.06-.7.51.51 0 00-.7.06z' fill='%23444'/%3E%3C/svg%3E");
 }
+
+.icon-blank {
+}
```
danschultz, line 115: Empty CSS selector. Also missing file new line.

danschultz:
👀

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -216,6 +226,23 @@ export class SceneTreeTableCell {
       !event.defaultPrevented &&
       event.button === 0 &&
       !this.interactionsDisabled
+    ) {
+      this.beginningPointerPosition = event.clientY;
+      window.addEventListener('pointerup', this.handleCellPointerUp);
```
danschultz, line 231: Instead of tracking the original position on `pointerdown` and having a separate `pointerup` handler, I'd suggest that we ditch the `pointerdown` altogether and just register a `pointerup` event. Note, that there's other logic in this component for calling `preventDefault` on some buttons, so those would need to be changed over to `pointerup`.

Also I think it's generally better if we relied on scroll events vs inferring scrolling through pointer positions. You could add something like this to the `scene-tree-table-layout` component to track if scrolling is currently happening, then setting an internal prop on the component to inform it that scrolling is happening and interactions should be disabled.

```ts
  private isScrolling = false;
  private scrollTimer: number | undefined;

  private handleScroll = (): void => {
    this.isScrolling = true;

    window.clearTimeout(this.scrollTimer);
    window.setTimeout(() => {
      this.isScrolling = false;
    }, 100);
  };
```

This could probably be simplified if we moved the selection/expansion/visibility calls to the layout component itself.

Lastly, I'd recommend that we not show hover states when on mobile. You can use `window.matchMedia("(pointer: coarse)").matches` to know if the user is using a touch device.

danschultz (commented):
> Thus, this changes the visibility button to always be present and uses a blank icon when the scene item is visible

I don't see a visibility icon unless I tap on an item.

danschultz (approved):

PR #360, opened 2022-03-31
[PLAT-1265] Correct initial orthographic state

jdm717:
## Summary

Corrects an issue where loading a scene/scene-view-state that has an orthographic camera would not update the interaction api properly (due to there not being an initial perspective frame to compare to). 

## Test Plan

- Test the `rotate-point` interaction when the model is not centered
  - The model should rotate around the center point of the bounding box
- Test loading a scene-view-state captured in orthographic mode then interacting with the model (with zoom/pan)
  -  The interactions should behave as expected

## Release Notes

N/A

## Possible Regressions

Rotate point interaction in orthographic

## Dependencies

N/A


mikemccarty-vertex (approved):
LGTM

mikemccarty-vertex (approved):

MadisonEhlers-Vertex (approved):

PR #364, opened 2022-04-12
[PLAT-1307] Pin label state improvements

jdm717:
## Summary

Makes a handful of improvements to the text pin label behavior:
- Dragging the label is now based off relative difference from the `pointerdown` event versus the label anchor, and will stay under the cursor when dragging from somewhere in the middle of the label
- Styles have been updated to be more consistent between the "readonly" state and the focused state
- The `input` element has been changed to a `textarea` along with auto resizing behavior to support multiline text content
- A max width + height have been added
- The labels are now constrained to the width and height of the viewer such that typing content will not go off-screen and shift the location of the overlay (and shift the 3D pin anchors)
- A `labelFocused` and a `labelBlurred` event have been added

## Test Plan

- Test text pins generally (single line + multiline)
- Test moving labels around starting from the center/end of the label
- Test typing content close to the edges of the viewer window
- Test the max width and height with large amounts of content

## Release Notes

N/A

## Possible Regressions

Text pins behavior, could be an impact due to the introduction of a `textarea`

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-pin-label/viewer-pin-label.css
+++ b/packages/viewer/src/components/viewer-pin-label/viewer-pin-label.css
@@ -3,38 +3,56 @@
    * @prop --viewer-annotations-pin-label-background-color: A CSS color that
    * specifies the color of the label background
    */
-   --viewer-annotations-pin-label-background-color: var(--blue-200);
+  --viewer-annotations-pin-label-background-color: var(--blue-200);
 
   /**
-   * @prop --viewer-annotations-pin-label-background-color: A CSS color that
-   * specifies the color of the label background
+   * @prop --viewer-annotations-pin-label-border-color: A CSS color that
+   * specifies the color of the label's border
    */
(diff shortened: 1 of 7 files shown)


packages/viewer/src/components/viewer-pin-label/viewer-pin-label.css
```diff
@@ -44,15 +80,76 @@
...
+  border-color: var(--viewer-annotations-pin-label-focused-border-color);
+}
+
+.pin-input-drag-target {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 1;
```
MadisonEhlers-Vertex, line 117: Is the z-index here required?

MadisonEhlers-Vertex (commented):
Tested & reviewed locally and this looks really good. Can give it a +1 when build is passing.

MadisonEhlers-Vertex (approved):

PR #368, opened 2022-04-22
[PLAT-813] Return number of results in metadata search

amvertex:
## Summary
We want to show the user the number of results from a metadata search in the scene tree.

## Test Plan
Will be tested with connect-app PR

## Release Notes
None

## Possible Regressions
None

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/scene-tree/lib/controller.ts
+++ b/packages/viewer/src/components/scene-tree/lib/controller.ts
@@ -534,8 +534,6 @@ export class SceneTreeController {
         ...this.state,
         numberOfRowsWithFilterHit: numberOfResults,
       });
-
-      return numberOfResults;
     });
   }
 
--- a/packages/viewer/src/components/scene-tree/lib/controller.ts
+++ b/packages/viewer/src/components/scene-tree/lib/controller.ts
@@ -123,17 +123,6 @@ export class SceneTreeController {
   public onStateChange = new EventDispatcher<SceneTreeState>();
 
   /**
-   * Registers an event listener that will be invoked when the number of results
-   * of a search in the scene tree changes
-   *
-   * @param listener The listener to add.
-   * @returns A disposable that can be used to remove the listener.
-   */
-  public onNumberOfRowsWithFilterHitChanged(listener: Listener<number | undefined>): Disposable {
-    return this.numberOfRowsWithFilterHitChanged.on(listener);
-  }
-
-  /**
    * A dispatcher that emits an event whenever the number of search results has changed.
    */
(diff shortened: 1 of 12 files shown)


packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -756,7 +780,23 @@ export class SceneTreeController {
         );
         const rows = [...start, ...fetchedRows, ...end, ...fill];
 
-        this.updateState({ ...this.state, totalRows, rows });
+        const rowsWithFilterHit = rows.filter((row) => row?.node?.filterHit);
```
MadisonEhlers-Vertex, line 783: Will this work with pagination?

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -121,8 +122,15 @@ export class SceneTreeController {
    */
   public onStateChange = new EventDispatcher<SceneTreeState>();
 
+  /**
+   * A dispatcher that emits an event whenever the number of search results has changed.
+   */
+  public numberOfRowsWithFilterHitChanged = new EventDispatcher<number>();
```
danschultz, line 128: Looks like this value is contained within state. Can we just rely on the `onStateChange` dispatcher?

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -496,14 +515,24 @@ export class SceneTreeController {
...
+          const req = new FilterRequest();
+          req.setFilter(term);
+          req.setFullTree((options.includeCollapsed ?? true) === true);
+          if (options.columns) req.setColumnsKeysList(options.columns);
+
+          this.client.filter(req, metadata, handler);
+        }
+      );
+
+      const { numberOfResults } = res.toObject();
```
MadisonEhlers-Vertex, line 531: Can we add a test for this? Otherwise, looks good to me.

amvertex: Good idea 👍

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -41,6 +42,7 @@ export type JwtProvider = () => string | undefined;
 
 export interface SceneTreeState {
   totalRows: number;
+  numberOfRowsWithFilterHit: number;
```
MadisonEhlers-Vertex, line 45: Can this match the naming convention of `totalRows`? Maybe `totalFilteredRows`?

MadisonEhlers-Vertex (approved):

PR #373, opened 2022-05-04
[PLAT-1035] Improved Scene Tree error states

MadisonEhlers-Vertex:
## Summary

These changes introduce some improvements to the messaging & retry-ability of a non loaded scene tree. Below are two things I think might be happening, both of them are error states, but the second could be a symptom of the first. Since this has been historically hard to reproduce, I am proposing to add custom error messages in either of these two conditions. 

1. Upon investigation, I found that when using the `viewer-selector`, the scene tree does not try to look up the viewer again in the dom. A potential longer term fix here could be to check if the viewer-selector is not null, and the viewer _is_ null, and re-assign that on a post-render event. For now I'd prefer to gather additional information, as well as display to the user what's really happening. This case specifically happens when the scene tree loads, but the reference to the viewer is either incorrect, or not yet in the dom. I added a unique error message for this case, and it should only show if the scene tree has registered that it has loaded, but the viewer reference is still null. Here is an example of what I've seen a few times, and the proposed error state:

https://user-images.githubusercontent.com/49169871/166744515-261fc736-602b-4e03-96e0-e01aff7ff576.mp4

2. Additionally, I found some cases where the scene-tree controller was in a `disconnected` state, but the viewer was not present. This, too, I believe we could remedy but for now I'd prefer to show a message to the user with a basic "retry" similar to the scene tree 'failure' conditions. This was not simple to reproduce, but I believe I saw it one time. This could be a symptom of the first issue, but I'd like this to be captured in case there is some scenario that the viewer is not null, but a connection to the controller has never been initiated.

### DIFF
--- a/packages/viewer/src/components/scene-tree/lib/errors.ts
+++ b/packages/viewer/src/components/scene-tree/lib/errors.ts
@@ -1,7 +1,7 @@
 export enum SceneTreeErrorCode {
   UNKNOWN = 0,
   SCENE_TREE_DISABLED = 1,
-  UNINITIALIZED_VIEWER = 2,
+  MISSING_VIEWER = 2,
   DISCONNECTED = 3,
 }
 
@@ -22,7 +22,7 @@ function getSceneTreeErrorMessage(code: SceneTreeErrorCode): string {
       return 'An unknown error occurred.';
     case SceneTreeErrorCode.SCENE_TREE_DISABLED:
       return 'The tree for this scene is not enabled. Enable the tree for this scene to interact with the tree.';
-    case SceneTreeErrorCode.UNINITIALIZED_VIEWER:
+    case SceneTreeErrorCode.MISSING_VIEWER:
       return 'Could not find reference to the viewer';
     case SceneTreeErrorCode.DISCONNECTED:
       return 'Disconnected from server.';
--- a/packages/viewer/src/components/scene-tree/lib/errors.ts
+++ b/packages/viewer/src/components/scene-tree/lib/errors.ts
@@ -1,6 +1,8 @@
 export enum SceneTreeErrorCode {
   UNKNOWN = 0,
   SCENE_TREE_DISABLED = 1,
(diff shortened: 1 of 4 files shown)


danschultz:
👀

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -709,10 +722,20 @@ export class SceneTree {
   private connectToViewer(): void {
     this.stateMap.viewerDisposable?.dispose();
 
+    if (this.viewer == null && this.viewerSelector != null) {
+      if (this.viewerSelector != null) {
```
MadisonEhlers-Vertex, line 726: nit: this can be removed

packages/viewer/src/components/scene-tree/lib/errors.ts
```diff
@@ -1,6 +1,8 @@
 export enum SceneTreeErrorCode {
   UNKNOWN = 0,
   SCENE_TREE_DISABLED = 1,
+  UNINITIALIZED_VIEWER = 2,
```
danschultz, line 4: `MISSING_VIEWER` is prob more appropriate.

packages/viewer/src/components/scene-tree/lib/errors.ts
```diff
@@ -20,5 +22,9 @@ function getSceneTreeErrorMessage(code: SceneTreeErrorCode): string {
       return 'An unknown error occurred.';
     case SceneTreeErrorCode.SCENE_TREE_DISABLED:
       return 'The tree for this scene is not enabled. Enable the tree for this scene to interact with the tree.';
+    case SceneTreeErrorCode.UNINITIALIZED_VIEWER:
+      return 'Could not find reference to the viewer';
+    case SceneTreeErrorCode.DISCONNECTED:
+      return 'Disconnected from server.';
```
danschultz, line 28: This doesn't appear to be used.

PR #375, opened 2022-05-05
[PLAT-1035] Load layout on render

MadisonEhlers-Vertex:
## Summary

These changes introduce a resolution to a long standing bug. Here were my steps to reproduce the issue, although it was still intermittent:

1. In a developer console in Chrome, set the CPU slowdown to 4x or 6x. 
2. Open the Scene tree (Ideally if you're on a fast network)
3. The scene tree intermittently does not load. 

The nature of this bug is that the scene tree controller is dispatching updates from the network responses, prior to the scene tree component rendering for the first time. The first time render is where we ensure that the scene-tree-layout is defined:
```ts
  public componentWillRender(): void {
    this.ensureLayoutDefined(); // <-- here
    ...
  }
```

The `updateLayoutElements` only gets called on a state change, and within that function is the following:
```ts
  if (layout != null) {
      layout.rows = this.rows;
      layout.tree = this.el as HTMLVertexSceneTreeElement;
      layout.totalRows = this.totalRows;
      layout.controller = this.controller;
      layout.rowData = this.rowData;
    }
```

Currently we are eating any updates from the controller if the layout is not yet defined. My proposed solution here is to call this function `updateLayoutElement()` _after_ we have ensured that the template is defined in the above function `componentWillRender()`

Additionally, I have added a debug statement within the updateLayoutElements under the following condition: `else if (componentNotLoaded && this.rows > 0)`

## Test Plan

- See repro steps defined above. Expect the scene tree to _always_ load if the user is getting frames.

### DIFF
--- a/packages/viewer/src/components/scene-tree/scene-tree.spec.tsx
+++ b/packages/viewer/src/components/scene-tree/scene-tree.spec.tsx
@@ -325,6 +325,48 @@ describe('<vertex-scene-tree>', () => {
       )[0] as HTMLVertexSceneTreeTableCellElement;
       expect(row.node?.name).toEqual(res.toObject().itemsList[0].name);
     });
+
+    it('renders the scene tree data, even if the network requests completed before the scene tree render', async () => {
+      const client = mockSceneTreeClient();
+      const res = mockGetTree({ client });
+
+      const { stream, ws } = makeViewerStream();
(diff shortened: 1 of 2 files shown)


amvertex (approved):
Great job figuring this out!

danschultz (approved):

PR #376, opened 2022-05-05
[PLAT-645] Select scene tree search results

amvertex:
## Summary
We want the ability to select all of the matches for a given metadata search. 

## Test Plan
Tested with connect-app

## Release Notes
None

## Possible Regressions
None

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/scene-tree/scene-tree.tsx
+++ b/packages/viewer/src/components/scene-tree/scene-tree.tsx
@@ -554,12 +554,9 @@ export class SceneTree {
   @Method()
   public async selectFilteredItems(term: string): Promise<Row[] | void> {
     if (this.viewer != null) {
-      await selectFilterResults(
-        this.viewer,
-        term,
-        this.metadataKeys,
-        { append: false }
-      );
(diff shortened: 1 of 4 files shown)


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -543,6 +544,25 @@ export class SceneTree {
...
+   * @param term The filter term.
+   * @returns A promise that resolves with the selected rows.
+   */
+  @Method()
+  public async selectFilteredItems(term: string): Promise<Row[] | void> {
+    if (this.viewer != null) {
+      await selectFilterResults(this.viewer, term, this.metadataKeys, {
+        append: false,
+      });
+      return this.rows.filter((row) => row?.node.filterHit);
```
danschultz, line 560: I don't think we should return the filtered rows from this operation. The client doesn't have the full list of rows, so this result is very likely to be incomplete. If for some reason the caller needs this list, we'll have to find another way to return it to them.

MadisonEhlers-Vertex (approved):

PR #379, opened 2022-05-11
[PLAT-1194] [PLAT-1196] WebGL Transform Widget

jdm717:
## Summary

Adds initial support for a `vertex-viewer-transform-widget`. This component expects a `position` vector that represent the location to draw the widget, and dragging the X/Y/Z axis arrows will perform a world-delta translation of any selected items. This component internally uses [`regl`](https://www.npmjs.com/package/regl), and [`regl-shape`](https://www.npmjs.com/package/regl-shape) to draw the widget.

Below is an example setup to start using the widget - this will update the widget with the hit result's position, then select the item at that point. Afterward the widget will transform the part selected (or multiple selected).
```typescript
async function main() {
  await window.customElements.whenDefined('vertex-viewer');

  const viewer = document.getElementById('viewer');
  const widget = document.getElementById('widget');

  viewer.addEventListener('tap', async (e) => {
    const raycaster = scene.raycaster();

    const results = await raycaster.hitItems(e.detail.position);

    if (results.hits.length > 0) {
      widget.position = results.hits[0].hitPoint;
      scene.items(op => 
        op.where(q => q.withItemIds([results.hits[0].itemId.hex])).select()
      ).execute();
    } else {
      widget.position = undefined;
      scene.items(op => op.where(q => q.withSelected()).deselect()).execute();
    }
  });
}
```

```html
<vertex-viewer
  id="viewer"
  config-env="platdev"
  src='urn:vertexvis:stream-key:key'
>
  <vertex-viewer-transform-widget id="widget">
  </vertex-viewer-transform-widget>
</vertex-viewer>
```

Video:

https://user-images.githubusercontent.com/5252480/167930807-d81b88a8-6d64-420d-824b-b5de7fad8e54.mov

## Test Plan

- Test the newly added transform widget with a variety of models
  - Multi-select transforms
  - Clearing selection
  - Performing sequential (i.e. x-axis translation => y-axis translation) transforms 

## Release Notes

N/A

## Possible Regressions

Interactions (newly added payload)

## Dependencies

N/A


packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -0,0 +1,356 @@
...
+  private lastWorldPosition?: Vector3.Vector3;
+
+  private canvasBounds?: DOMRect;
+  private canvasResizeObserver!: ResizeObserver;
+  private canvasRef?: HTMLCanvasElement;
+
+  private hoveredChangeDisposable?: Disposable;
+
+  protected componentDidLoad(): void {
+    window.addEventListener('pointermove', this.handleWindowPointerMove);
```
MadisonEhlers-Vertex, line 94: Does this have to be on the window?

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -0,0 +1,356 @@
...
+  public disabledColor: Color.Color | string = '#cccccc';
+
+  /**
+   * @internal
+   * @private
+   *
+   * Visible for testing.
+   */
+  @Prop({ mutable: true })
+  public hovered?: Mesh;
```
MadisonEhlers-Vertex, line 79: Maybe a silly question, but why is this called a `Mesh`?

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -0,0 +1,356 @@
...
+      this.handleViewerDimensionsChanged
+    );
+
+    if (newViewer?.stream != null) {
+      this.controller?.dispose();
+      this.controller = new TransformController(newViewer.stream);
+    }
+
+    if (this.canvasRef != null && newViewer != null) {
+      this.canvasRef.width = newViewer.clientWidth;
```
MadisonEhlers-Vertex, line 136: Can this code be moved to a function and reused in the `handleViewerDemensionsChanged`? I see there that we use the viewer.viewport, but in this case we use the clientWidth. Can we use the viewport in both instances?

packages/geometry/src/rectangle.ts
```diff
@@ -206,6 +206,26 @@ export function pad(rect: Rectangle, padding: number): Rectangle {
...
+ * @param points The points to check.
+ */
+export function containsPoints(
+  rect: Rectangle,
+  ...points: Point.Point[]
+): boolean {
+  return points.every((point) => {
+    return (
+      rect.x < point.x &&
+      rect.x + rect.width > point.x &&
```
mikemccarty-vertex, line 222: if `rect.x == 0` and `rect.width == 1`, neither `point.x==0` nor `point.x==1` satisfy this condition.  I don't think that's what you want or expect.

What we've done before is define it has a half-interval, which leads to logic like `rect.x <= x && rect.x+rect.width < x`.  So `x==0` would pass but `x==1` would not.

packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
```diff
@@ -269,7 +269,7 @@ describe(Scene, () => {
               {
                 changeMaterial: {
                   material: {
-                    d: 100,
+                    d: 255,
```
mikemccarty-vertex, line 272: Did you discuss this with Dan at all?  This is somewhat complicated because we've already released this as a public API.  

To avoid potential backward compatibility issues, we've discussed changing the range to `[0,1]` just for opacity and then during the conversion on the server-side, take anything larger than `1.0` to mean fully opaque.

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
jdm717: It likely could be moved to the canvas itself, just with some conditional logic to remove the listener when we're actively transforming. I can take a look at updating that

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
jdm717: Honestly a bit of leftover terminology from when it was 3D, and had multiple triangles. Each line drawn however does contain multiple triangles - since WebGL doesn't necessarily support lines with a width higher than 1. Now that it's 2D though, maybe "mesh" isn't the best description...

packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
jdm717: I did not - just knew we had that ticket in the backlog to update these defaults. For the time being I can certainly revert this change

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.css
```diff
@@ -0,0 +1,18 @@
...
+
+.widget {
+  position: absolute;
+  top: 0;
+  left: 0;
+}
+
+.widget.hovered {
+  pointer-events: auto;
+}
```
MadisonEhlers-Vertex, line 18: nit: new line

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -0,0 +1,367 @@
...
+
+  private handleWindowPointerMove = (event: PointerEvent): void => {
+    const canvasPoint = convertPointToCanvas(
+      Point.create(event.clientX, event.clientY),
+      this.getCanvasBounds()
+    );
+    const widget = this.getTransformWidget();
+
+    if (canvasPoint != null && widget?.boundsContainsPoint(canvasPoint)) {
+      widget?.updateCursor(canvasPoint);
```
MadisonEhlers-Vertex, line 217: Is there a circumstance here that we would expect the widget to be undefined? If we expect the widget to be present, I would advocate for throwing an exception any place we are trying to use the widget and it's undefined. This will make troubleshooting & debugging a lot easier. Even if it's in an `if else` statement and we include a debug log in the `else` or something. Might be useful, given the sorts of stencil related bugs we've had recently.

MadisonEhlers-Vertex (commented):
Looks really good. I have a couple points of minor feedback. Also one suggestion to consider: What are your thoughts on adding a few debug statements in some of the components like the `vertex-viewer-transform-widget` and maybe in the `widget.ts`? Might help troubleshooting in the future to add some basic logging there or in any place that you think might have higher complexity. Just a thought. Looks really good!

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -0,0 +1,367 @@
...
+import { TransformWidget } from './widget';
+
+@Component({
+  tag: 'vertex-viewer-transform-widget',
+  styleUrl: 'viewer-transform-widget.css',
+  shadow: true,
+})
+export class ViewerTransformWidget {
+  /**
+   * The viewer to connect to measurements. If nested within a <vertex-viewer>,
```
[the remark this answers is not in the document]
jdm717: Copy/paste + forgot to update the comment 🙂  - good catch, I'll update

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -0,0 +1,367 @@
...
+   * The controller that is responsible for performing transforms.
+   */
+  @Prop({ mutable: true })
+  public controller?: TransformController;
+
+  /**
+   * The color of the translation arrow on the x-axis. Defaults to `#ea3324`.
+   */
+  @Prop()
+  public xArrowColor: Color.Color | string = '#ea3324';
```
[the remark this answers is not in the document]
jdm717: Yeah, I had considered using CSS variables on the widget and parsing those within the component. I'll take a look at that since we have some prior work that has done similar, seems like it'd be reasonable here as well

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
jdm717: That's fair, I'll take a look at adding some errors + logging to indicate what's wrong 👍

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
jdm717: Actually realized as I started looking into this why it was set up on the window to begin with - we need a `pointer-events: none` on the canvas so it doesn't prevent viewer interactions, which means we can't actually listen for the canvas-level events, and need to add to the window

mikemccarty-vertex:
> Additionally, as a small change with this PR, the default selection material has been updated to have an opacity of 255.

You didn't end up making this change, correct?

packages/geometry/src/rectangle.ts
jdm717: Ended up modifying this to allow for both the `point.x === 0` and `point.x === 1` case, particularly for the case I'm using this for - which is determining whether a point on screen is within the 2D bounds of the transform widget so we can start performing hit-tests for the hovered state. Open to renaming this, or adding some additional parameter to specify the behavior though

jdm717:
> > Additionally, as a small change with this PR, the default selection material has been updated to have an opacity of 255.
> 
> You didn't end up making this change, correct?

Correct - that's been reverted with [5a64ec6](https://github.com/Vertexvis/vertex-web-sdk/pull/379/commits/5a64ec656ff6ae1860f0db0f8ce4c2b86b4b1e66). Forgot to update the description to remove that, but it's updated now

packages/geometry/src/rectangle.ts
mikemccarty-vertex: The logic you have here only accepts points "in" the rect but not "on" the rect.  That's fine if that's what you need but it is a subtle distinction that should probably be documented somewhere.

mikemccarty-vertex (approved):
LGTM

MadisonEhlers-Vertex (approved):

PR #383, opened 2022-05-16
[PLAT-996] Toggle exact match setting for metadata search

amvertex:
## Summary
We want the user to be able to toggle between exact match and fuzzy search when searching the scene tree.

## Test Plan
Tested in connect-app

## Release Notes
None

## Possible Regressions
Scene tree search

## Dependencies
None


packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -518,6 +523,7 @@ export class SceneTreeController {
           const req = new FilterRequest();
           req.setFilter(term);
           req.setFullTree((options.includeCollapsed ?? true) === true);
+          req.setExactMatch((options.exactMatch ?? false) === true);
```
MadisonEhlers-Vertex, line 526: Can't this just be represented as `req.setExactMatch(!!options.exactMatch)`?

Also, the above line can be simplified to the opposite:
`req.setFullTree(!options.includeCollapsed)`

amvertex: Good idea. Updated.

MadisonEhlers-Vertex (approved):

PR #384, opened 2022-05-17
[PLAT-1541] Orthographic transform widget support

jdm717:
## Summary

Adds support for the transform widget in orthographic mode.

## Test Plan

- Test the widget in orthographic mode

## Release Notes

N/A

## Possible Regressions

Orthographic ray generation

## Dependencies

N/A


mikemccarty-vertex (approved):

packages/viewer/src/lib/transforms/hits.ts
```diff
@@ -49,5 +53,7 @@ export function testTriangleMesh(
 
   const r = Vector3.dot(edge2, q) / det;
 
-  return !isNaN(r) && r > epsilon;
+  // Ignore the case where the computed hit position is negative
+  // if in orthographic, as the near plane can be a negative value.
+  return !isNaN(r) && (r > 0 || frame.scene.camera.isOrthographic());
```
mikemccarty-vertex, line 58: A better solution would be if `transformPointToOrthographicRay()` constructed a ray where the origin was at the near plane.

mikemccarty-vertex (approved):
I have some minor reservations about the logic commented (I have a feeling it could cause subtle problems later) but not enough to hold up progress.

packages/viewer/src/lib/transforms/hits.ts
jdm717: I actually do think that should be the case, so thinking about this again, I'm not sure my comment here is accurate. I'll update my comment to reflect that it's a bit of a hack to correct the orthographic case where it's moving toward the camera, and create a ticket to revisit this logic

jdm717: https://vertexvis.atlassian.net/browse/PLAT-1549

mikemccarty-vertex (approved):

PR #385, opened 2022-05-18
[PLAT-1528] Emit Position Changing for the transform widget

MadisonEhlers-Vertex:
## Summary

Introduces a new event to be emitted each time the position of the widget changes.

```ts
      const widget = document.getElementById('widget');
      widget.addEventListener('positionChanged', (e) => {
        console.log('position: ', e.detail);
      })
```

### DIFF
--- a/packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
+++ b/packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
@@ -1,4 +1,13 @@
-import { Component, Element, h, Host, Prop, Watch } from '@stencil/core';
+import {
+  Component,
+  Element,
+  Event,
+  EventEmitter,
+  h,
+  Host,
+  Prop,
+  Watch,
+} from '@stencil/core';
 import { Point, Vector3 } from '@vertexvis/geometry';
 import { Color, Disposable } from '@vertexvis/utils';
 import classNames from 'classnames';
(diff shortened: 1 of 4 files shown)


packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
```diff
@@ -19,6 +28,12 @@ import { TransformWidget } from './widget';
   shadow: true,
 })
 export class ViewerTransformWidget {
+  /**
+   * An event that is emitted when the position of the widget changes.
+   */
+  @Event({ bubbles: true })
+  public positionChanged!: EventEmitter<Vector3.Vector3>;
```
jdm717, line 35: Should this be:
```typescript
public positionChanged!: EventEmitter<Vector3.Vector3 | undefined>;
```
? Or potentially just in an `else` on the condition above where we emit it if it'll never change to `undefined` internally. I'm actually a bit surprised that the `emit` doesn't complain that it could be undefined

jdm717 (approved):
Had one small nit, but LGTM!

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
MadisonEhlers-Vertex: Yeah, I had a similar thought. By default, the emitter accepts optional values:
```
export interface EventEmitter<T = any> {
    emit: (data?: T) => CustomEvent<T>;
}
```

I think preventing an emit on an undefined value is a fine approach here.

PR #388, opened 2022-05-23
[PLAT-1560] Correct orthographic depth

jdm717:
## Summary

Corrects the math for orthographic depth calculation. Previously, this math was attempting to account for a ray generated at `z = 0.5`, which is no longer generated by the `Viewport.transformPointToOrthographicRay` method. Additionally, this PR updates a couple tests that should have caught this bug (in packages/viewer/src/lib/types/__tests__/depthBuffer.spec.ts), as they were not using the correct ray generation/world point generation methods.

## Test Plan

- Test rotate around tap point in orthographic mode
- Unit tests

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/types/depthBuffer.ts
+++ b/packages/viewer/src/lib/types/depthBuffer.ts
@@ -102,7 +102,7 @@ export class DepthBuffer implements FrameImageLike {
       fallbackNormalizedDepth
     );
 
-    return depth * (far - near) + near / 2;
+    return depth * (far - near);
   }
 
   /**
--- a/packages/viewer/src/components/viewer-measurement-distance/__tests__/hitTest.spec.ts
+++ b/packages/viewer/src/components/viewer-measurement-distance/__tests__/hitTest.spec.ts
@@ -14,7 +14,7 @@ describe(PointToPointHitTester, () => {
 
     const viewport = new Viewport(100, 100);
     const depthBuffer = makeDepthBuffer(100, 100, undefined, camera);
-    const depth = camera.far - camera.near + camera.near / 2;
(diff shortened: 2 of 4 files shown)


packages/viewer/src/lib/types/depthBuffer.ts
```diff
@@ -102,7 +102,7 @@ export class DepthBuffer implements FrameImageLike {
       fallbackNormalizedDepth
     );
 
-    return depth * (far - near) + near / 2;
+    return depth * (far - near);
```
mikemccarty-vertex, line 105: NIT: if I'm understanding correctly, this probably ought to be `... + near` but `near==0` so it really doesn't matter mathematically but it might make more sense to a future reader.

mikemccarty-vertex (approved):

packages/viewer/src/lib/types/depthBuffer.ts
jdm717: I've typically seen the orthographic cameras have a near/far of `[-1000, 1000]`, so this results in a depth between `[0, 2000]` in those cases (where if we added `... + near` we'd wind up with a depth between `[0, 1000]`. I did test with that addition, but saw incorrect results. Does the range from `[0, 2000]` in that case actually make sense in orthographic? Or is that potentially another symptom of some incorrect math somewhere else?

mikemccarty-vertex: Well, it does compute exactly the value the docstring claims it computes so maybe I should be quiet.  :shipit:

PR #390, opened 2022-05-24
PLAT-1322: Add noDefaultLights property

sourada-e5:
## Summary
The value of the noDefaultLights property is passed on to the stream attributes when it is changed.
[PLAT-1322](https://vertexvis.atlassian.net/browse/PLAT-1322)


packages/viewer/src/lib/mappers/streamAttributes.ts
```diff
@@ -79,20 +79,30 @@ const toPbFeatureHighlight: M.Func<
   })
 );
 
+const toPbNoDefaultLights: M.Func<boolean | undefined, boolean> =
+  M.defineMapper(
+    (noDefaultLights) => {
+      return noDefaultLights ? noDefaultLights : false;
+    },
```
jdm717, line 86: Minor nit: a small shortcut here (assuming `noDefaultLights` initially is `boolean | undefined`) is to use `!!noDefaultLights`, which if the value is `undefined` will coerce it to false for the boolean operation. Then the mapper could just be `(noDefaultLights) => !!noDefaultLights`

jdm717 (approved):
LGTM! Just had one very minor nit

jdm717:
> I'm surprised that adding a change didn't dismiss your approval, @jdm717 .

Yeah, I don't think we have that setting enabled here, which sometimes is nice 🙂 

> Is there anything I need to do re: versioning?

Nope! When this PR gets merged it'll automatically publish a canary release of the SDK

PR #391, opened 2022-05-26
[PLAT-1570] -- opaque selection material

gbiv:
## Summary

Set opacity on default materials to 255
Set glossiness, ie specular exponent, to 50 (is this too glossy?)

## Test Plan

Verified in a test app
Will need to be picked up by Connect+ 

## Dependencies

[Ticket](https://vertexvis.atlassian.net/browse/PLAT-1570)

## NOTES
- [x] ~~Is 50 too high?~~ Set to 4
- [x] Does the version get bumped automatically or does it need to be specified in this PR
- [ ] Need to pick this up in Connect+

### DIFF
--- a/packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
+++ b/packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
@@ -269,7 +269,7 @@ describe(Scene, () => {
               {
                 changeMaterial: {
                   material: {
-                    d: 100,
+                    d: 255,
                     ka: {
                       a: 0,
                       b: 0,
@@ -294,7 +294,7 @@ describe(Scene, () => {
                       g: 0,
                       r: 0,
                     },
-                    ns: 10,
+                    ns: 50,
                   },
                 },
               },
--- a/packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
+++ b/packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
@@ -294,7 +294,7 @@ describe(Scene, () => {
                       g: 0,
                       r: 0,
                     },
-                    ns: 50,
+                    ns: ColorMaterial.defaultColor.glossiness,
                   },
                 },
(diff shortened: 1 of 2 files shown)


packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
```diff
@@ -294,7 +294,7 @@ describe(Scene, () => {
                       g: 0,
                       r: 0,
                     },
-                    ns: 10,
+                    ns: 50,
```
mikemccarty-vertex, line 297: RWS defaults to [4 when all else fails](https://github.com/Vertexvis/rendering-worker-service/blob/6eaa9442088f96d08a2e5c405b8415d82743b42f/src/Embree.cpp#L42).  There's no science behind that number, just something being better than nothing.

Shininess is expressed as an exponent so the effect is asymptotic as the values get larger.

packages/viewer/src/lib/scenes/colorMaterial.ts
```diff
@@ -13,8 +13,8 @@ export interface ColorMaterial {
 }
 
 const defaultColor: ColorMaterial = {
-  opacity: 100,
-  glossiness: 10,
+  opacity: 255,
+  glossiness: 50,
```
mikemccarty-vertex, line 17: Can we pull the default glossiness out as a constant to ensure it is the same everywhere it is used?

gbiv: exported the defaultColor referenced the value from there.

packages/viewer/src/lib/scenes/__tests__/scene.spec.ts
gbiv: Makes sense to me to sync it up with rw.

mikemccarty-vertex (approved):
LGTM

jdm717 (approved):

PR #404, opened 2022-06-15
PLAT-1680: Add experimental_rendering_options property to viewer

sourada-e5:
## Summary
and pass it through to FSS.

## Test Plan
Will be exercised in a later PR.


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -231,6 +231,11 @@ export class Viewer {
    */
   @Prop() public noDefaultLights = false;
 
+  /**
+   * Specifies experimental rendering options. For Vertex use only.
+   */
```
danschultz, line 236: Can we add this doc metadata so this field is not documented.

```ts
/**
 * @private
 * @internal
 */
```

packages/stream-api/package.json
```diff
@@ -35,7 +35,7 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
-    "@vertexvis/frame-streaming-protos": "^0.6.22"
+    "@vertexvis/frame-streaming-protos": "^0.7.0"
```
danschultz, line 38: I believe this should be updated in `@vertexvis/viewer` package as well.

danschultz (approved):

packages/viewer/src/components/viewer/viewer.tsx
sourada-e5: If I specify `@internal`, the reference to the property in these lines in connect-app don't work. Is there another way to do this to avoid that problem? https://github.com/Vertexvis/connect-app/pull/483/files#diff-d70ac4a03197f1c59a02349a3466d3dfc5d17f805caf796d5a7a03ab74cd7894R143

PR #408, opened 2022-06-22
[PLAT-1664] Adding UX Loading improvements to the scene tree

MadisonEhlers-Vertex:
## Summary

These changes support adding a loader to the search bar, as well as a loader to the scene tree when items are taking beyond `2s`

loading

PR #409, opened 2022-06-23
[PLAT-1501] Client Performance Improvements

MadisonEhlers-Vertex:
## Summary

The primary changes for this PR center around client side performance improvements for the scene tree. 

This PR does include some feedback pertaining a prior pr:
 https://github.com/Vertexvis/vertex-web-sdk/pull/408

See the before and after (more detailed description below the gifs):

Before:
scene tree before2

After:
Scene tree  improvements

The two primary improvements were around reducing the number of cell draws on many scroll events (set with a debounce). The secondary item that improved performance was to reduce the noise in the console logs in the scene tree. Those logs were producing quite a bit of computation load. [This was an insightful read regarding this.](https://stackoverflow.com/a/26754011)

The cell draw events would compute this many operations on each scroll event:
```
    r = number of items in range
    c = number of cells
    r * c
```

Reducing the number of cell drawn events when rapid scrolling as well as removing the noisy console logs vastly improve the performance here. 

## Test Plan

Rapidly scroll the scene tree with many columns, gently scroll, filter, etc.


packages/viewer/src/components/scene-tree-table-layout/scene-tree-table-layout.tsx
```diff
@@ -426,13 +391,16 @@ export class SceneTreeTableLayout {
...
           ? (depth: number) => `calc(${depth} * 0.5rem)`
           : () => `0`;
-      pool.iterateElements((el, binding, rowIndex) => {
-        const row = this.stateMap.viewportRows[rowIndex];
 
-        if (isLoadedRow(row)) {
-          this.updateCell(row, el, binding, rowIndex, cellPaddingLeft);
-        }
-      });
+      if (!this.isScrolling) {
```
danschultz, line 395: I think this is fine for a first pass to address the immediate issue, but is not an ideal fix. If we can't keep pace with rendering while scrolling, there's something fundamentally wrong that I think we should investigate as part of this ticket.

danschultz (approved):

PR #411, opened 2022-06-27
[PLAT-1724] Dimensions should be updated to whatever was set in the client after connection is successful

MadisonEhlers-Vertex:
## Summary

The repro steps are defined in the ticket [PLAT-1724](https://vertexvis.atlassian.net/browse/PLAT-1724).

This can be reproduced any time that the client side stream model gets an updated dimension while not in the `connected` state. This is easily reproducible by resizing the viewer window, while the frame stream was reconnecting. 

Prior to these changes, the method `updateDimensions` was only invoked when the stream was in a connecting state:
```ts
      this.dimensions = fields.dimensions;

      this.ifState('connected', () => {
        this.updateDimensions({ dimensions: this.dimensions });
      });
```

This led to the dimensions in the `ViewerStream` to be mismatched with what the server side frame stream was reporting. 

The change here, detects this situation, and updates the dimensions correctly.

### DIFF
--- a/packages/viewer/src/lib/stream/stream.ts
+++ b/packages/viewer/src/lib/stream/stream.ts
@@ -163,9 +163,9 @@ export class ViewerStream extends StreamApi {
 
     if (fields.dimensions != null && fields.dimensions !== this.dimensions) {
       this.dimensions = fields.dimensions;
-      this.ifState('connected', () =>
-        this.updateDimensions({ dimensions: this.dimensions })
-      );
+      this.ifState('connected', () => {
+        this.updateDimensions({ dimensions: this.dimensions });
+      });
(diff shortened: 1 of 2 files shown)


packages/viewer/src/lib/stream/stream.ts
```diff
@@ -321,6 +321,13 @@ export class ViewerStream extends StreamApi {
 
         if (this.state.type === 'connected') {
           this.updateState({ ...this.state, frame });
+
+          if (
+            this.dimensions &&
```
danschultz, line 326: The existence check for `this.dimensions` is unnecessary.

danschultz (approved):

PR #412, opened 2022-06-28
[PLAT-1743][PLAT-1744] Invalidate current page on Stream End Or Stream failure

MadisonEhlers-Vertex:
## Summary

These changes will invalidate the current page that will then allow for the scene tree to fetch the correct tree after a user takes an action, even on a `/Subscribe` failure.

## Test Plan

- Block the `/Subscribe` Network Request. 
- You should be able to do basic scene tree operations (visibility, expansion, etc)


packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -746,6 +746,19 @@ export class SceneTreeController {
         }
       });
 
+      stream.on('status', (s) => {
+        if (s.code !== 0) {
+          console.error(
+            `Failed to load scene tree with code=${s.code}, details=${s.details}`
```
danschultz, line 752: Let's reword to `Failed to subscribe to scene tree with code=${s.code}, details=${s.details}`

danschultz:
👀

danschultz (approved):

packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -177,7 +177,7 @@ export class SceneTree {
    * An object to configure the scene tree.
    */
   @Prop()
-  public config?: Config;
+  public config?: Config | string;
```
danschultz, line 180: This all should be `PartialConfig | string`.

danschultz (approved):

danschultz (approved):

PR #418, opened 2022-07-11
[PLAT-1791] Add support for custom scene-tree operation handlers

jdm717:
## Summary

Adds the `selectionHandler`, `expansionHandler`, and `visibilityHandler` properties to the `vertex-scene-tree-table-cell` which, if specified, will override the default selection, expansion, and visibility behaviors respectively. As this provides an alternate approach to what we created the `selectionToggled`, `expansionToggled`, and `visibilityToggled` events for, those events have been removed. Additionally, the `disableInteractions` and `recurseParentSelectionDisabled` flags have been removed in favor of providing custom handlers that modify the behavior.

## Test Plan

- Verify providing a `selectionHandler` with the same selection behavior, and an additional `!event.altKey` check prevents selection when the alt key is pressed
- Verify items are selected while holding the `Alt` key and not setting any `selectionHandler`

## Release Notes

N/A

## Possible Regressions

Selection while holding the `Alt` key and without this new flag set to `true`

## Dependencies

N/A


packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -94,6 +94,14 @@ export class SceneTreeTableCell {
   @Prop()
   public recurseParentSelectionDisabled = false;
 
+  /**
+   * A flag that disables selection if the Alt (or Option on Mac) key is held
+   * down. When enabled, selection will occur as though no modifier keys are
+   * pressed.
+   */
+  @Prop()
+  public disableAltKeySelection = false;
```
danschultz, line 103: This seems like a very Connect specific flag. Can we generalize selection behavior somehow, so it applies to applications more broadly?

jdm717:
@danschultz - made some changes to support passing a set of enabled modifier keys as opposed to having a single flag for the alt-key behavior. Let me know if that makes more sense, or if you had something more specific in mind around selection behavior. I had also considered supporting providing a predicate to run prior to selection, but that would be a somewhat new API, since I'm not sure that we've supported providing predicates as props before

danschultz (approved):

PR #419, opened 2022-07-12
[PLAT-1722] STS Subscription Handshake

MadisonEhlers-Vertex:
## Summary

Shows an error message if the tree does not get subscription events.
image

## Dependencies

https://github.com/Vertexvis/scene-tree-service/pull/115


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -809,7 +806,6 @@ export class SceneTree {
 
   private handleControllerStateChange(state: SceneTreeState): void {
     this.showLoader = !!state.shouldShowLoading;
-    this.isSearching = state.isSearching;
```
amvertex, line 812: Was this intentional to delete?

MadisonEhlers-Vertex: Yeah, this wasn't in use.

amvertex (approved):

PR #420, opened 2022-07-14
[PLAT-1492] Pin Tool Supports Primary & Accent Colors

MadisonEhlers-Vertex:
## Summary

These changes support coloring the pins in the pin tool. 

Explored approaches
1. My original approach was to apply templating to the pins, but since we draw the pins in the pin tool, we would have to create a 1:1 template per pin.
2. [PROPOSED] Setting a primary/secondary color on the pin tool, and applying the values as attributes in the pin model. 
3. There is also an option of applying some sort of `onDrawPin((pin) => // apply colors)` method, which I explored for a little bit, but found option 2 to be a cleaner approach so far. 

## Test Plan

1. You should be able to pass already colored pins to the pin tool and they will be drawn with the selected primary/accent colors
2. You should be able to change color of the pins by setting the accent/primary colors in the pin tool (applies to new pins)


packages/viewer/src/lib/pins/model.ts
```diff
@@ -44,10 +50,34 @@ export function isTextPin(pin?: Pin): pin is TextPin {
...
+  return typeof color === 'string' ? color : Color.toHexString(color);
+}
+export function getPinColors(pin?: Pin): {
+  primaryColor?: string;
+  accentColor?: string;
+} {
+  return {
+    primaryColor:
+      pin?.attributes?.style?.primaryColor == null
+        ? undefined
```
amvertex, line 72: Do we want to set a default value, or what will the pin look like if it's undefined?

packages/viewer/src/components/viewer-pin-group/viewer-pin-group.css
```diff
@@ -1,16 +1,18 @@
 :host {
 
+  --viewer-annotations-pin-primary-color: var(--blue-700);
+  --viewer-annotations-pin-accent-color: var(--blue-200);
   /**
    * @prop  --viewer-annotations-pin-label-line-color: A CSS color that
    * specifies the color of the label line
    */
-   --viewer-annotations-pin-dot-color: var(--blue-600);
+   --viewer-annotations-pin-dot-color: var(--viewer-annotations-pin-primary-color);
```
jdm717, line 9: Not specific to this PR, but it is a bit confusing that the comment here refers to `--viewer-annotations-pin-label-line-color`, but the actual variable is `--viewer-annotations-pin-dot-color`. Could we update this comment to indicate what the `--viewer-annotations-pin-dot-color` variable does?

packages/viewer/src/components/viewer-pin-label/viewer-pin-label.tsx
```diff
@@ -187,6 +193,8 @@ export class VertexPinLabel {
             minWidth: this.computeMinWidth(),
             maxWidth: this.computeMaxWidth(),
             maxHeight: this.computeMaxHeight(),
+            'border-color': primaryColor,
```
jdm717, line 196: Minor nit: could this just be `borderColor: primaryColor` to follow the same camelCase pattern of the other styles?

packages/viewer/src/components/viewer-pin-tool/viewer-pin-tool.tsx
```diff
@@ -60,6 +60,18 @@ export class ViewerPinTool {
   @Prop({ mutable: true })
   public mode: ViewerPinToolMode = 'view';
 
+  /**
+   * The primary color for new pins.
+   */
```
jdm717, line 65: Thoughts on expanding what a "primary" color for these pins is? I'm not sure I'd understand what these actually change from documentation currently. (or if this should be internal, could we add the `@internal`?)

packages/viewer/src/components/viewer-pin-tool/viewer-pin-tool.tsx
```diff
@@ -60,6 +60,18 @@ export class ViewerPinTool {
...
+  @Prop({ mutable: true })
+  public primaryColor: Color.Color | string | undefined;
+
+  /**
+   * The accent color for new pins.
+   */
```
jdm717, line 71: Same as above, might be worth expanding on what an "accent" color is as well

jdm717 (commented):
Overall LGTM! Just had a few thoughts, mainly around documentation

packages/viewer/src/components/viewer-pin-tool/viewer-pin-tool.tsx
MadisonEhlers-Vertex: This is not internal, this is currently used for setting the primary/accent color on a per pin basis. I'll expand on this in a comment.

jdm717 (approved):
LGTM!

PR #430, opened 2022-07-27
[PLAT-1624] Skip drawing older frames to the canvas

jdm717:
## Summary

This change adds a `drawPredicate` to the `canvasRenderer` in the viewer to determine whether frames should be drawn to the canvas after completing the `loadImageBytes` call. This corrects a race condition between smaller and larger frames that come in at close to identical times, such as when all parts are hidden in the viewer.

## Test Plan

- Load a (large-ish) model, and zoom in such that the image takes up the entire viewer (also, the larger the viewer window the more reproducible this bug is), then toggle visibility from the scene tree repeatedly
  - The image should always update with empty frames when the root node is hidden

## Release Notes

N/A

## Possible Regressions

Frame drawing behavior

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/rendering/canvas.ts
+++ b/packages/viewer/src/lib/rendering/canvas.ts
@@ -13,8 +13,7 @@ export interface DrawFrame {
   canvasDimensions: Dimensions.Dimensions;
   frame: Frame;
   viewport: Viewport;
-  loadPredicate?: () => boolean;
-  drawPredicate?: () => boolean;
+  predicate?: () => boolean;
   beforeDraw?: VoidFunction;
 }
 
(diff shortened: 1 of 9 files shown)


amvertex (approved):

MadisonEhlers-Vertex (approved):

packages/viewer/src/lib/rendering/canvas.ts
```diff
@@ -118,17 +119,23 @@ export function createCanvasRenderer(): CanvasRenderer {
     const frameNumber = data.frame.sequenceNumber;
     const frameIsNewer =
       lastFrameNumber == null || frameNumber > lastFrameNumber;
-    const image = await loadImageBytes(data.frame.image.imageBytes);
-    const predicatePassing = data.predicate?.() ?? true;
+    const loadPredicatePassing = data.loadPredicate?.() ?? true;
+
+    if (frameIsNewer && loadPredicatePassing) {
+      const image = await loadImageBytes(data.frame.image.imageBytes);
+      const drawPredicatePassing = data.drawPredicate?.() ?? true;
```
danschultz, line 126: Drawing an older frame is a bug with the renderer IMO. The default implementation shouldn't allow this, so I don't see a need of a predicate here. If we did introduce a predicate later, I'd suggest the predicate is not even called for a frame that is older.

I would suggest that we:
* Break up the decoding of an image from the drawing. Basically treat them as two different processes within the renderer.
* Don't block on image decoding, e.g. no `await`. It'll try to decode as fast as it can, and forward the decoded image to the drawing process. The decoding process will still not attempt to decode old frames. Because decoding is async, it's possible that images are forwarded to the drawing process out of order.
* The drawing process takes the decoded images and is aware of the last drawn frame. It will not attempt to draw a decoded image that is older.

amvertex (approved):

danschultz (approved):

PR #431, opened 2022-08-03
[PLAT-1808] Add utility to parse MSB, LSB from UUID strings

danschultz:
## Summary

Adds a utility to parse MSB and LSB from UUID strings.

https://vertexvis.atlassian.net/browse/PLAT-1808

### DIFF
--- a/packages/utils/src/uuid.ts
+++ b/packages/utils/src/uuid.ts
@@ -1,5 +1,28 @@
-import { v1 as uuid } from 'uuid';
+import Long from 'long';
+import { parse, v4 as uuid } from 'uuid';
 
 export type UUID = string;
 
-export const create = (): UUID => uuid();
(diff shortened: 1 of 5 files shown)


MadisonEhlers-Vertex (approved):

packages/utils/src/uuid.ts
```diff
@@ -1,5 +1,28 @@
-import { v1 as uuid } from 'uuid';
+import Long from 'long';
+import { parse, v4 as uuid } from 'uuid';
 
 export type UUID = string;
 
-export const create = (): UUID => uuid();
+export interface UUIDMsbLsb {
```
sourada-e5, line 6: Nit: I think I'd return these as numbers and let the caller decide to stringify them as necessary.

MadisonEhlers-Vertex (approved):

PR #433, opened 2022-08-04
[PLAT-1773] preventDefault when dragging transform widget

amvertex:
## Summary
When dragging the transform widget outside of the viewer, the sidebar and header text became selected.  Further, if a Windows user dragged the transform widget, interacted with the view cube, and then attempted to drag the transform widget again, the widget continued to move after lifting the mouse button.

## Test Plan
Interact with the transform widget.  Verify that text is not selected when the widget is dragged outside of the viewer and the widget continues to work after interacting with the view cube.

## Release Notes
None

## Possible Regressions
Dragging the transform widget

## Dependencies
None


gbiv (approved):
LGTM

PR #435, opened 2022-08-23
[PLAT-1876] Expose scene ID in viewer

danschultz:
## Summary

Exposes scene ID in the viewer.

https://vertexvis.atlassian.net/browse/PLAT-1876

### DIFF
--- a/packages/viewer/src/lib/scenes/scene.ts
+++ b/packages/viewer/src/lib/scenes/scene.ts
@@ -236,6 +236,7 @@ export class Scene {
     private decodeFrame: FrameDecoder,
     private imageScaleProvider: ImageScaleProvider,
     private dimensions: Dimensions.Dimensions,
+    public readonly sceneId: UUID.UUID,
     public readonly sceneViewId: UUID.UUID,
     private defaultSelectionMaterial: ColorMaterial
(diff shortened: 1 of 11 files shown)


packages/viewer/src/lib/scenes/scene.ts
```diff
@@ -236,6 +236,7 @@ export class Scene {
     private decodeFrame: FrameDecoder,
     private imageScaleProvider: ImageScaleProvider,
     private dimensions: Dimensions.Dimensions,
+    public readonly sceneId: UUID.UUID,
```
MadisonEhlers-Vertex, line 239: Question: Why not a private variable with a public getSceneId function?

MadisonEhlers-Vertex (approved):
+11. One suggestion, though. We have a `Stream connected [stream-id={streamId}, scene-view-id={sceneViewID}]
` log, I would advocate for adding the scene id on that log so it's easily accessible in the dev tools.

packages/viewer/src/lib/scenes/scene.ts
danschultz: That's equivalent to `public readonly`, but requires more code.

PR #436, opened 2022-08-23
[PLAT-1959] Fix SonarQube warnings

danschultz:
## Summary

https://vertexvis.atlassian.net/browse/PLAT-1959


packages/viewer/src/components/viewer-view-cube/viewer-view-cube.css
```diff
@@ -54,7 +54,6 @@
    */
   --viewer-view-cube-triad-z-axis-color: var(--z-axis-color);
 
-  display: block;
```
amvertex, line 57: Was this meant to be included in this PR?

danschultz: Yep, there are two `display` properties that were set. `display: flex` is the one that should be used.

PR #442, opened 2022-10-12
[PLAT-2134] Add support for WebSocketTransport-based Scene Tree subscriptions

jdm717:
## Summary

Adds a new flag to the scene-tree controller's `ConnectOptions` that enables usage of the `WebsocketTransport` for subscription requests (defaulted to `true`). This will then use `grpc.setDefaultTransport` to modify the default transport prior to the subscribe request, and immediately set it back to the default `CrossBrowserHttpTransport`. This is supported on a per-request basis as well, but due to the way our protobuf packages are generated, we don't have the ability to modify the transport for those directly.

## Test Plan

- Verify that scene tree requests work as expected (expand/collapse/etc)
- Verify that scene tree subscriptions are over websocket by default
- Verify that scene tree subscriptions are over http when `subscriptionUseWebSocketTransport` is false

## Release Notes

N/A

## Possible Regressions

Scene tree interactions/subscription

## Dependencies

https://github.com/Vertexvis/scene-tree-service/pull/120

### DIFF
--- a/packages/viewer/src/components/scene-tree/lib/controller.ts
+++ b/packages/viewer/src/components/scene-tree/lib/controller.ts
@@ -38,6 +38,7 @@ export interface ConnectOptions {
   idleReconnectInSeconds?: number;
   lostConnectionReconnectInSeconds?: number;
   subscriptionHandshakeGracePeriodInMs?: number;
+  subscriptionUseWebSocketTransport?: boolean;
 }
 
 export type JwtProvider = () => string | undefined;
@@ -163,6 +164,7 @@ export class SceneTreeController {
     private connectOptions: ConnectOptions = {
       spinnerDelay: 2000,
       subscriptionHandshakeGracePeriodInMs: 5000,
+      subscriptionUseWebSocketTransport: true,
     }
   ) {}
 
@@ -989,7 +991,20 @@ export class SceneTreeController {
     req: (metadata: grpc.Metadata) => ResponseStream<R>
   ): ResponseStream<R> {
     const metadata = this.createJwtMetadata(jwt);
-    return req(metadata);
+    return this.connectOptions.subscriptionUseWebSocketTransport
+      ? this.serverStreamWithWebSocket(() => req(metadata))
+      : req(metadata);
+  }
+
+  private serverStreamWithWebSocket<R>(
+    req: () => ResponseStream<R>
+  ): ResponseStream<R> {
+    grpc.setDefaultTransport(grpc.WebsocketTransport());
+    const stream = req();
+    grpc.setDefaultTransport(
+      grpc.CrossBrowserHttpTransport({ withCredentials: false })
+    );
+    return stream;
   }
 
   private createJwtMetadata(jwt: string): grpc.Metadata {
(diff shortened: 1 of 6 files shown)


packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -989,7 +991,20 @@ export class SceneTreeController {
...
+  }
+
+  private serverStreamWithWebSocket<R>(
+    req: () => ResponseStream<R>
+  ): ResponseStream<R> {
+    grpc.setDefaultTransport(grpc.WebsocketTransport());
+    const stream = req();
+    grpc.setDefaultTransport(
+      grpc.CrossBrowserHttpTransport({ withCredentials: false })
+    );
```
jdm717, line 1006: Open to other ideas here on how to approach usage of this `WebsocketTransport`. I've opted to go with the `setDefaultTransport` swap here to avoid having to instantiate another `SceneTreeAPIClient` with the transport and change the signature of this controller's constructor. This is also likely an issue that goes away once we can pull the public protos directly and generate a client that handles this transport changing

amvertex (approved):

packages/viewer/src/components/scene-tree/lib/controller.ts
danschultz: This implementation isn't ideal because it's configuring the transport globally. I can't confirm this, but conceptually it could mean changing the transport of another library that's using grpc-web.

I would look into configuring the transport using the `RpcOptions` that you pass when constructing a `SceneTreeAPIClient`. On it there's a `transport` property where you can define the transport based on the RPC method you're trying to call.

jdm717: @danschultz Yeah, that's correct - the other option here would be to create a `subscriptionClient` (or similar) that is instantiated as:
```typescript
new SceneTreeAPIClient(sceneTreeHost, {
  transport: grpc.WebSocketTransport(),
});
```
I can certainly make that change, but it would mean maintaining two separate clients + a breaking change to the controller's constructor signature. I can definitely make that change though

packages/viewer/src/lib/types/flags.ts
```diff
@@ -28,8 +28,14 @@ type Flag =
...
    */
-  | 'letterboxFrames';
+  | 'letterboxFrames'
 
+  /**
+   * Enables or disables the use of the `grpc.WebsocketTransport` for scene
+   * tree subscriptions. By default, this transport will be used, and
+   * subscription messages will be sent over WebSocket.
+   */
+  | 'useSubscriptionWebSocketTransport';
```
danschultz, line 38: Let's generalize to something like `grpcUseStreamingWebSocketTransport`. May need to be used for other streaming endpoints.

danschultz, line 38: Let's generalize to something like `grpcUseStreamingWebSocketTransport`. May need to be used for other streaming endpoints.

danschultz (approved):
LGTM after change.

danschultz (approved):

PR #443, opened 2022-10-19
[PLAT-2093] onPinAdded listener

amvertex:
## Summary
This PR adds a listener to our pin model specifically for when a pin is added.  This allows us to accurately track the number of pins added in our analytics, as we were previously using the _onEntitiesChanged_ listener, which is called in several cases other than adding a pin, such as when saving a snapshot or loading an existing snapshot, and resulted in a higher number of 'pin added' events in mixpanel than actually occurred. 

## Test Plan
After the connect-app PR, verify that the number of 'pin added' events in mixpanel is equal to the number of pins you added

## Release Notes
None

## Possible Regressions
Adding pins and the analytics around them

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/pins/__tests__/model.spec.ts
+++ b/packages/viewer/src/lib/pins/__tests__/model.spec.ts
@@ -1,0 +1,55 @@
+import { Point, Vector3 } from '@vertexvis/geometry';
+
+import { PinModel, TextPin } from '../model';
+
+describe('PinModel', () => {
+  describe(PinModel.prototype.setPin, () => {
+    it('adds pin to the supplied set', () => {
+      const model = new PinModel();
+      const onChange = jest.fn();
+      const onAdd = jest.fn();
+      model.onEntitiesChanged(onChange);
+      model.onPinAdded(onAdd);
+
+      const pin: TextPin = {
+        type: 'text',
+        id: 'my-pin-id',
+        worldPosition: Vector3.create(),
+        label: {
+          point: Point.create(),
+          text: 'My New Pin',
+        },
+      };
+
+      model.setPin(pin);
+      expect(model.getPins()).toEqual([pin]);
+      expect(onChange).toHaveBeenCalledWith([pin]);
+      expect(onAdd).toHaveBeenCalledWith([pin]);
+    });
+  });
+
+  describe(PinModel.prototype.addPin, () => {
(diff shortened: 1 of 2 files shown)


MadisonEhlers-Vertex (approved):

danschultz (commented):
This is inconsistent depending on the APIs that get called. What if a caller uses `setPins`?

danschultz:
I'd either recommend one of the following:

1. Modify the change event to emit a summary of what changed, like: `{ added: Pin[], removed: Pin[], pins: Pin[] }`.
2. Provide a closure over the event listener in Connect to detect if new pins are added, something like:


```ts
functions changes(added) {
  let last = [];
  return pins => {
    const new = // do a diff with `last` and pins`
    added(new);
    last = pins;
  }
}

function track(pins) {
  if (pins.length > 0) {
    // track new pins
  }
}

pins.onEntitiesChanged(changed(track));
```

amvertex:
> I'd either recommend one of the following:
> 
> 1. Modify the change event to emit a summary of what changed, like: `{ added: Pin[], removed: Pin[], pins: Pin[] }`. And update the SDK to compute this summary.
> 2. Provide a closure over the event listener in Connect to detect if new pins are added, something like:
> 
> ```ts
> function changes(added) {
>   let last = [];
>   return pins => {
>     const new = // do a diff with `last` and `pins`
>     added(new);
>     last = pins;
>   }
> }
> 
> function track(pins) {
>   if (pins.length > 0) {
>     // track new pins
>   }
> }
> 
> pins.onEntitiesChanged(changes(track));
> ```

I looked into comparing the `last` and `pins` in Connect initially, but I couldn't differentiate between when a user added a pin and when they loaded a snapshot that contained pins.

We could potentially modify the change event to emit a summary of what changed, but the summary would still need to be different between `setPin` and `setPins` so that we could detect when an individual pin was actually added, such that we would probably only populate the _added_ field when `setPin` was called and leave it empty for `setPins`.

MadisonEhlers-Vertex:
> We could potentially modify the change event to emit a summary of what changed, but the summary would still need to be different between `setPin` and `setPins` so that we could detect when an individual pin was actually added, such that we would probably only populate the _added_ field when `setPin` was called and leave it empty for `setPins`.

Could we only send this event when we are in draft mode? Wouldn't that solve the issue here?

PR #447, opened 2022-10-26
[PLAT-2148] Volume intersection query tool

jdm717:
## Summary

Adds a new `<vertex-viewer-box-query-tool>` element that supports dragging to draw boxes over portions of the viewer and performing various operations on the contents of that specified volume (this defaults to selection-based operations, but the `controller` supports replacing with any operation).

## Test Plan

- Add a `<vertex-viewer-box-query-tool>` underneath the viewer and verify that drag selection works
- Set the `defaultOperation` to `deselect` and verify that drag deselection works
- Set a custom operation on the `controller` using `setOperationTransform` and verify that operation works with the query
- Add and style a custom drag overlay

## Release Notes

N/A

## Possible Regressions

N/A - new functionality

## Dependencies

N/A


packages/viewer/src/components/viewer-intersection-query-tool/viewer-intersection-query-tool.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+import { VolumeIntersectionQueryController } from '../../lib/volume-intersection/controller';
+import { VolumeIntersectionQueryInteractionHandler } from '../../lib/volume-intersection/interactions';
+import {
+  QueryType,
+  VolumeIntersectionQueryDetails,
+  VolumeIntersectionQueryModel,
+} from '../../lib/volume-intersection/model';
+
+export type DefaultOperationType = 'select' | 'deselect';
```
MadisonEhlers-Vertex, line 12: nit: should this be named something like `IntersectionOperationType`?

packages/viewer/src/components/viewer-intersection-query-tool/viewer-intersection-query-tool.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+@Component({
+  tag: 'vertex-viewer-intersection-query-tool',
+  styleUrl: 'viewer-intersection-query-tool.css',
+  shadow: true,
+})
+export class ViewerIntersectionQueryTool {
```
MadisonEhlers-Vertex, line 19: Would it be too specific to name this thing `BoxQueryTool`? Or something like that?

packages/viewer/src/components/viewer-intersection-query-tool/viewer-intersection-query-tool.tsx
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * The default operation to perform when a drag has completed and the intersection
+   * query will be run. Defaults to `select`, and can be changed to `deselect`.
+   *
+   * The operation behavior for this intersection query tool can also be changed by
+   * providing a custom implementation of the `VolumeIntersectionQueryController`, or
+   * by using the `setOperationTransform` method of the default controller.
+   */
+  @Prop()
+  public defaultOperation: DefaultOperationType = 'select';
```
MadisonEhlers-Vertex, line 51: Should this just be named, `type` or `operationType`?

packages/viewer/src/lib/volume-intersection/controller.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  public async execute(): Promise<void> {
+    const screenBounds = this.model.getScreenBounds();
+    const type = this.model.getType();
+    this.viewer.cameraControls = this.previousViewerCameraControls ?? true;
+    this.model.complete();
+
+    if (
+      screenBounds != null &&
+      this.inFlightOperations < DEFAULT_CONCURRENT_VOLUME_QUERY_LIMIT
```
MadisonEhlers-Vertex, line 51: Should this be configurable in the controller? From my perspective this is something we could enforce on the backend rather than on the client. That might be a harder problem, so take this comment with a grain of salt. 😄

packages/viewer/src/components/viewer-intersection-query-tool/viewer-intersection-query-tool.tsx
jdm717: Yeah, I think it's worth considering some alternate names for this tool. I do like `box-query-tool` a bit better, so I'll update it!

packages/viewer/src/components/viewer-intersection-query-tool/viewer-intersection-query-tool.tsx
jdm717: I'd probably lean toward `operationType`, but kept the `default` to indicate that it can be overridden in the controller. Granted with the comment indicating that's possible, that may be redundant - I'll update this

packages/viewer/src/lib/volume-intersection/controller.ts
jdm717: Agreed, I think this is effectively a placeholder right now with the intent of revisiting in https://vertexvis.atlassian.net/browse/PLAT-2169. I wanted to add some sort of limitation here for internal testing, but I think this is something that's likely removed after that ticket. I think there are some other ideas around cancelling in-flight queries that we could try as well once we get a better idea of whether that's needed

amvertex (approved):

packages/viewer/src/lib/volume-intersection/controller.ts
```diff
@@ -0,0 +1,79 @@
...
+export const DEFAULT_CONCURRENT_VOLUME_QUERY_LIMIT = 5;
+
+export type OperationTransform = (
+  builder: SceneItemOperationsBuilder
+) => SceneItemOperationsBuilder;
+
+export class VolumeIntersectionQueryController {
+  private previousViewerCameraControls?: boolean;
+  private operationTransform: OperationTransform;
+  private inFlightOperations = 0;
```
MadisonEhlers-Vertex, line 15: nit: this could be configurable as a controller setting.

packages/viewer/src/lib/volume-intersection/model.ts
```diff
@@ -0,0 +1,97 @@
...
+      };
+    } else {
+      throw new Error(
+        'Failed to create query details, the start and end points must be set.'
+      );
+    }
+  }
+
+  private updateQueryType(): void {
+    if (this.startPoint != null && this.endPoint) {
```
MadisonEhlers-Vertex, line 90: nit: && `this.endPoint != null` ? Should they both be the same checks?

MadisonEhlers-Vertex (approved):

packages/viewer/src/lib/volume-intersection/controller.ts
jdm717: Yeah, my thinking was to come back and revisit this once we had more information from testing, but it's easy enough to make this configurable for now as well. I'll update this

packages/viewer/src/lib/volume-intersection/model.ts
jdm717: Yeah, no reason they shouldn't be, I must've forgotten to add the `!= null` -- will update

MadisonEhlers-Vertex (approved):

PR #448, opened 2022-11-01
[PLAT-1806] Update Twist to use offset values

jdm717:
## Summary

Updates the `TwistInteraction` to leverage `offsetX` and `offsetY` so the points are relative to the viewer rather than the window.

## Test Plan

- Render an element that reduces the size of the viewer to 50% of the window and verify that the center point of the Twist interaction is the center of the viewer

## Release Notes

N/A

## Possible Regressions

Twist interactions

## Dependencies

N/A


amvertex (approved):
Nice fix!

MadisonEhlers-Vertex (approved):

PR #450, opened 2022-11-08
[PLAT-2189] Apply partial SceneViewState

amvertex:
## Summary
We want to support the ability to apply a subset of a SceneViewState's features to the current scene, instead of the full SceneViewState.

## Test Plan
Verify that a user can apply a partial SceneViewState

## Release Notes
None

## Possible Regressions
Viewing snapshots

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/scenes/scene.ts
+++ b/packages/viewer/src/lib/scenes/scene.ts
@@ -224,6 +224,17 @@ export type TerminalItemOperationBuilder =
 export type ImageScaleProvider = () => Point.Point | undefined;
 
 /**
+ * The features of a scene view state that can be applied to the current scene
+ */
+export type SceneViewStateFeature =
+  | 'camera'
+  | 'cross_section'
+  | 'material_overrides'
+  | 'selection'
+  | 'transforms'
+  | 'visibility';
+
+/**
  * A class that represents the `Scene` that has been loaded into the viewer. On
  * it, you can retrieve attributes of the scene, such as the camera. It also
  * contains methods for updating the scene and performing requests to rerender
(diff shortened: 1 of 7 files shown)


packages/viewer/src/lib/scenes/scene.ts
```diff
@@ -259,6 +259,28 @@ export class Scene {
     );
   }
 
+  /**
+   * Applies the specified features of the provided scene view state to the scene.
+   */
+  public async applyPartialSceneViewState(
+    sceneViewStateId: UUID.UUID,
+    featuresToApply: string[],
```
MadisonEhlers-Vertex, line 267: I don't think that we should accept any payload from the client - we should create a new feature type that is public facing to expose to developers which features are accessible to partially apply.

packages/viewer/src/lib/scenes/scene.ts
```diff
@@ -259,6 +270,28 @@ export class Scene {
     );
   }
 
+  /**
+   * Applies the specified features of the provided scene view state to the scene.
+   */
+  public async applyPartialSceneViewState(
```
MadisonEhlers-Vertex, line 276: Should this be a separate method, or a modification to the existing method? @amvertex @jdm717 Any thoughts?

MadisonEhlers-Vertex: Beyond this, lgtm!

amvertex: I personally liked it as a separate method, but it would be simple to change it to a modification of the existing method.

jdm717: I think I'd lean toward the separate method for now as well. Makes it a little more discoverable, plus I think this parameter order makes more sense than having the `SceneExecutionOptions` second, so it'd be a breaking change to the `applySceneViewState` method

PR #451, opened 2022-11-09
[PLAT-2161] Scene tree error improvements

jdm717:
## Summary

Makes a few small adjustments to the error handling and errors emitted by the scene-tree component, specifically:
- "Cancelled" controller connections will no longer throw an error
  - This is the case when the component is disconnected very quickly after being initially connected
- Unauthenticated gRPC responses will now map to an `UNAUTHORIZED` error code, rather than `UNKNOWN`
- The name of the `SceneTreeErrorCode` enum value is now included with the error details

## Test Plan

- Verify that quickly unloading and loading the scene-tree does not throw errors
- Verify that errors include additional detail

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/scene-tree/lib/controller.ts
+++ b/packages/viewer/src/components/scene-tree/lib/controller.ts
@@ -28,7 +28,14 @@ import {
 import { Disposable, EventDispatcher, Listener } from '@vertexvis/utils';
 
 import { MetadataKey } from '../interfaces';
-import { SceneTreeErrorCode, SceneTreeErrorDetails } from './errors';
+import {
+  SceneTreeConnectionCancelledError,
+  SceneTreeError,
+  SceneTreeErrorCode,
+  SceneTreeErrorDetails,
+  SceneTreeOperationFailedError,
+  SceneTreeUnauthorizedError,
+} from './errors';
 import { isGrpcServiceError } from './grpc';
 import { decodeSceneTreeJwt } from './jwt';
 import { fromNodeProto, isLoadedRow, Row } from './row';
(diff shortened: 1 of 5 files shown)


packages/viewer/src/components/scene-tree/lib/errors.ts
```diff
@@ -29,5 +33,43 @@ function getSceneTreeErrorMessage(code: SceneTreeErrorCode): string {
...
+    // Allows for `instanceof` checks.
+    Object.setPrototypeOf(this, SceneTreeError.prototype);
+  }
+}
+
+export class SceneTreeUnauthorizedError extends SceneTreeError {
+  public constructor(message: string, e?: Error) {
+    super(message, e);
+
+    // Allows for `instanceof` checks.
```
MadisonEhlers-Vertex, line 54: that's interesting you have to set this to allow it. I didn't know this.

jdm717: Yeah, definitely an interesting detail! We had to do similar [here](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/lib/errors.ts)

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -84,6 +91,12 @@ export interface ConnectionFailedState {
   sceneViewId: string;
 }
 
+export interface ConnectionCancelledState {
+  type: 'cancelled';
+  jwtProvider?: JwtProvider;
+  sceneViewId?: string;
```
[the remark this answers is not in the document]
jdm717: It's not currently, just added for consistency with the `disconnected` state. Not opposed to removing it though, since I don't see as much value in the `cancelled` state

MadisonEhlers-Vertex (approved):

PR #453, opened 2022-11-14
[PLAT-2169] Box query tool polish

jdm717:
## Summary

Adds some UX/interaction polish to the `<vertex-viewer-box-query-tool>` element, specifically:
- Corrects behavior when switching between `perspective` and `orthographic` camera types
- Adds cursors to indicate when an operation is in-flight
  - Additionally, this has now been limited to a single in-flight operation
- Fixes an issue where moving the mouse outside the viewer would cause the box to behave a bit erratically 
- Fixes an issue where the ability to draw additional boxes while an operation was in-flight would visually appear to work
- Adds support for a `mode` on the query tool, which can be set to `exclusive`, `inclusive`, or `undefined` to limit to either mode (with `undefined` being default/supporting both)

## Test Plan

- Test the above mentioned improvements/changes

## Release Notes

N/A

## Possible Regressions

Cursor management with the query tool, combined `exclusive`/`inclusive` query behavior

## Dependencies

N/A


packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.spec.tsx
```diff
@@ -216,6 +217,91 @@ describe('vertex-viewer-box-query-tool', () => {
...
+      });
+
+    await drawInclusiveBox(page, viewer);
+    window.dispatchEvent(new MouseEvent('pointerup'));
+
+    await drawExclusiveBox(page, viewer);
+    window.dispatchEvent(new MouseEvent('pointerup'));
+
+    await page.waitForChanges();
+    await Async.delay(10);
```
MadisonEhlers-Vertex, line 251: Does the test fail without a manual delay?

packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
```diff
@@ -11,6 +11,8 @@ import {
 
 export type VolumeIntersectionQueryType = 'select' | 'deselect';
 
+export type VolumeIntersectionQueryMode = 'exclusive' | 'inclusive';
```
MadisonEhlers-Vertex, line 14: Should we consider adding a third option, `both`? Rather than having `undefined` function as both?

packages/viewer/src/lib/volume-intersection/controller.ts
```diff
@@ -67,11 +73,14 @@ export class VolumeIntersectionQueryController {
...
+        this.operationInFlight = false;
+        this.executeComplete.emit();
       }
-    } else if (this.inFlightOperations >= this.maxInFlightOperations) {
+    } else if (this.operationInFlight) {
       throw new Error(
-        `Unable to perform volume intersection query due to the limit of ${this.maxInFlightOperations}.`
+        `Unable to perform volume intersection query as there is already one in-flight.`
       );
     }
```
MadisonEhlers-Vertex, line 85: not related to this PR, but should we add an `else` here to handle if the screenBounds _are_ null? Log or throw an error or something?

packages/viewer/src/lib/volume-intersection/interactions.ts
```diff
@@ -1,51 +1,104 @@
...
+    this.enabled = false;
+  }
+
+  public enable(): void {
+    this.enabled = true;
   }
 
   private handleDragBegin(event: PointerEvent): void {
-    if (event.buttons === 1 && !this.isInteracting) {
+    if (this.enabled && event.buttons === 1 && !this.isInteracting) {
```
MadisonEhlers-Vertex, line 57: 💡 Just an idea here, I often forget what all of the button numbers map to. Do you think it'd be a good idea to add some utils around the PointerEvent buttons to make it more readable? i.e. 

```ts
private function isPrimaryClick(event: PointerEvent): boolean {
  return event.buttons === 1;
}
```
Just an idea to make some of the code more readable. Not that important.

packages/viewer/src/lib/volume-intersection/interactions.ts
```diff
@@ -1,51 +1,104 @@
...
+
+      this.crosshairCursorDisposable?.dispose();
+      this.addWaitCursor();
+      try {
+        await this.controller.execute();
+      } finally {
+        this.waitCursorDisposable?.dispose();
+        this.addCrosshairCursor();
+      }
+    }
```
MadisonEhlers-Vertex, line 92: What should happen if the drag ends when it's not enabled?

packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.spec.tsx
jdm717: Yeah, unfortunately. At least the first delay is necessary to "complete" the first alteration and allow for the third attempt to complete successfully. Looks like I'm able to remove the second await though, so I'll do that

packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
jdm717: Yeah, I thought about having a `both`, but it felt like that didn't communicate enough about the types that would be present. Though the same argument could probably be made for `undefined`. I also considered `exclusive-and-inclusive`, but that felt a bit wordy. Open to ideas here 🙂

packages/viewer/src/lib/volume-intersection/controller.ts
jdm717: Sure, I can take a look at adding that. Probably at least worth an error

packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
mikemccarty-vertex: What would `both` do?

As defined right now:
`inclusive == In | Intersects`
`exclusive == In & ~(Intersects | Out)`

jdm717: @mikemccarty-vertex the `both` would just refer to what queries the tool itself is capable of. I.e. with `both` as the query mode, dragging left-to-right would perform an `exclusive` query and right-to-left would perform an `inclusive` query. With this change, the `inclusive` and `exclusive` modes mean that no matter which way you drag, you'll get that type of query

MadisonEhlers-Vertex (approved):

PR #460, opened 2022-12-14
[PLAT-2334] FovY Support for Perspective Cameras

MadisonEhlers-Vertex:
## Summary

Introduces support for fovY on a perspective camera. The main plumbing was here for this, was just not mapped via protos. 

## Test Plan

Verify that the fovY can be passed into perspective cameras. Valid ranges are float values between 1 & 179, which are enforced server-side

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/308

### DIFF
--- a/packages/stream-api/package.json
+++ b/packages/stream-api/package.json
@@ -36,7 +36,7 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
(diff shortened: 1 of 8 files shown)


jdm717 (approved):
LGTM!

PR #464, opened 2023-01-09
[PLAT-2131] Text pins improvements

amvertex:
## Summary
Improvements to text markup, specifically:
- Auto-focuses on new pins when the pin is created such that the user can start typing right away
- Includes a visible label line when the pin is created instead of the label touching the anchor. The label is always positioned towards the center of the screen relative to the anchor point and should be present on the screen.
- Prevents typing 'f' in the pin label from triggering a fit all
- Pressing "Enter" now submits the label text 
- Pressing either "Shift+Enter" or "Control/Command+Enter" now adds a new text line in the existing label

## Test Plan
Verify the items listed above

## Release Notes
None

## Possible Regressions
Text markup

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/viewer-pin-tool/utils.ts
+++ b/packages/viewer/src/components/viewer-pin-tool/utils.ts
@@ -30,7 +30,9 @@ export function translatePointToScreen(
 export function translatePointToRelative(
   pt: Point.Point,
   canvasDimensions: Dimensions.Dimensions,
-  includeOffset?: boolean
+  includeOffset?: boolean,
+  xOffset?: number,
+  yOffset?: number
 ): Point.Point {
   const verticalScaleFactor = 1 / canvasDimensions.height;
   const horizontalScaleFactor = 1 / canvasDimensions.width;
@@ -44,7 +46,7 @@ export function translatePointToRelative(
 
   if (includeOffset) {
     const offset = Point.scale(
-      Point.create(20, 20),
+      Point.create(xOffset, yOffset),
       horizontalScaleFactor,
       verticalScaleFactor
     );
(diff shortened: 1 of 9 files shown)


packages/viewer/src/components/viewer-pin-tool/utils.ts
```diff
@@ -29,13 +29,39 @@ export function translatePointToScreen(
...
+    // when placing it to the left or above to ensure that part of the label line is visible
+    const centerIsToTheRightInXDirection = canvasCenterPoint.x > pt.x;
+    const xOffset = centerIsToTheRightInXDirection ? offset.x : -3 * offset.x;
+
+    const centerIsBelowInYDirection = canvasCenterPoint.y > pt.y;
+    const yOffset = centerIsBelowInYDirection ? offset.y : -3 * offset.y;
+
+    const offsetPoint = Point.add(point, Point.create(xOffset, yOffset));
+
+    return constrainRelativePoint(offsetPoint, Range.create(-0.5, 0.5));
```
jdm717, line 63: Wondering if we could make these offset values configurable from the component? Seems like it would be nice to have a bit of control over how far the pin label is from the initial point. Maybe something that could be read in from a CSS variable? I.e. `viewer-pin-tool-initial-label-offset-x: 20px` or something similar.

Additionally, any specific reason this is scaled by `-3` if it needs to flip to the other side? Thinking we could run into issues with a constant there if the label starts off with a larger width.

jdm717 (commented):
Overall LGTM! Had one minor thought on the offset values

packages/viewer/src/components/viewer-pin-tool/utils.ts
amvertex: I added a CSS variable, good idea. 

The point that this returns corresponds to the upper left corner of the label, so if the label is placed above or to the left, the label line is smaller since the distance from the anchor point to the upper left corner of the label crosses the body of the label. Multiplying the offset value by 3 helps to ensure that we have a visible label line when a pin is first placed.  Since it's relative to the value of the CSS variable, I think a user could adjust that variable if they also adjusted the initial width of the label.

jdm717 (approved):

PR #467, opened 2023-01-19
[PLAT-2474] Prevent highlighting text in label

amvertex:
## Summary
When moving a text markup label around the screen, the text in other labels was being highlighted.  Further, it was requested that we reduced the extra space on the right side of the label.

## Test Plan
Verify that moving a text markup label no longer highlights the text in other labels

## Release Notes
None

## Possible Regressions
Text markup

## Dependencies
None


jdm717 (approved):
LGTM, thanks for fixing the typo/comment inconsistencies as well!

PR #468, opened 2023-01-20
[PLAT-2258] Subscription Retries

MadisonEhlers-Vertex:
## Summary

These changes introduce a retry on the subscription call when the handshake timeout fails within the allotted timeout. This is to improve resiliency for first time connections, and attempts so establish a handshake more consistently.


amvertex (approved):
Nice job!

PR #469, opened 2023-01-24
[PLAT-2244] Correct transform widget position initialization

jdm717:
## Summary

Corrects an issue where the `startingTransform` was not set when loading the transform widget with a starting position.

## Test Plan

- Verify the transform widget works as expected when initialized with a position

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.spec.tsx
+++ b/packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.spec.tsx
@@ -258,6 +258,117 @@ describe('vertex-viewer-transform-widget', () => {
     expect(onInteractionStarted).toHaveBeenCalled();
   });
 
+  it.only('performs a transform when initialized with a position', async () => {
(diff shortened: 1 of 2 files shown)


packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.spec.tsx
```diff
@@ -258,6 +258,117 @@ describe('vertex-viewer-transform-widget', () => {
     expect(onInteractionStarted).toHaveBeenCalled();
   });
 
+  it.only('performs a transform when initialized with a position', async () => {
```
MadisonEhlers-Vertex, line 261: Is setting this test to be `.only` intentional here?

jdm717: Nope! Removed

MadisonEhlers-Vertex (approved):

PR #470, opened 2023-01-24
[PLAT-2437] Add executeAborted event for box-query tool

jdm717:
## Summary

Adds an `executeAborted` event dispatcher to the volume intersection controller in the case that we receive too many items that match the query. This can then be handled on the consumer-side rather than logging an error for the expected behavior.

## Test Plan

- Verify that there is no console.error log for queries that are aborted
- Verify the listener registered using the `onExecuteAborted` method is called with the error

## Release Notes

N/A

## Possible Regressions

Volume intersection query error handling

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
+++ b/packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
@@ -252,7 +252,7 @@ export class ViewerBoxQueryTool {
   private setDefaultClearAndSelectOperation(): void {
     this.controller?.setOperationTransform((builder) => builder.select());
     this.controller?.setAdditionalTransforms([
-      (op) => op.where((q) => q.all()).deselect(),
+      (op) => op.where((q) => q.withSelected()).deselect(),
     ]);
   }
 
--- a/packages/viewer/src/lib/volume-intersection/controller.ts
+++ b/packages/viewer/src/lib/volume-intersection/controller.ts
@@ -16,14 +16,19 @@ export type AdditionalTransform = (
   executor: SceneItemQueryExecutor
 ) => TerminalItemOperationBuilder;
 
+export interface CompleteExecutionDetails {
+  aborted: boolean;
+}
+
 export class VolumeIntersectionQueryController {
   private previousViewerCameraControls?: boolean;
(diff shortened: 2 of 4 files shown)


packages/viewer/src/lib/volume-intersection/controller.ts
```diff
@@ -108,8 +114,15 @@ export class VolumeIntersectionQueryController {
           ])
           .execute();
       } catch (e) {
-        console.error('Failed to perform volume intersection query', e);
-        throw e;
+        if (
+          e instanceof StreamRequestError &&
+          e.summary?.toLocaleLowerCase().includes('operation aborted')
```
jdm717, line 119: May be a good tech-talk topic, but we don't have any other context around this error message currently, so the summary is the only way to determine this is intentional. Should we be sending more information with these request errors from FSS? I.e. should these messages also include codes that we can leverage?

packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
```diff
@@ -252,7 +252,7 @@ export class ViewerBoxQueryTool {
   private setDefaultClearAndSelectOperation(): void {
     this.controller?.setOperationTransform((builder) => builder.select());
     this.controller?.setAdditionalTransforms([
-      (op) => op.where((q) => q.all()).deselect(),
+      (op) => op.where((q) => q.withSelected()).deselect(),
```
MadisonEhlers-Vertex, line 255: We should figure out why this change is necessary.

packages/viewer/src/lib/volume-intersection/controller.ts
MadisonEhlers-Vertex: I agree, we should be using status codes, and sending those back to the client. Worth a discussion and maybe a ticket.

MadisonEhlers-Vertex (approved):

packages/viewer/src/components/viewer-box-query-tool/viewer-box-query-tool.tsx
jdm717: Agreed, I created https://vertexvis.atlassian.net/browse/PLAT-2506 to capture this issue

PR #477, opened 2023-03-06
[PLAT-2576] Correct skipping of correlation IDs on skipped frames

jdm717:
## Summary

Corrects a case where skipping the loading/drawing of a frame due to either passed predicate being false (specifically used for resizing behavior), or due to an attempt to draw a frame out of sequence. This approach will track the correlation IDs of frames coming through, and append any skipped values to the next successfully drawn frame.

## Test Plan

- Verify that correlation IDs sent come through on the `frameDrawn` event
  - My approach for reproducing this bug has been to add an alteration on `sceneReady` with a custom ID, then log out correlation IDs as I receive `frameDrawn` events, then resizing the browser while refreshing

## Release Notes

N/A

## Possible Regressions

Correlation IDs on the `frame` attribute + events

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/rendering/canvas.ts
+++ b/packages/viewer/src/lib/rendering/canvas.ts
@@ -112,6 +112,29 @@ export function measureCanvasRenderer(
 }
 
 export function createCanvasRenderer(): CanvasRenderer {
+  let skippedCorrelationIds: string[] = [];
+
+  async function trackCorrelationIds(
+    data: DrawFrame,
+    frame: Frame | undefined
+  ): Promise<Frame | undefined> {
+    if (frame == null) {
+      skippedCorrelationIds = [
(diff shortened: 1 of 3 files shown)


packages/viewer/src/lib/rendering/canvas.ts
```diff
@@ -112,6 +112,27 @@ export function measureCanvasRenderer(
 }
 
 export function createCanvasRenderer(): CanvasRenderer {
+  let accumulatedCorrelationIds: string[] = [];
+
+  async function addCorrelationIds(
+    frame: Frame | undefined
+  ): Promise<Frame | undefined> {
```
danschultz, line 119: This doesn't appear to be `async`.

danschultz (approved):

PR #482, opened 2023-04-24
[PLAT-2862] Using a unique ID for comparing frames, since the copy may contain additional correlation IDs

MadisonEhlers-Vertex:
## Summary

These changes add a unique identifier on a frame. We then check the ID of the frame instead of doing an object compare on these frames. This solves the problem of not updating the frame on a reconnect when the frames are in fact equal, but have mismatching correlation IDs which resulted in unnecessarily discarding of frames. 

## Test Plan

- Open a scene in the viewer
- Interact with it for a while to increase the sequence count on a frame (Rotate for 10-15 seconds)
- Let the viewer sit for a minute or so until you identify that a websocket reconnection has occurred (This can be observed in the network tab when you see a new connection to the web socket)
- Verify the scene you have open is immediately able to be interacted with, and is not stale.

### DIFF
--- a/packages/viewer/src/lib/types/frame.ts
+++ b/packages/viewer/src/lib/types/frame.ts
@@ -32,9 +32,7 @@ export class Frame {
     private readonly depthBufferBytes: Uint8Array | undefined,
     private readonly featureMapBytes: Uint8Array | undefined,
     private readonly id = UUID.create()
-  ) {
-    this.id = UUID.create();
-  }
+  ) {}
 
   public getId(): UUID.UUID {
     return this.id;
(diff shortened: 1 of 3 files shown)


mikemccarty-vertex (approved):

packages/viewer/src/lib/types/frame.ts
```diff
@@ -29,8 +30,15 @@ export class Frame {
     public readonly image: FrameImage,
     public readonly scene: FrameScene,
     private readonly depthBufferBytes: Uint8Array | undefined,
-    private readonly featureMapBytes: Uint8Array | undefined
-  ) {}
+    private readonly featureMapBytes: Uint8Array | undefined,
+    private readonly id = UUID.create()
+  ) {
+    this.id = UUID.create();
```
jdm717, line 36: Looks like this will clear out any value passed to the constructor, can we just rely on the defaulted value?

jdm717 (approved):
LGTM!

PR #486, opened 2023-04-27
[PLAT-2519] Additional events for analytics in Connect

jdm717:
## Summary

Adds a couple events for tracking time to the first visible node in the `<vertex-scene-tree>`, as well as adding an emit of the `ConnectionFailedStatus` in the viewer.

## Test Plan

- Verify that the `firstRowRendered` event is fired when the first row of the scene-tree has rendered
- Verify that the `connectionChange` event in the viewer is fired when a connection fails with a reason

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


amvertex (approved):
Nice job with this!

PR #488, opened 2023-05-03
[PLAT-2787] Hit Result indicator

jdm717:
## Summary

Adds a new `<vertex-viewer-hit-result-indicator>` element that can be used to display visuals for the position, normal, and computed plane of a hit result. This component leverages WebGL to draw the indicator, reusing some of the types added with the `<vertex-viewer-transform-widget>`, and makes updates to extract some of the shared behavior between the two.

https://user-images.githubusercontent.com/5252480/236006708-97e1eb69-8979-4ead-968c-ac7d81cf4ce0.mov

## Test Plan

- Verify that the transform widget continues to work as expected
  - Setting position, clearing position, etc
- Place multiple `<vertex-viewer-hit-result-indicator>`s using data from hit results, and verify the position, normal, and computed plane visually match the data in the hit result

## Release Notes

N/A

## Possible Regressions

Transform widget

## Dependencies

N/A


packages/viewer/src/components/viewer-hit-result-indicator/lib/point.ts
```diff
@@ -0,0 +1,79 @@
...
+  const xRay = Ray.create({
+    origin: position,
+    direction: worldX,
+  });
+  const yRay = Ray.create({
+    origin: position,
+    direction: worldY,
+  });
+
+  const bottomLeft = Ray.at(xRay, -(rectangleSize / 2.5));
```
MadisonEhlers-Vertex, line 32: What is significant here about 2.5?

packages/viewer/src/components/viewer-hit-result-indicator/readme.md
```diff
@@ -0,0 +1,29 @@
+# vertex-viewer-hit-result-indicator
+
+
```
MadisonEhlers-Vertex, line 3: Can you include an example of how to use this with hit results here?

packages/viewer/src/components/viewer-transform-widget/widget.ts
```diff
@@ -46,22 +47,7 @@ export interface DisabledAxis {
...
-// components to determine the relative size of the meshes.
-// This attempts to keep the widget approximately the same
-// size as zooming occurs.
-export const DEFAULT_ORTHOGRAPHIC_MESH_SCALAR = 0.00625;
-
-export class TransformWidget implements Disposable {
-  private reglCommand?: regl.Regl;
-
-  private viewport: Viewport;
+export class TransformWidget extends ReglComponent {
```
MadisonEhlers-Vertex, line 50: 👍

packages/viewer/src/components/viewer-hit-result-indicator/viewer-hit-result-indicator.tsx
```diff
@@ -0,0 +1,254 @@
...
+      this.setupIndicator(this.canvasRef);
+    }
+
+    readDOM(() => {
+      const hostStyles = window.getComputedStyle(this.hostEl);
+
+      this.arrowColor = hostStyles
+        .getPropertyValue('--viewer-hit-result-indicator-arrow-color')
+        .trim()
+        .replace(/["]*/g, '');
```
MadisonEhlers-Vertex, line 68: nit: this regex is reused and could be pulled out to a const.

packages/viewer/src/components/viewer-hit-result-indicator/viewer-hit-result-indicator.tsx
```diff
@@ -0,0 +1,254 @@
...
+      {
+        arrow: this.arrowColor,
+        plane: this.planeColor,
+      },
+      {
+        plane: this.planeOpacity,
+      }
+    );
+
+    if (this.position != null) {
```
MadisonEhlers-Vertex, line 220: In what cases do we expect the position to be null? Should the `else` condition here include a warning or error?

packages/viewer/src/components/viewer-hit-result-indicator/lib/point.ts
jdm717: Nothing specific, just an arbitrary value similar to what we do for the arrow/axis point computation. That being said, we could probably at least pull this into a constant in the `indicator.ts` file when we pass this value and remove the constant here

MadisonEhlers-Vertex: Good idea - let's do that. Especially if it makes sense for all of these to be consistent.

MadisonEhlers-Vertex (approved):

PR #490, opened 2023-05-05
[PLAT-2808] Translation & Rotation delta helpers for two positions & normals

MadisonEhlers-Vertex:

### DIFF
--- a/packages/viewer/src/components/viewer-transform-widget/util.ts
+++ b/packages/viewer/src/components/viewer-transform-widget/util.ts
@@ -199,6 +199,12 @@ function computeTranslation(
   return rotatedDelta.x + rotatedDelta.y + rotatedDelta.z;
 }
 
+/**
+ * This helper makes a translation matrix from the negated position
+ * @param rotationAxis
+ * @param current
+ * @returns
+ */
 export function computeRotation(
   rotationAxis: Quaternion.Quaternion,
   current: Matrix4.Matrix4
--- a/packages/viewer/src/index.ts
+++ b/packages/viewer/src/index.ts
@@ -1,9 +1,12 @@
 /**
  * @module Viewer
  */
+import * as Matricies from './lib/rendering/matrices';
 import * as ColorMaterial from './lib/scenes/colorMaterial';
 
 export { ColorMaterial };
+
+export { Matricies };
 
 export * from './components';
 export * from './components/scene-tree/lib';
--- a/packages/viewer/src/lib/rendering/matrices.ts
+++ b/packages/viewer/src/lib/rendering/matrices.ts
@@ -3,7 +3,7 @@
  * @see https://github.com/openframeworks/openFrameworks/blob/356612673304e2f8b0584bc4080b58ee36db62f7/libs/openFrameworks/math/ofMatrix4x4.cpp
  */
 
-import { Angle, Matrix4, Vector3 } from '@vertexvis/geometry';
+import { Angle, Matrix4, Quaternion, Vector3 } from '@vertexvis/geometry';
 
 import { FrameCamera } from '../types';
 import { isOrthographicFrameCamera } from '../types/frameCamera';
@@ -154,4 +154,30 @@ export function makeLookAtMatrix(
     0  , 0  , 0  , 1
   ];
   /* eslint-enable prettier/prettier */
+}
+
+export function experimentalMakeTransformationMatrix(
+  normal1: Vector3.Vector3,
+  position1: Vector3.Vector3,
+  normal2: Vector3.Vector3,
+  position2: Vector3.Vector3
+): Matrix4.Matrix4 {
+  const translationDeltaMatrix = Matrix4.makeTranslation(
+    Vector3.subtract(position2, position1)
+  );
+  const angle = Vector3.angleTo(normal2, normal1);
+
+  const axisDirection = Vector3.normalize(Vector3.cross(normal1, normal2));
+
+  const translation = Matrix4.makeTranslation(position1);
+
+  const rotationMatrix = Matrix4.makeRotation(
+    Quaternion.fromAxisAngle(axisDirection, angle)
+  );
+
+  const r = Matrix4.multiply(
+    Matrix4.multiply(translation, rotationMatrix),
+    Matrix4.makeTranslation(Vector3.negate(position1))
+  );
+  return Matrix4.multiply(translationDeltaMatrix, r);
 }
--- a/packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
+++ b/packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.tsx
@@ -251,7 +251,7 @@ export class ViewerTransformWidget {
     newRotation: Euler.Euler,
     oldRotation?: Euler.Euler
   ): void {
-    this.currentTransform = this.getTransformForHewRotation(newRotation);
+    this.currentTransform = this.getTransformForNewRotation(newRotation);
     this.startingTransform = this.currentTransform;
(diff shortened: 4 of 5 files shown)


packages/viewer/src/lib/rendering/matrices.ts
```diff
@@ -155,3 +155,29 @@ export function makeLookAtMatrix(
   ];
   /* eslint-enable prettier/prettier */
 }
+
+export function experimentalMakeTransformationMatrix(
```
MadisonEhlers-Vertex, line 159: I will plan on including this in the web-sdk, but I do not think this is the right place - and obviously once we get through the review process, I will plan to remove `experimental` from the name.

Looking for feedback on where to store this? I was thinking in the Matrix4.ts or potentially in the transform-widget `utils.ts` file

packages/viewer/src/lib/rendering/matrices.ts
```diff
@@ -155,3 +155,29 @@ export function makeLookAtMatrix(
...
+  position1: Vector3.Vector3,
+  normal2: Vector3.Vector3,
+  position2: Vector3.Vector3
+): Matrix4.Matrix4 {
+  const translationDeltaMatrix = Matrix4.makeTranslation(
+    Vector3.subtract(position2, position1)
+  );
+  const angle = Vector3.angleTo(normal2, normal1);
+
+  const axisDirection = Vector3.normalize(Vector3.cross(normal1, normal2));
```
mikemccarty-vertex, line 170: Cross product of (anti-)parallel vectors leads to zero-length vectors so you will need to test `angle` for a value of nearly 0.0 or 180.0. If so, `rotationMatrix == identity`.

[The intuition here is the length of the cross-product vector is the area of the implied parallelogram formed by the two vectors.](https://www.geogebra.org/m/psMTGDgc)

packages/viewer/src/lib/rendering/matrices.ts
```diff
@@ -155,3 +155,29 @@ export function makeLookAtMatrix(
...
+
+  const rotationMatrix = Matrix4.makeRotation(
+    Quaternion.fromAxisAngle(axisDirection, angle)
+  );
+
+  const r = Matrix4.multiply(
+    Matrix4.multiply(translation, rotationMatrix),
+    Matrix4.makeTranslation(Vector3.negate(position1))
+  );
+  return Matrix4.multiply(translationDeltaMatrix, r);
```
mikemccarty-vertex, line 182: The translation matrices are a little overly complicated I believe.  `translation` represents `+ position1` while `translationDeltaMatrix` represents `position2 - position1`.  Simplifying, you get:
```
(position2 - position1) + position1
== position2
```

I.e. the two consecutive left-multiplies by translation matrices can be combined into one simpler matrix.  This works because when left-multiplying by translation matrices, the [rotation part of the right matrix has no affect on the translation values of the left matrix](https://github.com/Vertexvis/vector-math-lib-java/blob/b252e536b873ef65eb5046c50dfcf7372cff79c4/src/main/java/com/vertexvis/vectormath/AffineTransform3f.java#L31).

packages/viewer/src/components/viewer-transform-widget/util.ts
```diff
@@ -199,6 +199,12 @@ function computeTranslation(
   return rotatedDelta.x + rotatedDelta.y + rotatedDelta.z;
 }
 
+/**
+ * This helper makes a translation matrix from the negated position
+ * @param rotationAxis
+ * @param current
+ * @returns
+ */
```
jdm717, line 207: nit: Could we update this description (and also potentially rename the `rotationAxis` param)? This method is actually applying the rotation at the current widget position, then translating it back to convert it to a world delta. The `rotationAxis` naming is a bit misleading, since it's actually a quaternion

packages/viewer/src/index.ts
```diff
@@ -2,9 +2,12 @@
  * @module Viewer
  */
 import * as ColorMaterial from './lib/scenes/colorMaterial';
+import * as TransformationDelta from './lib/transforms/transformation-delta';
```
jdm717, line 5: Could we consider adding an `index.ts` file to the `./lib/transforms` directory and exporting this there? I think we'll also want to export the `TransformController`, so it'd be nice to have an index file there

packages/viewer/src/lib/transforms/transformation-delta.ts
```diff
@@ -0,0 +1,77 @@
+import { Matrix4, Quaternion, Vector3 } from '@vertexvis/geometry';
+
+export const ALMOST_ONE = 0.9999;
+/**
+ * There are an infinite number of orthogonal vectors, this function is used to choose one.
+ * @param normal
+ * @returns
+ */
```
jdm717, line 8: nit: Could we update this comment to explain a bit more about this method? Maybe something along the lines of 
```typescript
/**
 * For any single vector, there are an infinite number of potential orthogonal vectors. This function will determine 
 * one orthogonal vector by crossing the provided vector with a unit vector in the positive X, Y, or Z directions.
 */
```

packages/viewer/src/lib/transforms/transformation-delta.ts
```diff
@@ -0,0 +1,77 @@
...
+  const vector = Vector3.create(x, y, z);
+  return Vector3.normalize(Vector3.cross(normal, vector));
+}
+
+/**
+ * Computes the rotation matrix for two normal vectors.
+ * @param normal1
+ * @param normal2
+ * @returns
+ */
```
jdm717, line 30: nit: Same as above here, could we explain a bit further how the rotation matrix is computed? I.e. explain that it's computing the rotation matrix such that both normals are anti-parallel

packages/viewer/src/lib/transforms/transformation-delta.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+/**
+ * Computes the translation & rotation matrix delta between two world positions and two normals vectors.
+ * @param normal1
+ * @param position1
+ * @param normal2
+ * @param position2
+ * @returns
+ */
```
jdm717, line 63: nit: Same here, could we update the description to reflect that the computed translation and rotation matrix will translate + rotate such that position 1 is moved to position 2 and the normals are rotated to be anti-parallel?

jdm717 (commented):
Overall LGTM! Had a few thoughts on some of the method comments or organization, but otherwise looks good to go

packages/viewer/src/lib/transforms/transformation-delta.ts
```diff
@@ -23,10 +24,14 @@ export function chooseOrthogonalVector(
 }
 
 /**
- * Computes the rotation matrix for two normal vectors.
+ * Computes the rotation matrix for two normal vectors. If both normals are anti-parallel,
```
jdm717, line 27: nit: Looks like there's a duplicate reference to `anti-parallel` here, could we change this to `If both normals are neither parallel or anti-parallel`?

packages/viewer/src/lib/transforms/transformation-delta.ts
```diff
@@ -53,13 +58,19 @@ export function computeRotationMatrix(
     );
   }
 }
+
 /**
- * Computes the translation & rotation matrix delta between two world positions and two normals vectors.
+ * Computes the translation & rotation matrix delta between two world positions and two normals vectors
```
jdm717, line 63: minor nit: ~~`two normals vectors`~~ -> `two normals`/`two normal vectors`

jdm717 (approved):
LGTM! Had two more small wording changes, but otherwise the updated comments look great

jdm717 (approved):
LGTM! Had two more small wording changes, but otherwise the updated comments look great

PR #495, opened 2023-05-09
[PLAT-2787] Fix WebGL warnings

jdm717:
## Summary

Fixes an issue with our components that leverage WebGL where the buffers created were not large enough for draw calls. Looking at the internals of the `regl-shape` library, this appears to be related to the typed arrays there being created using the length of the provided `pointsArray`, and there is a specific callout in the README around ensuring the `count` provided is less than half the length of the provided `pointsArray` (which we used exactly half). The specific value of `4` appears to relate to the fill colors, as providing a value of `2` results in a loss of fill colors in certain scenarios, where a value of `3` results in the same warning (in addition to other issues related to the odd-length array).

Specific warning:
```
 GL_INVALID_OPERATION: Vertex buffer is not big enough for the draw call
```

## Test Plan

- Render the transform widget and verify that these errors no longer appear
- Render hit result indicators and verify that these errors no longer appear
- Verify the fill colors for both of the above elements continue to work as expected

## Release Notes

N/A

## Possible Regressions

Transform widget and hit result indicators

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-transform-widget/__tests__/widget.spec.ts
+++ b/packages/viewer/src/components/viewer-transform-widget/__tests__/widget.spec.ts
@@ -36,6 +36,7 @@ import {
   makePerspectiveFrame,
 } from '../../../testing/fixtures';
 import { TransformWidget } from '../widget';
+import { DrawablePoints } from '../../../lib/transforms/drawable';
 
 type MockShapeBuilder = jest.Mock<{ createShape: jest.Mock }>;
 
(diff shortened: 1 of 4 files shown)


packages/viewer/src/components/viewer-transform-widget/__tests__/widget.spec.ts
```diff
@@ -143,6 +144,13 @@ function updateFrameCameraPosition(
   );
 }
 
+function createdPaddedFloat64Array(points: DrawablePoints): Float64Array {
```
MadisonEhlers-Vertex, line 147: This is duplicated in two tests - can we combine this somewhere?

MadisonEhlers-Vertex (approved):

PR #497, opened 2023-05-10
[PLAT-2919] Update hit indicator default colors and transform widget disabled state initialization

jdm717:
## Summary

Updates the `<vertex-viewer-transform-widget>` element to support having its disabled state updated prior to initialization of the widget by providing it when the widget is initialized. Additionally, this PR updates the CSS definitions for the `<vertex-viewer-hit-result-indicator>` element to not include single quotes, and updates the `parseCssColorValue` method to remove single quotes when retrieving a property value.

## Test Plan

- Verify that the transform widget does not log errors when the disabled states change prior to initialization
- Verify that the transform widget still properly updates disabled states after initialization
- Verify that the hit result indicator has a default color of `#0099cc`

## Release Notes

N/A

## Possible Regressions

Transform widget disabled states

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-hit-result-indicator/viewer-hit-result-indicator.tsx
+++ b/packages/viewer/src/components/viewer-hit-result-indicator/viewer-hit-result-indicator.tsx
@@ -79,8 +79,6 @@ export class ViewerHitResultIndicator {
         .getPropertyValue('--viewer-hit-result-indicator-plane-opacity')
         .trim();
 
-      console.log(this.arrowColor);
-
       this.indicator?.updateColors({
         arrow: this.arrowColor,
         plane: this.planeColor,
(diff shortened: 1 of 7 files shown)


packages/viewer/src/components/viewer-hit-result-indicator/viewer-hit-result-indicator.tsx
```diff
@@ -79,6 +79,8 @@ export class ViewerHitResultIndicator {
         .getPropertyValue('--viewer-hit-result-indicator-plane-opacity')
         .trim();
 
+      console.log(this.arrowColor);
```
MadisonEhlers-Vertex, line 82: Nit: is this needed or can we remove it?

MadisonEhlers-Vertex (approved):

packages/viewer/src/components/viewer-hit-result-indicator/viewer-hit-result-indicator.tsx
jdm717: Shoot, missed that being left in. Removed!

PR #498, opened 2023-05-11
[PLAT-2949] Opacity on Disabled Transform State

MadisonEhlers-Vertex:
## Summary

These changes are to update the disabled state of the transform widget. Instead of specifying a color, the disabled widget shows an opacity to maintain the color to denote which axis are disabled. These changes are being made to make the disabled state a better experience for users who have reported that the disabled state is not obvious. 

image

## Test Plan

Verify the opacity is shown on a disabled axis by making the opacity lighter for the transform widget.


jdm717 (approved):
LGTM!

PR #500, opened 2023-05-15
[PLAT-1181] Handle long press in the scene tree

amvertex:
## Summary
We want to handle long presses in the scene tree on mobile separately from single tap events. As a result, long pressing on a node in the scene tree will no longer select the item.

## Test Plan
Verify that long pressing on a node in the scene tree does not select the item.

## Release Notes
None

## Possible Regressions
Long presses in the scene tree

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
+++ b/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
@@ -159,6 +159,7 @@ export class SceneTreeTableCell {
 
   public disconnectedCallback(): void {
     this.hoverListener?.dispose();
+    this.clearLongPressTimer();
   }
 
   public componentWillRender(): void {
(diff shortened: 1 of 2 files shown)


packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -324,4 +339,18 @@ export class SceneTreeTableCell {
   ): void => {
     tree.toggleExpandItem(node);
   };
+
+  private clearLongPressTimer(): void {
```
MadisonEhlers-Vertex, line 343: Should we invoke the `clearLongPressTimer()` function on the disconnected callback?

MadisonEhlers-Vertex (approved):
Lgtm, left one suggestion

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
amvertex: Good idea! Added.

MadisonEhlers-Vertex (approved):
I think this makes sense. One question about this is if we should handle a longPress event emit from the cell? Just a thought. That way we can use the tap event emitted from here.

PR #502, opened 2023-05-17
[PLAT-3012] Ignore invalid summaries on incoming frames

jdm717:
## Summary

Adds support for ignoring "invalid" display list summaries on incoming frames, instead copying the previous frame's summary. This corrects a case on reconnects where the new FrameStream had not received a summary, and would send an empty set of visible/selected items rather than the current counts/bounds.

## Test Plan

- Verify that the `sceneViewSummary` property on the `viewer.frame` maintains its state across reconnects
- Verify that the `sceneViewSummary` property is updated after a reconnect if a selection/visibility change is made

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/339

### DIFF
--- a/packages/viewer/src/lib/types/__tests__/sceneViewSummary.spec.ts
+++ b/packages/viewer/src/lib/types/__tests__/sceneViewSummary.spec.ts
@@ -64,8 +64,5 @@ function copyWithValidSummary(frame: Frame, count: number): Frame {
 }
 
 function copyWithInvalidSummary(frame: Frame): Frame {
-  return copyWithSummary(frame, {
-    selectedVisibleSummary: { count: -1 },
-    visibleSummary: { count: -1 },
-  });
+  return copyWithSummary(frame, {});
 }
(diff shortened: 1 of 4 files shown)


amvertex (approved):

amvertex (approved):

packages/viewer/src/lib/types/__tests__/sceneViewSummary.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    visibleSummary: {
+      count,
+      boundingBox: BoundingBox.create(Vector3.create(), Vector3.right()),
+    },
+  });
+}
+
+function copyWithInvalidSummary(frame: Frame): Frame {
+  return copyWithSummary(frame, {
+    selectedVisibleSummary: { count: -1 },
```
MadisonEhlers-Vertex, line 68: Should this be `selectedVisibleSummary: null`?

jdm717: Not currently with the way this is set up, as the `visibleSummary` and `selectedVisibleSummary` will still always come back from FSS with a count of `-1` if it's invalid (as the `count` on the item set summary is required). We could consider making more things at the FSS level optional though to avoid this, or support undefined values in the SDK though they'd never truly be undefined

PR #504, opened 2023-05-30
[PLAT-2801] Wait for connected state when retrieving a Scene

jdm717:
## Summary

Updates the `scene()` method of the `<vertex-viewer>` to support requesting a scene when the stream is not in a `connected` state. With this change, calling the `scene()` method will wait up to 15 seconds before rejecting to allow for the call to succeed in the case that the method is called while in a `reconnecting` state.

## Test Plan

- Verify that interacting with the scene during a reconnect succeeds
  - An easy test case here is performing a `tap`, then using the `scene()` method to perform a ray-pick + select
- Verify that beginning and ending camera interactions continue to work as expected (updated to handle the async retrieval of the scene)

## Release Notes

N/A

## Possible Regressions

Camera interactions

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/interactions/interactionApi.ts
+++ b/packages/viewer/src/lib/interactions/interactionApi.ts
@@ -56,6 +56,7 @@ export interface ZoomData {
  */
 export abstract class InteractionApi {
   protected currentCamera?: Camera;
+  private sceneLoadingPromise?: Promise<Scene>;
   private lastAngle: Angle.Angle | undefined;
   private worldRotationPoint?: Vector3.Vector3;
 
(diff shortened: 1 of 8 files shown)


packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -204,7 +205,9 @@ export abstract class InteractionApi {
   public async beginInteraction(): Promise<void> {
     if (!this.isInteracting()) {
       this.interactionStartedEmitter.emit();
-      this.currentCamera = this.getScene().camera();
+      this.sceneLoadingPromise = this.getScene();
+      this.currentCamera = (await this.sceneLoadingPromise).camera();
```
amvertex, line 209: Could we await the line above instead? Like
```
this.sceneLoadingPromise = await this.getScene();
this.currentCamera = this.sceneLoadingPromise.camera();
```

amvertex (approved):

MadisonEhlers-Vertex (approved):

packages/viewer/src/lib/interactions/interactionApi.ts
jdm717: In this case we need the actual `Promise` so that the `endInteraction` method can `await` it in case the attempt comes in before the scene has been loaded, so unfortunately we need a setup like this. Open to other ideas on how to approach this though!

PR #505, opened 2023-05-31
[PLAT-2763] Only dispose scene-tree subscription once

amvertex:
## Summary
We were calling _dispose_ on the scene tree connection twice, which was throwing errors in scene-tree.  Therefore, this PR removes the duplicate call.

## Test Plan
Verify that the scene tree works as expected. Particularly, verify that scene view states can be applied to an instance of the scene tree.

## Release Notes
None

## Possible Regressions
Scene-tree connection

## Dependencies
None


MadisonEhlers-Vertex (approved):

MadisonEhlers-Vertex:
It's kind of amusing we had a unit test asserting the very problem! 😄

PR #506, opened 2023-06-15
[PLAT-2649] Support phantom state

amvertex:
## Summary
This PR adds support for phantom stream alterations, such that parts can be set as phantom. An example demonstrating how the feature can be used is also included.

## Test Plan
Verify that parts can be set as phantom using stream alterations.

## Release Notes
None

## Possible Regressions
Stream alterations

## Dependencies
None


examples/phantom-parts/index.html
```diff
@@ -0,0 +1,61 @@
...
+    <title>Phantom Parts</title>
+
+    <!-- Live reload script -->
+    <script src="http://localhost:35729/livereload.js?snipver=1"></script>
+
+    <script type="module" src="main.js"></script>
+
+    <link
+            rel="stylesheet"
+            href="https://unpkg.com/@vertexvis/viewer@0.17.x/dist/viewer/viewer.css"
```
amvertex, line 13: We might need to publish a new version in order for the example to work for external users.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -227,6 +227,11 @@ export class Viewer {
    */
   @Prop() public experimentalGhostingOpacity = 0;
 
+  /**
+   * Specifies the opacity, between 0 and 1, for a phantom feature, where 0 is completely hidden and 1 is completely visible.
+   */
+  @Prop() public phantomOpacity = 0.5;
```
MadisonEhlers-Vertex, line 233: I think the default here should be much lower. @mikemccarty-vertex Did you have any thoughts on what this value should be by default? Maybe 0.3, or even 0.1?

mikemccarty-vertex: Agreed.  It should be pretty _wispy_ by default.  The RCL default is 0.1 which is definitely wispy but (IMO) not outrageously so.

mikemccarty-vertex: This is `opacity=0.1`: 
image

packages/viewer/src/components.d.ts
```diff
@@ -2005,6 +2009,10 @@ declare namespace LocalJSX {
      * Emits an event when a provided oauth2 token is about to expire, or is about to expire, causing issues with establishing a websocket connection, or performing API calls.
      */
     onTokenExpired?: (event: VertexViewerCustomEvent<void>) => void;
+    /**
+     * Specifies the opacity, between 0 and 1, for a phantom feature, where 0 is completely hidden and 1 is completely visible.
+     */
+    phantomOpacity?: number;
```
MadisonEhlers-Vertex, line 2015: @jdm717 Curious on your thoughts here, if we should consider having this be a `phantomOptions` object, or if we should have the opacity right on here like this?

mikemccarty-vertex: I think I would prefer the `phantomOptions` route.

jdm717: Yeah, I think I'd agree with the `phantomOptions` route, since we'll likely add to those parameters down the road. That being said, the direction we're moving with these `options`-like properties does prevent a pure HTML based configuration of the viewer, which would be nice to maintain for simplicity. I'd vote for going with the `phantomOptions` for this work, then I'll add a FE Tech Talk topic around how we could support the HTML attribute approach in the future

MadisonEhlers-Vertex (approved):
LGTM!

PR #510, opened 2023-07-06
[PLAT-3103] Support end item operations

amvertex:
## Summary
This PR adds support for end item operations, such that parts can be set as end item. An example demonstrating how the feature can be used is also included.

## Test Plan
Verify that parts can be set as end item.

## Release Notes
None

## Possible Regressions
Part operations

## Dependencies
None

### DIFF
--- a/examples/end-items/index.html
+++ b/examples/end-items/index.html
@@ -1,0 +1,105 @@
+<!DOCTYPE html>
+<html dir="ltr" lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta
+            name="viewport"
+            content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0"
+    />
+    <title>End Items</title>
+
+    <!-- Live reload script -->
+    <script src="http://localhost:35729/livereload.js?snipver=1"></script>
+
+    <script type="module" src="main.js"></script>
+
+    <link rel="preconnect" href="https://fonts.gstatic.com" />
+    <link
+            href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;1,400;1,700&display=swap"
+            rel="stylesheet"
+    />
+
+    <link rel="stylesheet" href="/dist/viewer/viewer.css" />
+    <script type="module" src="/dist/viewer/viewer.esm.js"></script>
+    <script nomodule src="/dist/viewer.js"></script>
+
+    <style>
+      html,
+      body {
+        font-family: Roboto, Arial, Helvetica, sans-serif;
+        width: 100%;
+        height: 100%;
+        padding: 0;
+        margin: 0;
+      }
+
+      .app {
+        display: flex;
+        width: 100%;
+        height: 75%;
+      }
+
+      .tree {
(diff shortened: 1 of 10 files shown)


examples/end-items/main.js
```diff
@@ -0,0 +1,54 @@
...
+    const result = await raycaster.hitItems(position);
+
+    console.log(result);
+    if (result.hits && result.hits.length > 0) {
+      const hit = result.hits[0];
+
+      await scene
+        .items((op) => [
+          op.where((q) => q.all()).deselect(),
+          op.where((q) => q.withItemId(hit.itemId.hex)).select(),
```
jdm717, line 43: We may want to consider leveraging the `ancestors` field on these hit results rather than the `itemId` for the part hit, since it'll always be a leaf item. We may be able to demonstrate that relatively easily by doing something like `hit.ancestors.reverse()[0]` (basically just get the parent assembly), but that might still not always work depending on the assembly structure. What would you think about adding in a quick ancestor selection approach here? I.e. something like:

```typescript
let lastSelectedId;

viewer.addEventListener('tap', async (event) => {
  // setup

  const hit = result.hits[0];
  if (hit.ancestors.length > 0) {
    const ancestors = hit.ancestors.reverse();
    const toSelectId =
      ancestors[ancestors.findIndex((a) => a.itemId.hex === lastSelectedId) + 1]
        .itemId.hex;

    await scene
        .items((op) => [
          op.where((q) => q.all()).deselect(),
          op.where((q) => q.withItemId(toSelectId)).select(),
        ])
        .execute();

    lastSelectedId = toSelectId;
  }
})
```
Thinking that'd give this example a bit more flexibility to select the "correct" assembly to mark as an end item.

jdm717 (commented):
Looks great! Just had one thought on the example

examples/end-items/main.js
amvertex: Good idea! The ancestor selection logic will make setting the end items easier.

What do you think about performing the ancestor selection if the user is holding the shift key? I think there's also value in the user seeing that after the end item is set, a simple selection of any of the leaf items select the entire subassembly.

jdm717: Yeah, good call! I think that makes sense! Seems like the base selection + shift ancestor selection should demonstrate the capability fairly well

examples/end-items/index.html
```diff
@@ -0,0 +1,92 @@
...
+
+    <link rel="preconnect" href="https://fonts.gstatic.com" />
+    <link
+        href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;1,400;1,700&display=swap"
+        rel="stylesheet"
+    />
+
+    <link rel="stylesheet" href="/dist/viewer/viewer.css" />
+    <script type="module" src="/dist/viewer/viewer.esm.js"></script>
+    <script nomodule src="/dist/viewer.js"></script>
```
jdm717, line 24: Should these reference a published version (or a "to-be published" version)? Ran into some issues getting this example running locally, since the `dist` for the viewer isn't at the top level

jdm717 (approved):
LGTM!

PR #511, opened 2023-07-07
[PLAT-3129] Scene tree support for end items

jdm717:
## Summary

Adds support for end items in the scene-tree. For the time being, this support is relatively simple, just removing the expansion toggle similar to how we represent leaf nodes.

## Test Plan

- Verify that end items in a base scene are represented without an expansion toggle
- Verify that overriding items to mark them as end items properly removes any children displayed from expanding the node, and removes the expansion toggle
- Verify that overriding items to unmark them as end items properly re-adds any children displayed from expanding the node, and adds back the expansion toggle

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

https://github.com/Vertexvis/scene-tree-service/pull/172


packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -192,7 +193,7 @@ export class SceneTreeTableCell {
                 this.toggleExpansion(event);
               }}
             >
-              {!this.node?.isLeaf && (
+              {!this.node?.isLeaf && !this.node?.endItem && (
```
amvertex, line 196: 💯

PR #522, opened 2023-07-24
[PLAT-2057] Orthographic Touch Gestures

MadisonEhlers-Vertex:
## Summary

These changes are being introduced as a proposal to improve the orthographic touch gestures. Prior to this implementation, both pan & zoom were a little buggy. Zoom didn't work at all, and pan was very slow. These changes remove the delta scaling to be handled within the specific camera interactions, and use common points for both orthographic and perspective to perform operations on. 

## Test Plan

- Regression test two finger gestures in perspective
- Verify the two finger gestures in orthographic are now resolved and working as expected.

### DIFF
--- a/packages/viewer/src/lib/interactions/multiTouchInteractionHandler.ts
+++ b/packages/viewer/src/lib/interactions/multiTouchInteractionHandler.ts
@@ -25,25 +25,7 @@ export abstract class MultiTouchInteractionHandler
     point2: Point.Point
   ): void {
     if (this.currentPosition1 != null && this.currentPosition2 != null) {
-      // const delta = Point.scale(
-      //   Point.add(
-      //     Point.subtract(point1, this.currentPosition1),
-      //     Point.subtract(point2, this.currentPosition2)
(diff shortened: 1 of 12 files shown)


packages/viewer/src/lib/interactions/multiTouchInteractionHandler.ts
```diff
@@ -48,8 +44,13 @@ export abstract class MultiTouchInteractionHandler
           Matrix2.dot(previousToCurrent)
         )
       );
+      const center = Point.create(
+        (this.currentPosition1.x + this.currentPosition2.x) / 2,
+        (this.currentPosition1.y + this.currentPosition2.y) / 2
+      );
```
jdm717, line 50: Should this actually be looking at `point1` and `point2`? Looks like this may mean that the center point we compute is always one behind with this setup

jdm717 (commented):
LGTM! Just had one thought

jdm717 (approved):

PR #527, opened 2023-08-02
[PLAT-3218] Remove experimental ghosting

jdm717:
## Summary

Removes the `experimentalGhostingOpacity` attribute from the viewer, and sets the next release up for a minor version bump. Additionally, another scene-tree spinner case is fixed with this PR.

## Test Plan

- Verify the viewer no longer responds to the `experimentalGhostingOpacity`
- Verify the scene-tree spinner case where the `activeRowRange` would be set to `NaN` due to the `rowHeight` being zero no longer occurs
  - (this can be verified by logging the scene-tree element's `controller` and inspecting the ranges)

## Release Notes

N/A

## Possible Regressions

Scene tree row height computation

## Dependencies

N/A


packages/viewer/src/components/viewer/viewer.spec.tsx
```diff
@@ -300,11 +298,11 @@ describe('vertex-viewer', () => {
         })
       );
 
-      viewer.experimentalGhostingOpacity = 1;
+      viewer.phantom = { opacity: 1 };
       expect(update).toHaveBeenCalledWith(
         expect.objectContaining({
           streamAttributes: expect.objectContaining({
-            experimentalGhosting: 1,
+            phantom: expect.objectContaining({ opacity: 1 }),
```
amvertex, line 305: 💯

amvertex (approved):

PR #528, opened 2023-08-04
[PLAT-3218] Release v0.18.0

jdm717:
## Summary

Releases version 0.18.0, which includes the following changes:
- Fixes an issue where the `<vertex-viewer-transform-widget>` element would disconnect from a part if the handles were double-clicked (#526)
- Fixes a few issues where the `<vertex-scene-tree>` could display an infinite spinner when loading the tree for the first time (#526, #527)
- Removes experimental ghosting in favor of phantom states (#527)

## Test Plan

- Verify version 0.18.0 of all packages is published

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "private": true,
-  "nextVersionBump": "minor",
+  "nextVersionBump": "patch",
```
amvertex, line 4: 👍

amvertex (approved):

PR #529, opened 2023-08-09
[PLAT-1586] Support zooming past hit points

jdm717:
## Summary

Adds a `useMinimumPerspectiveZoomDistance` flag (defaulted to true) to allow a zoom interaction to zoom past the determined hit point. This allows for a continual zooming experience when the initial interaction was started with the cursor over geometry closer to the camera than the ultimate destination. When `useMinimumPerspectiveZoomDistance` is false, zoom will behave as it did previously, but with the flickering addressed when too close to the hit point.

This PR also attempts to fix one more scene-tree spinner 🙂 

## Test Plan

- Verify that zooming now allows you to pass through objects/hit planes
- Verify that setting `useMinimumPerspectiveZoomDistance` to false causes zoom to behave as it did previously

## Release Notes

N/A

## Possible Regressions

Zoom interactions

## Dependencies

N/A


packages/viewer/src/lib/interactions/interactionApiPerspective.ts
```diff
@@ -11,9 +11,19 @@ import {
...
 import { TapEventDetails } from './tapEventDetails';
 
+interface ZoomPositionData {
+  position: Vector3.Vector3;
+  distance: number;
+  isPastHitPlane: boolean;
+  keepCurrent: boolean;
+}
+
+const CAMERA_MIN_ZOOM_SCALAR = 0.2;
```
amvertex, line 25: 💯

amvertex (approved):
Looks great! This is a big improvement.

PR #532, opened 2023-08-22
[PLAT-3291] Update examples

amvertex:
## Summary
This PR updates all our SDK examples to use the latest version of the SDK.  Further, this updates the 'picking' example to use selection alterations.

## Test Plan
Verify all the examples work as expected.

## Release Notes
None

## Possible Regressions
Examples

## Dependencies
None

### DIFF
--- a/examples/align-two-surfaces/index.html
+++ b/examples/align-two-surfaces/index.html
@@ -10,15 +10,15 @@
 
     <link
       rel="stylesheet"
-      href="https://unpkg.com/@vertexvis/viewer@0.17.x/dist/viewer/viewer.css"
+      href="https://unpkg.com/@vertexvis/viewer@0.18.x/dist/viewer/viewer.css"
     />
     <script
       type="module"
-      src="https://unpkg.com/@vertexvis/viewer@0.17.x/dist/viewer/viewer.esm.js"
+      src="https://unpkg.com/@vertexvis/viewer@0.18.x/dist/viewer/viewer.esm.js"
     ></script>
     <script
       nomodule
-      src="https://unpkg.com/@vertexvis/viewer@0.17.x/dist/viewer.js"
+      src="https://unpkg.com/@vertexvis/viewer@0.18.x/dist/viewer.js"
     ></script>
 
     <style>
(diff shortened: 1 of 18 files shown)


examples/align-two-surfaces/index.html
```diff
@@ -39,7 +39,7 @@
 
       #viewer {
         width: 100%;
-        height: 100%;
+        height: calc(100% - 250px);
```
jdm717, line 42: minor nit: could we potentially add a container to size the viewer rather than having the fixed value removed? Bit of an edge case, but if I change my Chrome font size to very large, it cuts off part of the viewer. Thinking something like:

```html
<style>
  .container {
    display: flex;
    flex-direction: column;
    width: 100%;
    height: 100%;
  }
</style>

<body>
  <div class="container>
    <!-- example setup -->
  </div>
</body>
```

examples/picking/index.html
```diff
@@ -6,25 +6,52 @@
...
+        height: 100%;
+        padding: 0;
+        margin: 0;
+        overflow: hidden;
+      }
+
+      #viewer {
+        display: flex;
+        width: 100%;
+        height: calc(100% - 200px);
```
jdm717, line 38: nit: same as above

examples/picking/index.html
```diff
@@ -6,25 +6,52 @@
...
   <body>
-    <vertex-viewer id="viewer" class="viewer">
-      <vertex-viewer-toolbar data-viewer="viewer"></vertex-viewer-toolbar>
-    </vertex-viewer>
+  <h1 class="heading">Picking Instructions</h1>
+  <ul>
+    <li>Click a part to select it. Click the part again to select its parent.</li>
+    <li>Hold 'shift' while selecting a part to maintain the current selection while selecting the new part.</li>
+    <li>Press the spacebar or click in the empty space to deselect all parts.</li>
+  </ul>
```
jdm717, line 48: 💯 makes this much easier to follow when loaded up!

jdm717 (approved):
LGTM! Just had a couple small nits

examples/align-two-surfaces/index.html
amvertex: Good idea! Updated.

examples/picking/index.html
amvertex: Good idea! Updated.

PR #533, opened 2023-08-23
[PLAT-2348] Support changing frame background color for an active stream

jdm717:
## Summary

Adds support for changing the frame background color after a stream has already been established. Prior to this change, it was only possible to change the frame background color when starting a stream or reconnecting to a stream.

As part of this work, the way the `background` property is set on the `<vertex-viewer>` containers has been changed to sync after frames are drawn, rather than having the value immediately reflected. This prevents the state where the background color of the container has changed, but we haven't received the updated image with the matching background color.

https://github.com/Vertexvis/vertex-web-sdk/assets/5252480/f074f55d-10df-4335-84b2-6977b4b495e8

## Test Plan

- Verify that changing the `--viewer-background` and `--image-background` CSS properties updates the background color of the image
- Verify that by default the background color is still `#ffffff`
- Verify that removing the `--viewer-background` and `--image-background` CSS properties (or setting them to invalid values) changes the background color back to `#ffffff`

## Release Notes

N/A

## Possible Regressions

Viewer background colors

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/361

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.css
+++ b/packages/viewer/src/components/viewer/viewer.css
@@ -27,7 +27,6 @@
   width: 100%;
   height: 100%;
   position: relative;
-  background: #ffffff;
 }
 .enable-pointer-events {
   touch-action: none;
@@ -37,7 +36,6 @@
   width: 100%;
   height: 100%;
   overflow: hidden;
-  background: #ffffff;
 }
 
 .error-message {
--- a/packages/viewer/src/components/viewer/viewer.css
+++ b/packages/viewer/src/components/viewer/viewer.css
@@ -27,7 +27,7 @@
   width: 100%;
   height: 100%;
   position: relative;
-  background: var(--image-background, var(--viewer-background, #ffffff));
+  background: #ffffff;
 }
 .enable-pointer-events {
   touch-action: none;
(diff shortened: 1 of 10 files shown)


packages/viewer/src/components/viewer/viewer.css
```diff
@@ -27,7 +27,7 @@
   width: 100%;
   height: 100%;
   position: relative;
-  background: var(--image-background, var(--viewer-background, #ffffff));
+  background: #ffffff;
```
amvertex, line 30: Is this still used as the default? Or can it be removed altogether?

amvertex (approved):

packages/viewer/src/components/viewer/viewer.css
jdm717: Let me take a look! This was required with an initial approach, but after converting the fallback to just `#ffffff`, this may be possible to remove

jdm717: Looks like it's no longer required - removed with [4765024](https://github.com/Vertexvis/vertex-web-sdk/pull/533/commits/4765024acdbb22d03c3221fcfc04639d474e1ac6). Good catch!

amvertex (approved):

PR #534, opened 2023-08-28
[PLAT-2432] Add walk mode support

jdm717:
## Summary

Adds support for walk mode by exposing a couple new tools, specifically the `<vertex-viewer-walk-mode-tool>` and `<vertex-viewer-teleport-tool>` with the thought that these can be dropped in underneath a `<vertex-viewer>` and toggled on/off.

## Test Plan

- Verify the `<vertex-viewer-walk-mode-tool>`'s behavior
  - Should be able to walk through the model with `enabled=true` (default)
  - Should be able to provide a custom `<vertex-viewer-teleport-tool>` to set animation configurations
  - Should be able to enable/disable
  - Should be able to change the `teleportTool` and have it update the underlying `<vertex-viewer-teleport-tool>`
- Verify the `<vertex-viewer-teleport-tool>`'s behavior
  - Should be able to change the `tool` and see the behavior change
  - Should be able to provide custom animation configuration

## Release Notes

N/A

## Possible Regressions

Interaction APIs - small change to update `transformCamera` in the perspective/orthographic APIs + added support for a render configuration

## Dependencies

N/A


packages/geometry/src/boundingBox.ts
```diff
@@ -84,3 +84,15 @@ export function union(box: BoundingBox, ...rest: BoundingBox[]): BoundingBox {
...
 }
 /* eslint-enable padding-line-between-statements */
+
+/**
+ * Returns the distance between the min and max for the provided
+ * bounding box for each axis.
+ */
+export const lengths = (box: BoundingBox): Vector3.Vector3 => {
+  return Vector3.create(
+    Math.abs(box.min.x) + Math.abs(box.max.x),
```
amvertex, line 94: Should this be `box.max.x - box.min.x` instead? This would return the wrong value if the min and max are both positive or both negative.

packages/viewer/src/components/viewer-teleport-tool/viewer-teleport-tool.tsx
```diff
@@ -0,0 +1,172 @@
...
+   */
+  @Prop({ mutable: true })
+  public mode?: ViewerTeleportMode;
+
+  /**
+   * Indicates whether animations will be used when performing camera
+   * operations. Defaults to `true`.
+   */
+  @Prop()
+  public animationsDisabled = false;
```
amvertex, line 57: The description says this defaults to true, but we assign it to false.

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -713,7 +713,7 @@ export class Viewer {
     this.initializeDefaultKeyboardInteractionHandlers();
   }
 
-  /**
+  /**re
```
amvertex, line 716: Is the 're' needed?

packages/viewer/src/lib/walk-mode/controller.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.model.setTeleportMode(mode);
+  }
+
+  /**
+   * Updates the configuration for downstream walk mode interaction handlers.
+   *
+   * `teleportHeightScalar` - scalar used for fine-tuning the distance to offset
+   * the camera from a surface when performing a `teleport-and-align`. A larger
+   * number here will result in a shorter distance from the surface and vice-versa.
+   * Defaults to `8.5`.
```
amvertex, line 49: It's accurate with the current math, but I'm not sure that it's intuitive that a larger _teleportHeightScalar_ will place the user closer to the ground.

packages/viewer/src/lib/walk-mode/controller.ts
```diff
@@ -0,0 +1,161 @@
...
+    ...keys: string[]
+  ): void {
+    this.model.addKeyBinding(operation, new KeyBinding(...keys));
+  }
+
+  /**
+   * Adds a custom keybinding for a specific walk operation. This will replace any
+   * existing keybindings to allow for overriding the default behavior.
+   *
+   * @example
```
amvertex, line 90: I like the quick example here!

packages/viewer/src/lib/walk-mode/interactions.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.api?.walk(Vector3.scale(this.relativeWalkSpeed(), Vector3.down()));
+  }
+
+  private moveDown(): void {
+    this.api?.walk(Vector3.scale(this.relativeWalkSpeed(), Vector3.up()));
+  }
+
+  private relativeWalkSpeed(): number {
+    const speedScalar = (5 / this.model.getKeyboardWalkSpeed()) * 100;
+    return 1 / speedScalar;
```
amvertex, line 206: Is there an advantage to writing it this way as opposed to `return this.model.getKeyboardWalkSpeed()/500`?

packages/geometry/src/boundingBox.ts
jdm717: Yeah, good point. This was copied over from some other logic, and I'm guessing there's potential for a bug there and we just haven't encountered those edge cases. I'll update this

packages/viewer/src/components/viewer/viewer.tsx
jdm717: Nope, guessing I meant to type this into the terminal and accidentally saved. Good catch! I'll remove

packages/viewer/src/lib/walk-mode/controller.ts
jdm717: Yeah, that's fair. My thought with changing the way this value worked was primarily to have a whole number instead of a decimal, but I'll rethink this a bit to see if we can get the opposite behavior where a higher number results in a larger distance from the ground

packages/viewer/src/lib/walk-mode/interactions.ts
jdm717: Yeah, this is probably overly confusing. This was pulled directly from Connect, and I'm guessing there used to be some additional logic here. I'll simplify this a bit!

packages/viewer/src/lib/walk-mode/controller.ts
amvertex: That's fair, I agree that having a whole number instead of a decimal would be nice. If it's too much work, we can leave it as it is.

packages/viewer/src/lib/walk-mode/controller.ts
```diff
@@ -43,10 +43,11 @@ export class WalkModeController {
...
    *
-   * `teleportHeightScalar` - scalar used for fine-tuning the distance to offset
-   * the camera from a surface when performing a `teleport-and-align`. A larger
-   * number here will result in a shorter distance from the surface and vice-versa.
-   * Defaults to `8.5`.
+   * `teleportHeightPercentage` - percentage used for fine-tuning the distance to offset
+   * the camera from a surface when performing a `teleport-and-align`. This percentage is
+   * used alongside the shortest side of the visible bounding box to determine how far to
+   * place the camera from the surface that has been hit, with a larger percentage placing
+   * the camera further from the surface and vice-versa. Defaults to 11.75%.
```
amvertex, line 50: I like this a lot better!

packages/viewer/src/lib/walk-mode/model.ts
```diff
@@ -56,7 +56,7 @@ export class WalkModeModel {
   };
 
   private configuration: ViewerWalkModeConfiguration = {
-    teleportHeightScalar: 8.5,
+    teleportHeightPercentage: 8.5,
```
amvertex, line 59: Do you want to update this to 11.75?

amvertex (commented):
Nice job! This is looking great.

packages/viewer/src/lib/walk-mode/model.ts
jdm717: Ah, yes. Unsaved change when I pushed this up 🙃

amvertex (approved):

amvertex (approved):

amvertex (approved):

PR #545, opened 2023-09-12
[PLAT-3197] Use flyTo for fitToBoundingBox and remove client logic

jdm717:
## Summary

Moves to using `flyTo` for the `fitToBoundingBox` and `viewAll` methods of the `Camera`. This allows us to remove the client-side logic for fitting to a specific bounding box, and resolves an issue with performing a `fitToBoundingBox` in orthographic mode when on mobile.

Additionally, this PR converts our `wheel` listener for zoom to be an explicitly non-passive event listener, since it calls `event.preventDefault()`, and converts our `touchstart` handler for `tap` events to a passive event listener since it does not prevent default behavior. This resolves a couple warnings that were being displayed each time the viewer initialized.

## Test Plan

- Verify that the view cube and `viewAll` work as expected in orthographic mode
- Verify that the view cube and `viewAll` work as expected in perspective mode
- Verify that the view cube and `viewAll` in orthographic mode properly fit to the screen when on a mobile device (or aspect ratio)

## Release Notes

N/A

## Possible Regressions

Fit to bounding box behavior, zoom, tap interactions

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/364


MadisonEhlers-Vertex (commented):
lgtm!

MadisonEhlers-Vertex (approved):

PR #546, opened 2023-09-13
[PLAT-3187] Handle view cube interactions with no geometry

amvertex:
## Summary
Previously, we always performed a 'view all' when selecting a standard view on the view cube, which caused exceptions when no geometry was visible.  Therefore, this PR adjusts the standard view interactions to only perform a 'view all' when geometry is visible.  If there's no visible geometry, the default camera position for the view will be used instead.

## Test Plan
Verify view cube interactions work as expected, particularly when there's no visible geometry

## Release Notes
None

## Possible Regressions
Interacting with the view cube

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/viewer-view-cube/viewer-view-cube.spec.tsx
+++ b/packages/viewer/src/components/viewer-view-cube/viewer-view-cube.spec.tsx
@@ -4,7 +4,7 @@ jest.mock('../../lib/rendering/imageLoaders');
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { h } from '@stencil/core';
 import { newSpecPage } from '@stencil/core/testing';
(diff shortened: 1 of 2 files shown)


MadisonEhlers-Vertex (approved):
One observation, is that when interacting with 0 geometry & the view cube, after I turn on geometry again, the view is in the origin, rather than my "full" view. But I think this is a good solution over what we're doing today. :) LGTM

amvertex:
> One observation, is that when interacting with 0 geometry & the view cube, after I turn on geometry again, the view is in the origin, rather than my "full" view. But I think this is a good solution over what we're doing today. :) LGTM

Yeah, I saw that too when I was testing, but I couldn't determine what a universally better camera placement would be, because we don't know what parts a user will turn on after applying the standard view. I agree that this is better than what we were doing!

PR #547, opened 2023-09-25
[PLAT-3386] Always request images with at least 1x1 dimensions

jdm717:
## Summary

Updates the `ViewerStream` to always request images with at least 1x1 dimensions rather than requesting a zero-area image and displaying an error.

## Test Plan

- Verify that errors are not logged when a frame is requested with zero-area dimensions
  - For starting a stream and reconnecting, adding `display: none` to the viewer when initialized previously logged this error
  - For updating stream dimensions, adding `display: none` to the viewer after it has loaded initially previously logged this

## Release Notes

N/A

## Possible Regressions

Stream dimensions

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/stream/stream.ts
+++ b/packages/viewer/src/lib/stream/stream.ts
@@ -164,7 +164,7 @@ export class ViewerStream extends StreamApi {
     if (fields.dimensions != null && fields.dimensions !== this.dimensions) {
       this.dimensions = fields.dimensions;
       this.ifState('connected', () =>
-        this.updateDimensions({ dimensions: this.dimensions })
+        this.updateDimensions({ dimensions: this.getDimensions() })
       );
     }
 
--- a/packages/viewer/src/lib/stream/stream.ts
+++ b/packages/viewer/src/lib/stream/stream.ts
@@ -387,7 +387,7 @@ export class ViewerStream extends StreamApi {
     const res = fromPbStartStreamResponseOrThrow(
       await this.startStream({
         streamKey: { value: resource.resource.id },
-        dimensions: this.dimensions,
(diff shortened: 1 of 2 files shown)


packages/viewer/src/lib/stream/stream.ts
```diff
@@ -581,6 +581,15 @@ export class ViewerStream extends StreamApi {
     }
   }
 
+  private getDimensions(): Dimensions.Dimensions {
+    if (Dimensions.area(this.dimensions) === 0) {
+      // Ensure we always request at least a 1-pixel frame, even if the dimensions
+      // haven't been set higher than zero.
+      return Dimensions.create(1, 1);
```
amvertex, line 588: 💯

amvertex (approved):
Looks great! It will be nice to not hit this when testing locally!

packages/viewer/src/lib/stream/stream.ts
```diff
@@ -581,6 +581,15 @@ export class ViewerStream extends StreamApi {
     }
   }
 
+  private getDimensions(): Dimensions.Dimensions {
+    if (Dimensions.area(this.dimensions) === 0) {
```
mikemccarty-vertex, line 585: `... <= 0`?

jdm717: Personally, I think I'd rather let FSS throw an exception in the case of `<= 0`, but open to other thoughts there. This `=== 0` case is a valid state in the UI (`display: none` or auto-sized to  a width/height of 0), where I don't think there's a valid case for negative dimensions. Whether that exception is helpful today though is another question, so I could add a case to throw a custom error here

PR #552, opened 2023-10-13
[PLAT-3441] Update default scene-tree-search behavior

jdm717:
## Summary

Updates the default behavior of the `<vertex-scene-tree-search>` element to require either an `Enter` press or a `blur` event to trigger the search by making the `debounce` field optional. If a `debounce` is specified, the search behavior will be the same as it was prior to this change.

## Test Plan

- Verify default behavior of the search component
  - Should search when `Enter` is pressed
  - Should search when a `blur` event occurs (this also requires the value to have changed)
  - Should always search when the input is cleared (either by button or by deleting all text)
- Verify debounce behavior of the search component continues to work as expected

## Release Notes

N/A

## Possible Regressions

scene-tree-search behavior

## Dependencies

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "private": true,
-  "nextVersionBump": "patch",
+  "nextVersionBump": "minor",
   "devDependencies": {
     "eslint": "^8.17.0",
     "http-server": "^14.1.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "private": true,
-  "nextVersionBump": "patch",
+  "nextVersionBump": "minor",
```
jdm717, line 4: Making this change a `minor` bump here since it changes the default behavior of the search element substantially. Open to reconsidering this though, since technically there are no breaking changes that require code updates.

PR #554, opened 2023-10-17
[PLAT-3557] Add support for animation when loading SceneViewStates

jdm717:
## Summary

Adds support for similar animation configuration as the `scene.camera().render()` method to the `scene.applySceneViewState()` and `scene.applyPartialSceneViewState()` methods. If this animation duration is configured, these methods will now perform a `flyTo` (and wait for the animation to complete) using the provided ID prior to applying the SceneViewState. The wait for the animation to complete is also configurable using the `waitForAnimation` flag, and if no animation duration is provided, it will maintain the same behavior as today. Additionally, the `scene.applyPartialSceneViewState()` method will only respect the provided animation if the `'camera'` feature is provided in the list of items to apply.

## Test Plan

- Verify that the `scene.applySceneViewState()` without an animation configuration continues to work as expected
- Verify that the `scene.applyPartialSceneViewState()` without an animation configuration continues to work as expected
- Verify that providing an animation configuration to `scene.applySceneViewState()` works as expected
  - By default, should wait for the animation before applying other features
  - Providing `waitForAnimation: false`, the other features should apply during the animation
- Verify that providing an animation configuration to `scene.applyPartialSceneViewState()` works as expected
  - By default, should wait for the animation before applying other features
  - Providing `waitForAnimation: false`, the other features should apply during the animation
  - Not providing the `'camera'` feature should not change the camera

## Release Notes

N/A

## Possible Regressions

Loading of SceneViewStates

## Dependencies

https://github.com/Vertexvis/frame-streaming-service/pull/367

### DIFF
--- a/packages/stream-api/package.json
+++ b/packages/stream-api/package.json
@@ -36,7 +36,7 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
-    "@vertexvis/frame-streaming-protos": "^0.11.1"
+    "@vertexvis/frame-streaming-protos": "^0.11.2"
   },
   "devDependencies": {
     "@types/jest": "^27.5.1",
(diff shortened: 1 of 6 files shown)


amvertex (approved):
Looks good!

PR #563, opened 2024-02-06
[PLAT-3771] Support for markup line end styles

amvertex:
## Summary
This PR adds support for designating a line end style for a markup arrow, specifically either a closed-triangle arrow, an open-line arrow, a dot, a hash mark, or an open line.

## Test Plan
Verify arrow markup can be created with specific line end styles

## Release Notes
None

## Possible Regressions
Arrow markup

## Dependencies
None


packages/geometry/src/point.ts
```diff
@@ -89,6 +89,69 @@ export function scale(pt: Point, scaleX: number, scaleY: number): Point {
   };
 }
 
+/**
+ * Returns a new `Point` where `x` and `y` are multiplied by the given scale
+ * factor.
+ */
+export function scaleProportional(pt: Point, scale: number): Point {
+  return {
```
MadisonEhlers-Vertex, line 97: nit: this function could call the other scale:

return scale(pt, scale, scale);

packages/viewer/src/components.d.ts
```diff
@@ -1071,6 +1088,10 @@ export namespace Components {
      * This property will automatically be set when a child of a `<vertex-viewer-markup>` element.
      */
     disabled: boolean;
+    /**
+     * The line end style of the ending anchor. This default to 'arrow-triangle.'
+     */
+    endLineEndStyle: LineEndStyle;
```
MadisonEhlers-Vertex, line 1094: nit: are there any other naming conventions we could explore besides `EndStyle`? I don't love the word `end` used twice here.

A few ideas:
- `endLineStyle`
- `endLineTipStyle`
- `endLineAnchorType`

MadisonEhlers-Vertex (approved):

PR #564, opened 2024-02-07
[PLAT-3728] Fix zoom + twist touch interaction

jdm717:
## Summary

Fixes an interaction where performing a slight zoom alongside a twist interaction would result in the twist not being applied. This was a result of the check we were previously doing on the angle (`Math.abs(angle) > 0.5`) to prevent a wobbling behavior when panning. The underlying cause of this behavior appears to be related to the `PointerEvent`s firing for each individual touch point (or `TouchEvent`s firing when only one touch point has been updated), which calculates a very small angle to twist, followed by another event which calculates the same very small (or similar) angle in the opposite direction.

This PR adds logic to perform the interactions when an animation frame fires, grouping events until that point, such that they become partially batched. This results in multiple touch point changes being processed at the same time, allowing those very small opposing angles to cancel out and not result in a twist interaction. This still doesn't completely remove the potential for the wobbling behavior, as it's easy when performing these interactions on a mobile device to have very subtle changes in the angle between two touch points, but this corrects the ability to perform small twist gestures or zoom + twist gestures, while still reducing the wobbling effect.

## Test Plan

- Perform two-touch interactions with a model on a mobile device
  - Verify that small zooms + twists properly twists the model
  - Verify that panning does not result in significant wobbling

## Release Notes

N/A

## Possible Regressions

Two point touch interactions

## Dependencies

N/A


amvertex (approved):
Looks great! Nice job figuring all this out!

PR #565, opened 2024-02-08
[PLAT-3942] Exposing the volume selection model types and emitting the controller when changed

MadisonEhlers-Vertex:
## Summary

These changes enable developers to use the volume selection tool to get the controller when the controller is initialized or changes.


jdm717 (approved):
LGTM!

PR #568, opened 2024-02-23
[PLAT-3952] Clear material overrides

amvertex:
## Summary
This PR updates the behavior when clearing a material override such that the part reverts to its default material instead of inheriting any material overrides applied to an ancestor of the item.
 
## Test Plan
Verify material overrides work as expected. Specifically, verify the following use case:
1. Apply a material override to a parent item.
2. Clear the material override for one of the children.
3. Verify that the child now appears with its default material and the appearance of the remainder of the parts is unchanged.

## Release Notes
Ability to set an override default material for a scene item

## Possible Regressions
Material overrides

## Dependencies
None

### DIFF
--- a/packages/stream-api/package.json
+++ b/packages/stream-api/package.json
@@ -36,7 +36,7 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
-    "@vertexvis/frame-streaming-protos": "^0.13.0"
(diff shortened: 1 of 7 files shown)


jdm717 (approved):
LGTM!

jdm717 (approved):

PR #571, opened 2024-02-27
[PLAT-3965] [PLAT-3960] Handle updating markup

amvertex:
## Summary
We now dispatch an event after a piece of markup has changed containing the updated markup element.  Specifically, the event is dispatched after markup is moved, resized, or the line anchor styles are changed for arrow markup.

## Test Plan
Verify the 'markupChanged' event is dispatched after existing markup is changed.

## Release Notes
Emits an event when markup is changed

## Possible Regressions
Editing markup

## Dependencies
None


packages/viewer/src/components/viewer-markup-arrow/interactions.ts
```diff
@@ -18,7 +18,8 @@ export class ArrowMarkupInteractionHandler extends MarkupInteractionHandler {
   public constructor(
     private readonly markupEl: HTMLVertexViewerMarkupArrowElement,
     private readonly editBegin: EventEmitter<void>,
-    private readonly editEnd: EventEmitter<void>
+    private readonly editEnd: EventEmitter<void>,
+    private readonly markupUpdated: EventEmitter<HTMLVertexViewerMarkupArrowElement>
```
jdm717, line 22: Thoughts on leveraging the `editEnd` for this event in the arrow, circle, and freeform elements? Looks like that's the only place we're dispatching the new event, so we might just be able to change the `EventEmitter<void>` to be a `EventEmitter<HTMLVertexViewerMarkupArrowElement>` (or other element types) and avoid the new event on the individual pieces of markup

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -351,16 +362,36 @@ export class ViewerMarkup {
...
+
+    // If a piece of arrow markup is currently selected, then update its anchor style
+    const markup = await this.getMarkupElements();
+    markup.forEach((m) => {
+      if (m.id === this.selectedMarkupId) {
+        (m as HTMLVertexViewerMarkupArrowElement).endLineAnchorStyle =
+          this.endLineAnchorStyle;
+        this.markupChanged.emit(m);
+      }
+    });
```
jdm717, line 394: nit: We may be able to simplify the updates here, and avoid casting the element to an arrow element so we don't emit events/set properties on other types of markup. Thoughts on adding something like this?
```
private async getSelectedArrowMarkup(): Promise<HTMLVertexViewerMarkupArrowElement | undefined> {
  if (this.selectedMarkupId != null) {
    const selected = await this.getMarkupElement(this.selectedMarkupId);
    
    return isVertexViewerArrowMarkup(selected) ? selected : undefined;
  }
}

...

protected async handleEndLineAnchorStyleChanged(): Promise<void> {
  this.updatePropsOnMarkupTool();

  const selectedArrow = await this.getSelectedArrowMarkup();

  if (selectedArrow != null) {
    selectedArrow.endLineAnchorStyle = this.endLineAnchorStyle;
    this.markupChanged.emit(selectedArrow);
  }
}
```
with `handleStartLineAnchorStyleChanged` doing the same but setting the `startLineAnchorStyle`. Might be some further simplification, but mainly thinking to avoid re-writing the `forEach` logic and emitting events for other types of markup

amvertex: Good idea, updated!

packages/viewer/src/components/viewer-markup-arrow/interactions.ts
amvertex: This was poorly named, since editEnd was dispatched when both creating and editing markup. I changed the name of the event to _interactionEnd_ and it now returns both the current piece of markup as well as a boolean describing if the markup was just created.

I think this makes it much easier to read now!

packages/viewer/src/components/viewer-markup-arrow/interactions.ts
```diff
@@ -17,8 +18,8 @@ export class ArrowMarkupInteractionHandler extends MarkupInteractionHandler {
 
   public constructor(
     private readonly markupEl: HTMLVertexViewerMarkupArrowElement,
-    private readonly editBegin: EventEmitter<void>,
-    private readonly editEnd: EventEmitter<void>
+    private readonly interactionBegin: EventEmitter<void>,
+    private readonly interactionEnd: EventEmitter<MarkupInteraction>
```
jdm717, line 22: 💯 nice to have these events reflect the use a bit better!

jdm717 (approved):
LGTM! Only other thought is that it does look like we'll want to update the `nextVersionBump` [here](https://github.com/Vertexvis/vertex-web-sdk/blob/master/package.json#L4) to be `minor` with this PR since this is a breaking change

PR #574, opened 2024-03-01
[PLAT-3827] Support for visibility query

amvertex:
## Summary
This PR adds support for visibility scene alteration queries, so operations can be performed on all visible or all hidden parts in a scene. 

## Test Plan
Verify scene alteration operations can be performed using visibility queries

## Release Notes
Visibility queries

## Possible Regressions
Scene alteration queries

## Dependencies
None


packages/viewer/src/lib/scenes/queries.ts
```diff
@@ -46,6 +46,14 @@ interface AllSelectedQueryExpression {
   type: 'all-selected';
 }
 
+interface AllVisibleQueryExpression {
+  type: 'all-visible';
+}
+
+interface AllHiddenQueryExpression {
```
MadisonEhlers-Vertex, line 53: Do we have a known use-case for the hidden query?

I may recommend us not adding that until it's needed.

PR #575, opened 2024-03-04
[PLAT-1372] Add Vector3 Euler between method

jdm717:
## Summary

Adds an `eulerTo` utility for `Vector3`s to support determining the rotation to apply to the `<vertex-viewer-transform-widget>` based on a hit result normal and the global up vector.

## Test Plan

- Verify that the behavior in Connect continues to work as expected (after it has been updated with this version)
- Unit tests

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/geometry/src/vector3.ts
+++ b/packages/geometry/src/vector3.ts
@@ -1,5 +1,7 @@
+import * as Euler from './euler';
 import { lerp as lerpNumber } from './math';
 import * as Matrix4 from './matrix4';
+import * as Quaternion from './quaternion';
 
 /**
  * A `Vector3` represents a vector of 3 dimensions values. It may represent a
@@ -268,6 +270,21 @@ export function angleTo(a: Vector3, b: Vector3): number {
   const theta = dot(a, b) / (magnitude(a) * magnitude(b));
   // Clamp to avoid numerical problems.
   return Math.acos(theta);
+}
+
+/**
+ * Returns the Euler angle, in radians with the `xyz` ordering, between two vectors.
+ */
+export function eulerTo(a: Vector3, b: Vector3): Euler.Euler {
+  const normalizedA = normalize(a);
+  const normalizedB = normalize(b);
+
+  const angle = Math.acos(dot(normalizedA, normalizedB));
+  const axis = normalize(cross(normalizedA, normalizedB));
+
+  return Euler.fromRotationMatrix(
+    Matrix4.makeRotation(Quaternion.fromAxisAngle(axis, angle))
+  );
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/geometry/src/vector3.ts
```diff
@@ -270,6 +272,30 @@ export function angleTo(a: Vector3, b: Vector3): number {
...
+ * Returns the Euler angle, in radians with the `xyz` ordering, between two vectors.
+ */
+export function eulerTo(a: Vector3, b: Vector3): Euler.Euler {
+  const normalizedA = normalize(a);
+  const normalizedB = normalize(b);
+
+  const angle = Math.acos(dot(normalizedA, normalizedB));
+  const axis = normalize(cross(normalizedA, normalizedB));
+
+  const vectorsAreParallel = angle === Math.PI || angle === 0;
```
mikemccarty-vertex, line 285: Testing FP values for exact equality is an anti-pattern.

mikemccarty-vertex: I would write it  like this:

```
const dotDelta = Math.cos(some_very_small_angle); // e.g. 1 degree
const dotAB = dot(normalizedA, normalizedB);
const vectorsAreParallel = Math.abs(dotAB) > dotDelta;
```
... and don't try computing the cross-product until after you've confirmed the vectors aren't parallel.  The `crossAxisIsDefined` logic can be removed entirely.

In general, try to avoid the "arc cos" type functions as much as you can -- they aren't as precise as the other trig functions.

packages/geometry/src/vector3.ts
```diff
@@ -270,6 +273,31 @@ export function angleTo(a: Vector3, b: Vector3): number {
...
+  if (vectorsAreParallel) {
+    return dotAB > 1 - 1e-6 ? Euler.create() : Euler.create({ x: Math.PI });
+  }
+
+  const normalizedQ = Quaternion.normalize(
+    Quaternion.create({
+      w: 1 + dotAB,
+      ...cross(normalizedA, normalizedB),
+    })
+  );
```
mikemccarty-vertex, line 296: Consider adding this as a comment: https://www.xarg.org/proof/quaternion-from-two-vectors/

mikemccarty-vertex (approved):

packages/geometry/src/vector3.ts
jdm717: Good thought, added a comment with https://github.com/Vertexvis/vertex-web-sdk/pull/575/commits/dc032586238ce031398eefe205086b8c353c2f7e

PR #576, opened 2024-03-05
[PLAT-3897] Scene-tree filter on visible items

amvertex:
## Summary
This PR adds support for filtering only on the visible items in the scene tree.

## Test Plan
Verify that a scene tree can be filtered on only the visible items and it returns the correct filtered tree and number of results.

## Release Notes
Ability to filter on visible items in the scene tree

## Possible Regressions
Scene tree filtering

## Dependencies
None


packages/viewer/src/components.d.ts
```diff
@@ -205,6 +205,10 @@ export namespace Components {
      * A list of the metadata keys that a scene tree search should be performed on.
      */
     metadataSearchKeys: MetadataKey[];
+    /**
+     * Indicates whether the metadata search should remove hidden items from the results.
+     */
+    metadataSearchRemoveHiddenItems: boolean;
```
MadisonEhlers-Vertex, line 211: Is this feature only supported in the context of metadata search? Even in the context of a regular tree text search, shouldn't this be supported? My suggestion would be to name this something like 

💡  `filterHiddenItems` or something like thats, so it's clear to the developer that it's a setting on filtering.

or a filter object that contains something like:


```
FilterOptions: {
  text: string,
  removeHiddenItems: boolean
  } 
  ```

MadisonEhlers-Vertex (approved):

PR #583, opened 2024-03-20
[PLAT-3970] Single transform undo

jdm717:
## Summary

Adds support for a single undo step using Command+Z or Control+Z (if the `EXPERIMENTAL_undoKeybindings` flag is set) to the `<vertex-viewer-transform-widget>`. This undo step will undo the most recent completed transform performed by the widget.

Video of added behavior:

https://github.com/Vertexvis/vertex-web-sdk/assets/5252480/4079c233-b648-4072-b8c1-468902fc443d

In addition to the undo step, this PR also fixes an issue with computation of the rotation applied from the input when rotation has already been performed on all three axes.

## Test Plan

- Verify the transform input continues to work as expected
- Verify setting the `EXPERIMENTAL_undoKeybindings` allows for a single undo using Command+Z or Control+Z

## Release Notes

N/A

## Possible Regressions

Transform widget inputs

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/viewer-transform-widget/dom.ts
+++ b/packages/viewer/src/components/viewer-transform-widget/dom.ts
@@ -1,3 +1,6 @@
 export function focusInputElement(element: HTMLInputElement): void {
-  element.focus();
+  if (window.document.activeElement !== element) {
+    element.focus();
+    element.setSelectionRange(element.value.length, element.value.length);
+  }
 }
--- a/packages/viewer/src/components/viewer-transform-widget/util.ts
+++ b/packages/viewer/src/components/viewer-transform-widget/util.ts
@@ -83,8 +83,27 @@ export function convertCanvasPointToWorld(
   return undefined;
 }
 
-export function computeInputTransform(
+export function computeInputDeltaTransform(
   current: Matrix4.Matrix4,
+  identifier: string,
+  value: number,
+  lastValue: number,
+  distanceUnit: DistanceUnitType,
+  angleUnit: AngleUnitType
+): Matrix4.Matrix4 {
+  return appliedToCurrent(
+    current,
+    computeInputGlobalTransform(
+      identifier,
+      value,
+      lastValue,
+      distanceUnit,
+      angleUnit
+    )
+  );
+}
+
+function computeInputGlobalTransform(
   identifier: string,
   value: number,
   lastValue: number,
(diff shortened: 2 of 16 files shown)


packages/viewer/src/components/viewer-transform-widget/dom.ts
```diff
@@ -1,3 +1,6 @@
 export function focusInputElement(element: HTMLInputElement): void {
-  element.focus();
+  if (window.document.activeElement !== element) {
+    element.focus();
+    element.setSelectionRange(element.value.length, element.value.length);
+  }
```
jdm717, line 5: This change causes the input to only focus if it's not currently focused (fixes the cursor moving around when pressing the up arrow), as well as fixes an issue where the cursor would get placed at the start of the text

packages/viewer/src/components/viewer-transform-widget/util.ts
```diff
@@ -99,22 +100,40 @@ export function computeInputTransform(
...
+        )
       );
     case 'z-rotate':
-      return Matrix4.makeRotation(
-        Quaternion.fromAxisAngle(Vector3.forward(), rotation())
+      return appliedToCurrent(
+        current,
+        Matrix4.makeRotation(
+          Quaternion.fromAxisAngle(Vector3.forward(), rotation())
+        )
```
jdm717, line 136: This `appliedToCurrent` behavior fixes an issue where starting rotations on each axis would result in incorrect rotation calculation. This then caused the input to jump from something like `270` to `277` when pressing enter

packages/viewer/src/components/viewer-transform-widget/util.ts
```diff
@@ -99,22 +100,40 @@ export function computeInputTransform(
 
   switch (identifier) {
     case 'x-translate':
-      return Matrix4.makeTranslation(Vector3.create(position(), 0, 0));
+      return appliedToCurrent(
```
MadisonEhlers-Vertex, line 103: nit: Should the appliedToCurrent be invoked one time and the underlying translation that changes be moved to another function? That way the wrapped function can be called once. This is probably more of a stylistic preference. So take it or leave it. :) 



Something like:


```typescript
function computedTranslation(identifier) {
  switch (identifier)
    case 'x-translate':
   return Matrix4.makeTranslation(Vector3.create(position(), 0, 0))
   
   ... case 'y-translate': ...
   }
```

Then this function can just return  something like:

```typescript
return appliedToCurrent( current, computedTranslation(identifier));
```

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget-components.tsx
```diff
@@ -78,24 +82,28 @@ export const TransformWidgetInput: FunctionalComponent<
...
+    const commandOrControlModifier = event.ctrlKey || event.metaKey;
 
     if (event.key === 'ArrowUp') {
       onIncrement?.();
     } else if (event.key === 'ArrowDown') {
       onDecrement?.();
+    } else if (event.key === 'z' && commandOrControlModifier) {
+      event.preventDefault();
+
+      onUndo?.();
```
MadisonEhlers-Vertex, line 94: in what scenarios is `onUndo` not defined? Should we throw an exception or warning if it's not defined and we're trying to do the keybind?

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.css
```diff
@@ -52,30 +52,52 @@
...
+  position: relative;
   text-align: center;
   box-sizing: border-box;
   width: 75%;
+  height: 100%;
   outline: none;
   border-width: 0 1px 0 0;
-  border-color: #cccccc;
+  border-color: #444444;
+  border-radius: 2px 0 0 2px;
```
MadisonEhlers-Vertex, line 81: 🥇

MadisonEhlers-Vertex (approved):
Left a few minor suggestions and thoughts. Otherwise lgtm

packages/viewer/src/components/viewer-transform-widget/util.ts
jdm717: Yeah, good thought. I'll take a look at reorganizing this method. Might be some other reductions in logic that'd be possible in these utilities. I'll take a look

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget-components.tsx
jdm717: Yeah, this could probably just always be defined. I'll make that change 👍

MadisonEhlers-Vertex (approved):

PR #584, opened 2024-03-20
[PLAT-3893] Support standard views from the ViewCube without viewAll

jdm717:
## Summary

Adds a `viewAll` flag that can be set to `false` on the `<vertex-viewer-view-cube>` component that modifies the behavior to maintain the current `lookAt`, and align the `viewVector` to the standard view.

This PR also fixes a small bug with the transform inputs where the input would still reflect the prior axis value until some movement had occurred.

## Test Plan

- Verify the standard view cube behavior `viewAll={true}` continues to work as expected
- Verify the updated view cube behavior `viewAll={false}` maintains the current `lookAt` point 

## Release Notes

N/A

## Possible Regressions

View cube orientation

## Dependencies

N/A


amvertex (approved):
Nice job!

amvertex (approved):

PR #586, opened 2024-03-22
[PLAT-4008] Update utils and html-templates package versions manually

jdm717:
## Summary

This PR updates the versions of the `@vertexvis/utils`, `@vertexvis/html-templates`, and `@vertexwebsdk/build` packages manually, as well as updates them in other package dependencies. 

In addition, this PR updates the `detect_release` script to use the `@vertexvis/utils` package version temporarily to ensure we detect a release for the packages here, and updates the `publish_release` script to skip the tag creation, as we already have one pointing to the correct point in history. These two changes will be reverted with a PR immediately following this one. (https://github.com/Vertexvis/vertex-web-sdk/pull/587)

## Test Plan

- Verify that the `npx lerna changed` command correctly shows that the `@vertexvis/utils` and `@vertexvis/html-templates` have changed
- Verify that the `publish-release` action runs and publishes these new versions (this should work, as the [docs](https://lerna.js.org/docs/features/version-and-publish#from-package) for `lerna publish from-package` command specify that it should compare each package's version to the published version and only publish these two, but this may need follow-up)

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


scripts/publish_release.sh
```diff
@@ -11,15 +11,15 @@ sha="$(git rev-parse HEAD)"
...
-{
-  "tag_name": "$version",
-  "target_commitish": "$sha",
-  "name": "$version",
-  "body": "Automated release for $version\n",
-  "draft": false,
-  "prelease": false
-}
-EOF
+# curl -s -X POST https://api.github.com/repos/$REPOSITORY/releases \
```
amvertex, line 14: If we're leaving this here for now, can you add a comment explaining the problem?

jdm717: I can certainly also remove it, but my intention is to remove it with https://github.com/Vertexvis/vertex-web-sdk/pull/587 immediately after this publishes. This is hopefully just a temporary 15 minute (assuming all goes well) state while waiting for it to publish

amvertex: Sounds good 👍

amvertex (approved):

PR #590, opened 2024-03-27
[PLAT-4000] Add transform widget example

jdm717:
## Summary

Adds an example for the transform-widget that demonstrates usage with hit results and selection.

## Test Plan

- Run the `yarn examples:start` command and navigate to the transform-widget demo
- Verify selecting parts displays the transform widget, and selected parts can be transformed
- Verify the transform widget shows up with the frame displaying selection rather than immediately
- Verify clicking in whitespace clears selection and removes the transform widget
- Verify pressing `r` resets the scene and removes the transform widget

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


amvertex (approved):
Looks great! Thanks for improving the examples.

PR #596, opened 2024-05-01
[PLAT-4258] Fix view cube in WebKit browsers

danschultz:
## Summary

Fixes an issue in WebKit based browsers where the view cube is flattened. This appears to happen because of a bug in WebKit with the `transform-style: preserve-3d` behavior not cascading through to `<slot>` children. Fix is to have the 3D renderers and elements not use shadow DOM and instead use scoped styles.

https://vertexvis.atlassian.net/browse/PLAT-4258

### DIFF
--- a/packages/viewer/src/components/viewer-view-cube/viewer-view-cube.css
+++ b/packages/viewer/src/components/viewer-view-cube/viewer-view-cube.css
@@ -69,7 +69,7 @@
   position: unset;
   width: 100%;
   height: 100%;
-  overflow: visible;
+  overflow: visible !important;
 }
 
 .reference-point {
@@ -218,6 +218,12 @@
   background-color: black;
   opacity: 0.12;
   filter: blur(4px);
+}
+
+.cube-side-face,
+.cube-corner-face,
+.cube-edge-face {
+  pointer-events: initial;
 }
 
 .cube-corner .cube-corner-face,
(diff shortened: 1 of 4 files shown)


packages/viewer/src/components/viewer-view-cube/viewer-view-cube.css
```diff
@@ -69,7 +69,7 @@
   position: unset;
   width: 100%;
   height: 100%;
-  overflow: visible;
+  overflow: visible !important;
```
MadisonEhlers-Vertex, line 72: Is the `!important` required here? Usually a smell, but given the state of this prior, definitely open to this! :)

MadisonEhlers-Vertex (approved):

packages/viewer/src/components/viewer-view-cube/viewer-view-cube.css
MadisonEhlers-Vertex: BTW, tested locally with a linked version.... I'm so glad you found a fix for this!

danschultz: Yes unfortunately. The change to scoped styles changes the specificity of the selectors.

amvertex (approved):
Great job figuring this out!

PR #597, opened 2024-05-02
[PLAT-4175] Support occlusion for pins and annotation callouts

amvertex:
## Summary
This PR includes support for conditionally formatting pins and annotation callouts depending on whether they are occluded. 

As part of this work, we discovered that we were incorrectly updating the orthographic camera after some types of model interactions, which was preventing the correct occlusion state from being reported.  Specifically, we were not setting the camera's lookAt point in exactly the same place in the sdk and the back-end after a rotation or FlyToPositionKeyInteraction with orthographic cameras. The updated lookAt point is at the same spot in the viewport (2D) as it was previously, but has a different world position (3D). 

## Test Plan
Verify that:
- Rotating and zooming models works correctly using both perspective and orthographic cameras
- Placing and viewing pins and annotation callouts works as expected
  -  Pins and annotation callouts should appear transparent when they are occluded

## Release Notes
Support occlusion for pins and annotation callouts

## Possible Regressions
- Rotating and zooming a model
- Placing and viewing pins and annotation callouts

## Dependencies
None


packages/viewer/src/lib/interactions/interactionApiOrthographic.ts
```diff
@@ -164,62 +170,77 @@ export class InteractionApiOrthographic extends InteractionApi<OrthographicCamer
...
+          this.orthographicZoomData == null ||
+          Point.distance(point, this.orthographicZoomData.startingScreenPt) > 2
+        ) {
+          const frameCam = camera.toFrameCamera();
+          const dir = frameCam.direction;
+          const ray = viewport.transformPointToOrthographicRay(
+            point,
+            frame.image,
+            frameCam
+          );
```
danschultz, line 185: It's strange that we have `transformPointToOrthographicRay` and `transformPointToRay`. You could pass an incorrect camera type to either of these. Instead, I'd suggest just having a single `transformPointToRay` and that method inspects what type of camera is passed to it.

Also not sure why we have an interface for `FrameCameraWithMatrices`. It seems awkward to have this interface in addition to `FrameCameraBase`.

danschultz: Also, understand this isn't from your changes. I'll create a ticket to follow up.

packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -372,26 +372,39 @@ export abstract class InteractionApi<T extends Camera = Camera> {
...
+        // center of the bounding box to maintain zoom and pan behavior.
+        lookAt: Vector3.add(
+          Vector3.scale(
+            Math.abs(camera.signedDistanceToBoundingBoxCenter()) /
+              Vector3.magnitude(updated.viewVector),
+            updated.viewVector
+          ),
+          updated.position
+        ),
+      });
```
danschultz, line 407: Unless these changes are to fix a camera positioning issue, I wouldn't expect any changes here to fix computation of the depth buffer.

packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -370,25 +370,25 @@ export abstract class InteractionApi<T extends Camera = Camera> {
...
         y: delta.x * crossX.y + delta.y * crossY.y,
         z: delta.x * crossX.z + delta.y * crossY.z,
       });
 
-      const rotationAxis = Vector3.cross(mouseToWorld, lookAt);
+      const rotationAxis = Vector3.cross(mouseToWorld, directionVector);
 
-      const epsilonX = (3.0 * Math.PI * delta.x) / viewport.width;
-      const epsilonY = (3.0 * Math.PI * delta.y) / viewport.height;
+      // The 9.5 multiplier was chosen to match the desired rotation speed
```
amvertex, line 389: Since we're also multiplying by 3 (to set the desired rotation speed), there's no reason to multiply by PI. We can multiply by 9.5 instead, which will have essentially the same results and be slightly faster.

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -409,7 +409,7 @@ export abstract class Camera {
 
     return this.update({
       up: standardView.up,
-      position: Ray.at(standardViewRay, Vector3.magnitude(this.viewVector)),
+      position: updatedPosition,
```
amvertex, line 412: This is the same result - we don't need to calculate it twice.

packages/viewer/src/lib/interactions/interactionApiOrthographic.ts
amvertex: I agree. I created https://vertexvis.atlassian.net/browse/PLAT-4344.

packages/viewer/src/components/viewer-annotation-callout/viewer-annotation-callout.tsx
```diff
@@ -20,11 +29,79 @@ export class ViewerAnnotationCallout {
    */
   @Prop() public iconSize: ViewerIconSize = 'sm';
 
+  /**
+   * @internal
+   * Whether the callout is occluded
+   */
+  @Prop({ mutable: true })
+  public occluded = false;
```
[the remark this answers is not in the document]
amvertex: The current implementation supports only calculating the occlusion state on final frames, which was the desired behavior.

packages/viewer/src/components/viewer-annotation-callout/viewer-annotation-callout.css
```diff
@@ -46,3 +48,8 @@
 .icon {
   color: var(--viewer-annotation-callout-accent-color);
 }
+
+.occluded {
+  opacity: .3;
+  transition: opacity 0.75s ease-in;
+}
```
danschultz, line 55: There's no way for an application to customize these styles. Let's remove this and let the application (Connect) provide their styles for occlusion. This has the added benefit of not having to make changes to the SDK if we want to make styling changes in Connect. Related https://github.com/Vertexvis/vertex-web-sdk/pull/597/files/58a5dcd89baab5b59f5cc6df6d09b563d2c19af7#r1588393719

packages/viewer/src/components/viewer-annotation-callout/viewer-annotation-callout.css
```diff
@@ -23,6 +23,8 @@
 
   background-color: var(--viewer-annotation-callout-primary-color);
   border: 1px solid var(--viewer-annotation-callout-accent-color);
+
+  transition: opacity 0.5s ease-in;
```
danschultz, line 27: ^

packages/viewer/src/components/viewer-pin-group/viewer-pin-group.tsx
```diff
@@ -64,18 +66,47 @@ export class ViewerPinGroup {
...
    */
   @Prop()
   public selected = false;
 
+  /**
+   * @internal
+   * Whether the pin is occluded
+   */
+  @Prop({ mutable: true })
+  public occluded = false;
```
danschultz, line 79: I would suggest the following:

* Make this property part of the public API (non-internal), and pass `reflect: true` to the decorator. This way the application can provide it's own styles for occlusion.  
  ```ts
  /**
   * Whether the pin is occluded.
   */
  @Prop({ mutable: true, reflect: true })
  public occluded = false;
  ```
* We're replicating a lot of the behavior that already exists in `<vertex-viewer-dom-element>`. What if we instead exposed a `onOcclusionChanged` event on `<vertex-viewer-dom-element>` that this component listened for and would then set the `occluded` attribute whenever this event is fired?

packages/viewer/src/components/viewer-pin-group/viewer-pin-group.css
```diff
@@ -89,9 +89,15 @@
...
 }
 
 .pin-selected {
   stroke-width: 1;
   stroke: var(--viewer-annotations-pin-selected-stroke);
+}
+
+.pin-occluded {
+  opacity: .3;
+  transition: opacity 0.75s ease-in;
```
danschultz, line 102: Same as a prior comment, there's no way for an application to customize these styles and requires that we make changes to the SDK if we want to change styles in Connect. Lets remove this and put the styling in Connect.

packages/viewer/src/lib/types/frame.ts
```diff
@@ -427,7 +428,8 @@ export class FrameOrthographicCamera
     public readonly near: number,
     public readonly far: number,
     public readonly aspectRatio: number,
-    public readonly fovHeight: number
+    public readonly fovHeight: number,
+    public readonly rotationPoint?: Vector3.Vector3
```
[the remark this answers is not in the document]
amvertex: No, we only update the lookAt after a rotation: https://github.com/Vertexvis/vertex-web-sdk/pull/597/files#diff-a88c3b655d377ef7b3dd9503cdb49b4b780a8213994f2a64fa932483732ce2fdR303-R307. This allows for consecutive rotations to continue working as they currently do. If we don't keep the rotationPoint constant, then manually setting the lookAt point with FlyToPositionKeyInteraction and then rotating will not keep the lookAt point at the center of the viewer.

packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -370,25 +370,25 @@ export abstract class InteractionApi<T extends Camera = Camera> {
    *  viewer.
    */
   public async rotateCamera(delta: Point.Point): Promise<void> {
-    return this.transformCamera(({ camera, viewport }) => {
+    return this.transformCamera(({ camera, viewport, boundingBox }) => {
       const upVector = Vector3.normalize(camera.up);
-      const lookAt = Vector3.normalize(
+      const directionVector = Vector3.normalize(
```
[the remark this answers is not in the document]
amvertex: No, _directionVector_ is consistent because it's normalized. In our code, `viewVector = Vector3.subtract(camera.lookAt, camera.position)` and the direction vector is the normalized viewVector.

packages/viewer/src/lib/interactions/flyToPositionKeyInteraction.ts
```diff
@@ -38,6 +38,9 @@ export class FlyToPositionKeyInteraction
       hitResult.hitItems.hits[0].hitPoint != null
     ) {
       const camera = (await this.sceneProvider()).camera();
+      const boundingSphere = BoundingSphere.create(
+        (await this.sceneProvider()).boundingBox()
+      );
```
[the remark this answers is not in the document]
amvertex: Updated!

packages/viewer/src/lib/interactions/flyToPositionKeyInteraction.ts
```diff
@@ -75,4 +86,30 @@ export class FlyToPositionKeyInteraction
...
+      const viewVectorMagnitudeSquared = Vector3.magnitudeSquared(viewVector);
+      const offset = orthogonalOffset / viewVectorMagnitudeSquared;
+
+      const scaledViewVector = Vector3.scale(offset, viewVector);
+      return Vector3.add(scaledViewVector, hitPoint);
+    } else {
+      // For perspective, just return the hit point
+      return hitPoint;
+    }
+  }
```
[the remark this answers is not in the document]
amvertex: Updated!

packages/viewer/src/lib/interactions/interactionApiOrthographic.ts
```diff
@@ -164,61 +174,137 @@ export class InteractionApiOrthographic extends InteractionApi<OrthographicCamer
...
+        angle,
+        rotationPoint,
+        rotationAxisDirection
+      );
+
+      // Update the lookAt point to take the center of the model into account
+      // This change helps ensure that the lookAt point is consistent between
+      // the SDK and back-end system such that the calculated depth buffer is correct.
+      const updatedCenterPoint = Vector3.subtract(
+        BoundingSphere.create(boundingBox).center,
```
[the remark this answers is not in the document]
amvertex: Updated!

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -791,6 +795,25 @@ export class OrthographicCamera
...
+      rotationPoint != null &&
+      rotationPoint.x != null &&
+      rotationPoint.y != null &&
+      rotationPoint.z != null
+    ) {
+      return rotationPoint;
+    } else {
+      return undefined;
+    }
+  }
```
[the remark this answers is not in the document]
amvertex: Updated!

packages/viewer/src/lib/types/depthBuffer.ts
```diff
@@ -216,15 +217,33 @@ export class DepthBuffer implements FrameImageLike {
   public isOccluded(worldPt: Vector3.Vector3, viewport: Viewport): boolean {
     const { position, direction, projectionViewMatrix } = this.camera;
 
+    // Calculate the distance from the camera to the given world point
     const eyeToPoint = Vector3.subtract(worldPt, position);
-    const projected = Vector3.project(eyeToPoint, direction);
-    const distance = Vector3.magnitude(projected);
+    const distanceToPoint = Math.abs(Vector3.dot(eyeToPoint, direction));
```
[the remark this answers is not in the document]
amvertex: I added a comment. This is doing the same math as before, but we were unnecessarily multiplying by the magnitude of a vector only to immediately divide by the  magnitude of the vector.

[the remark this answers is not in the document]
danschultz: It's not quite the same. `project` applies a scalar that's missing here.

packages/viewer/src/lib/types/depthBuffer.ts
```diff
@@ -216,15 +217,33 @@ export class DepthBuffer implements FrameImageLike {
...
+    // Use the correct calculation for the camera type
+    const isPerspectiveCamera = this.camera.isPerspective();
+    const depthOfClosestGeometry = isPerspectiveCamera
+      ? this.getLinearDepthAtPoint(scaledPt)
+      : this.getOrthographicDepthAtPoint(scaledPt);
+
+    // Allow for a small rounding error
+    // Note that if the world point is coincident with the geometry,
+    // we want to err on the side of returning not occluded
+    const allowableDifference = Math.abs(0.02 * distanceToPoint);
```
danschultz, line 242: @mikemccarty-vertex noticed this, but assuming the math above is the same, I don't think this will apply a uniform tolerance across all points. In particular, the tolerance should be adjusted the further you move from the center of the viewport.

mikemccarty-vertex: Ideally, the tolerance is a function of something like `far - near` -- i.e. one that is constant for all pins in the view.

packages/viewer/src/lib/types/depthBuffer.ts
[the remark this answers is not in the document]
amvertex: The direction vector is already normalized, so its magnitude is 1. As a result, the scalar is only 1.

packages/viewer/src/lib/types/depthBuffer.ts
amvertex: Updated!

packages/viewer/src/components/viewer-annotation-callout/viewer-annotation-callout.css
amvertex: I removed the styling here.

packages/viewer/src/components/viewer-annotation-callout/viewer-annotation-callout.css
amvertex: I removed the styling here.

packages/viewer/src/components/viewer-pin-group/viewer-pin-group.tsx
amvertex: I removed the duplicated logic and set up an event that fires when the occluded attribute is fired.

packages/viewer/src/lib/types/frame.ts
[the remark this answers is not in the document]
danschultz: Ok, it sounds like this property is used for the purpose of interaction vs used for either a) generation of matrices or b) used to generate a frame. We should find a new home for this as fields related to interactions shouldn't be the responsibility of the Camera.

danschultz (approved):

packages/viewer/src/components/viewer-annotation-callout/viewer-annotation-callout.tsx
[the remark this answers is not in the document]
danschultz: Connect should be wrapping a `<vertex-viewer-annotation-callout>` in a `<vertex-viewer-dom-element>` to provide the callout positioning and occlusion behavior. You can apply the occlusion styling to the wrapping `<vertex-viewer-dom-element>` without adding all this other code.

MadisonEhlers-Vertex (approved):

packages/viewer/src/lib/types/depthBuffer.ts
amvertex: Actually, while I agree that it we should look into setting the tolerance such that it is constant for all pins in the view, I don't like the current results we're getting after changing it to a function of (far-near), so I reverted back to using a function of the distanceToPoint. Further work here can be included in the follow-up work.

MadisonEhlers-Vertex (approved):

PR #599, opened 2024-05-08
[PLAT-4304] Determine if point is detached from geometry

amvertex:
## Summary
This PR includes support for determining whether pins and annotation callouts are attached to geometry. Pins will automatically appear transparent when they are not attached to geometry. 

## Test Plan
Verify that placing and viewing pins and annotation callouts works as expected, specifically that they should appear transparent when they are not attached to geometry. This can be easily tested by hiding the part that the item was originally placed on.

## Release Notes
Support for determining whether pins and annotation callouts are attached to geometry

## Possible Regressions
Viewing pins and annotation callouts

## Dependencies
None


packages/viewer/src/components/viewer-dom-element/viewer-dom-element.tsx
```diff
@@ -198,6 +198,39 @@ export class ViewerDomElement implements HTMLDomRendererPositionableElement {
...
   public occlusionStateChanged!: EventEmitter<boolean>;
 
+  /**
+   * Disables detached testing for this element. Defaults to enabled. When
+   * enabled, the elements position will be tested against the current depth
+   * buffer. If the position is detached, then the `detached` attribute will be
+   * set.
+   */
+  @Prop()
+  public detachedOff = false;
```
danschultz, line 208: You'd want what sounds like an experimental feature to be opt-in, no?

packages/viewer/src/lib/types/depthBuffer.ts
```diff
@@ -247,4 +247,44 @@ export class DepthBuffer implements FrameImageLike {
...
+
+    // Find the depth of the closest geometry at the same point on the screen
+    // Use the correct calculation for the camera type
+    const isPerspectiveCamera = this.camera.isPerspective();
+    const depthOfClosestGeometry = isPerspectiveCamera
+      ? this.getLinearDepthAtPoint(scaledPt)
+      : this.getOrthographicDepthAtPoint(scaledPt);
+
+    // Allow for a small rounding error
+    const depthDifference = Math.abs(depthOfClosestGeometry - distanceToPoint);
```
danschultz, line 281: Consider extracting this and sharing it between `isOccluded` and `isDetached`.

packages/viewer/src/components/viewer-dom-element/viewer-dom-element.tsx
danschultz: Would also recommend we communicate somewhere in the docs that this is experimental.

amvertex: Good idea, updated!

packages/viewer/src/lib/types/depthBuffer.ts
amvertex: Updated!

PR #604, opened 2024-06-04
[PLAT-4344] Consolidate camera calculations for perspective and orthographic cameras

amvertex:
## Summary
This PR includes several improvements consolidate logic around camera calculations from individual methods for perspective and orthographic cameras into single methods.  This should reduce errors in the future where a user selected the incorrect method for their camera type.

## Test Plan
Verify both perspective and orthographic cameras work as expected

## Release Notes
Improvements to consolidate logic around camera calculations from individual methods for perspective and orthographic cameras into single methods. Specifically,
- Removes `viewport.transformPointToOrthographicRay` in favor of `viewport.transformPointToRay`
- Removes `viewport.transformPointToOrthographicWorldSpace` in favor of `viewport.transformPointToWorldSpace`
- Removes `depthBuffer.getLinearDepthAtPoint` and `depthBuffer.getOrthographicDepthAtPoint` in favor of `depthBuffer.getDepthAtPoint`

## Possible Regressions
Camera calculations

## Dependencies
None

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "private": true,
-  "nextVersionBump": "patch",
+  "nextVersionBump": "minor",
   "devDependencies": {
     "eslint": "^8.17.0",
     "http-server": "^14.1.1",
--- a/packages/viewer/src/lib/types/frame.ts
+++ b/packages/viewer/src/lib/types/frame.ts
@@ -167,28 +167,7 @@ interface FrameCameraMatrices {
   readonly worldMatrix: Matrix4.Matrix4;
 }
 
-interface FrameCameraLike {
-  readonly position: Vector3.Vector3;
-  readonly lookAt: Vector3.Vector3;
-  readonly up: Vector3.Vector3;
-  readonly near: number;
-  readonly far: number;
-  readonly aspectRatio: number;
-}
-
-interface FramePerspectiveCameraLike {
-  readonly fovY: number;
-}
-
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "private": true,
-  "nextVersionBump": "patch",
+  "nextVersionBump": "minor",
```
danschultz, line 4: 👍

packages/viewer/src/lib/types/frame.ts
```diff
@@ -356,7 +356,7 @@ export interface FrameCameraWithMatrices extends FrameCameraBase {
 
 export class FramePerspectiveCamera
   extends FrameCameraBase
-  implements FrameCameraWithMatrices, FramePerspectiveCameraLike
+  implements FramePerspectiveCameraLike
```
danschultz, line 359: My vote would be to remove these interface definitions as well:

* `FrameCameraLike`
* `FramePerspectiveCameraLike`
* `FrameOrthographicCameraLike`

I think this'd help reduce complexity and confusion.

danschultz (commented):
This is a world better. Have one suggestion, otherwise LGTM.

packages/viewer/src/lib/types/frame.ts
amvertex: Good idea. Removed!

danschultz (approved):

PR #605, opened 2024-06-11
[PLAT-4416] Handling Viewer events on external dom elements

MadisonEhlers-Vertex:
## Summary

These changes implement a new abstract class labeled `MultiElementInteractionHandler` that allows any interaction handler that extends it to accept external elements as elements to interact on. These changes were introduced so that the dom-renderer elements could propagate specific kinds of events over to the viewer. This enables us to zoom on annotation callouts, and also to do standard viewer actions on dom elements. 

The dom-renderer exposes a boolean prop `propagateEventsToViewer` that allows dom-renderers to opt-out of this feature, but is enabled by default.

Here is a video of propogating these events to the viewer:

https://github.com/Vertexvis/vertex-web-sdk/assets/49169871/018e4fdd-f1c7-4afa-b263-0f69087794c8

## Test Plan

- Verify the Zoom interaction propoagates via mouse event when hovering over a dom-element in the dom-renderer.


MadisonEhlers-Vertex (commented):
At a high level, this enables "external" elements to register themselves with the viewers interaction handlers. 

A few questions I have: 

1. Does it make sense to have a list of `externalElements` within the base interaction handler that tracks each element, so that on a `dispose()` we can properly remove those interactions from the interaction handler side?

2. Does it make sense to add a prop on the dom-renderer that allows an opt-out/opt-in on this feature? For instance, a property like `propagateEventsToViewer` (Default true)?

3. Does this implementation need to enable a more specific set of events to opt-in/out of or is coupling this feature with the interaction handlers a good plan? (For instance, should we instead have a solution that accepts a list of events ex: [`wheel`, `pointerDown`, etc.]

jdm717:
> At a high level, this enables "external" elements to register themselves with the viewers interaction handlers.
> 
> A few questions I have:
> 
> 1. Does it make sense to have a list of `externalElements` within the base interaction handler that tracks each element, so that on a `dispose()` we can properly remove those interactions from the interaction handler side?
> 2. Does it make sense to add a prop on the dom-renderer that allows an opt-out/opt-in on this feature? For instance, a property like `propagateEventsToViewer` (Default true)?
> 3. Does this implementation need to enable a more specific set of events to opt-in/out of or is coupling this feature with the interaction handlers a good plan? (For instance, should we instead have a solution that accepts a list of events ex: [`wheel`, `pointerDown`, etc.]

1. I think it does make sense to have the interaction handler track those values. Without some sort of tracking within the handler, it seems like we could encounter memory leaks with event listeners being added and never removed in the case that the interaction handler is recreated. I know at the very least that can happen when the `cameraControls` property changes. That being said, We'd still have to have some way of detecting the change on the dom-renderer side so it could re-register with the new one.
2/3. I do think we'd want to have an opt-in/opt-out behavior, since this changes the default behavior of these dom-elements. I could also see cases where we might want different behavior depending on the element itself (for instance, the view-cube might not want these events to propagate). I do wonder if it might make sense to define a set of "viewer interactions" rather than events specifically like `wheel` or `pointerdown` since those are a bit dependent on the underlying implementation in the handler itself. Maybe options like `['tap', 'rotate', 'zoom', ...]` could explain the behavior a bit more? That being said, the `tap` event is a little less clear to me on whether there's a use case for allowing those.

packages/viewer/src/components/viewer-pin-group/viewer-pin-group.tsx
```diff
@@ -141,7 +141,7 @@ export class ViewerPinGroup {
           data-testid={`drawn-pin-${this.pin.id}`}
           position={this.pin.worldPosition}
           onPointerDown={(e) => {
-            if (e.buttons !== 2) {
+            if (this.pinController?.getToolMode() === 'edit') {
```
MadisonEhlers-Vertex, line 144: Ideally we won't have to stop propagation, but in edit mode, the text label has not submitted, so dragging an anchor causes the latest text to not be saved.

packages/viewer/src/components/viewer-dom-renderer/viewer-dom-renderer.tsx
```diff
@@ -39,6 +41,16 @@ export class ViewerDomRenderer {
   @Prop()
   public drawMode: ViewerDomRendererDrawMode = '3d';
 
+  /**
+   * Specifies whether or not to propogate events to the viewers' interaction handlers
```
amvertex, line 45: Nit: Could change to `Specifies whether to propagate events to the viewer's interaction handlers`

amvertex (approved):

amvertex (approved):

jdm717 (approved):
LGTM!

PR #609, opened 2024-06-18
[PLAT-4504] Customizable Mouse Related Zoom Interactions

MadisonEhlers-Vertex:
## Summary

These changes are to introduce a shorter delay in requesting a final frame on a ZoomInteraction. The 350ms was determined by performing a series of interactions from multiple internal users. Ultimately the feedback we received to make the change here is that the Zoom interaction felt slower. This should help a bit so that a final frame gets queued a little sooner.

The delay is not shortened further because the mouse `wheel` events could be invoked by the user in quick succession, which cause this zoom interaction to be "jerky" or "jumpy". This value of 350 appears to be high enough to avoid that, while still requesting the final frame quite a bit sooner. This value is also customizable if a consumer would like to override the default 350 in favor of a different debounce.

These changes also include the ability to reverse the mouse zoom direction as a viewer setting. 

## Test Plan

Verify that the mouse wheel zoom, and 2 finger touch pad zoom interactions request final frames slightly sooner than the previous implementation

### DIFF
--- a/packages/viewer/src/lib/interactions/mouseInteractions.ts
+++ b/packages/viewer/src/lib/interactions/mouseInteractions.ts
@@ -1,6 +1,7 @@
 import { Point } from '@vertexvis/geometry';
 
 import { getMouseClientPosition } from '../dom';
+import { defaultMouseWheelInteractionEndDebounce } from '../types/interactions';
 import { InteractionType } from './baseInteractionHandler';
 import { InteractionApi, InteractionConfigProvider } from './interactionApi';
 
@@ -114,8 +115,6 @@ export class ZoomInteraction extends MouseInteraction {
   private interactionTimer: number | undefined;
   private startPt?: Point.Point;
 
-  private defaultInteractionDelay = 350;
-
   public constructor(
     private interactionConfigProvider: InteractionConfigProvider
   ) {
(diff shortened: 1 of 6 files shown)


amvertex (approved):
Thanks for all the testing to determine the new value!

packages/viewer/src/lib/interactions/mouseInteractions.ts
```diff
@@ -139,7 +141,7 @@ export class ZoomInteraction extends MouseInteraction {
       const delta = Point.subtract(position, this.currentPosition);
 
       if (this.startPt != null) {
-        api.zoomCameraToPoint(this.startPt, delta.y);
+        this.zoomCameraToPointWithDirectionConfig(this.startPt, -delta.y, api);
```
MadisonEhlers-Vertex, line 144: This `-delta.y` is intentional in order to sync up the zoom by mouse drag to be the same as the mouse wheel, in that a mouse-forward drag produces the same result as a mouse wheel forward event.

packages/viewer/src/lib/types/interactions.ts
```diff
@@ -26,11 +26,29 @@ export interface InteractionConfig {
...
+  reverseMouseZoomDirection: boolean;
 }
 
 export const defaultInteractionConfig: InteractionConfig = {
   finePointerThreshold: 1,
   coarsePointerThreshold: 3,
   interactionDelay: 75,
   useMinimumPerspectiveZoomDistance: true,
+  mouseWheelInteractionEndDebounce: 350,
+  reverseMouseZoomDirection: false,
```
MadisonEhlers-Vertex, line 53: FWIW, I did confirm that the deep partial config mapping was working correctly.

amvertex (approved):

jdm717:
With @MadisonEhlers-Vertex out, I've pushed an update to this PR to reflect further discussion around our desired behavior here. Specifically, this PR has been updated to only reverse the `wheel`-based zoom interactions, leaving drag-based zoom interactions unchanged. The flag for configuration has also been changed back to `reverseMouseWheelDirection` to reflect that it only applies to the `wheel` events.

CC @amvertex

amvertex (approved):
Thanks for finishing this @jdm717!

PR #611, opened 2024-07-08
[PLAT-4159] Add types for model views

danschultz:
## Summary

Adding types for model views so they're easier to work with than PBs.

https://vertexvis.atlassian.net/browse/PLAT-4159


jdm717 (approved):
LGTM

PR #616, opened 2024-07-24
[PLAT-4149] Add support for loading and unloading ModelViews

jdm717:
## Summary

Updates the ModelViewController to support loading and unloading of ModelViews.

## Test Plan

- Verify that calling `load` on the ModelViewController makes an UpdateSceneView gRPC request with both the current scene view ID and provided model view ID
- Verify that calling `unload` on the ModelViewController makes an UpdateSceneView gRPC request with the current scene view ID, then resets the scene view

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/model-views/controller.ts
+++ b/packages/viewer/src/lib/model-views/controller.ts
@@ -3,11 +3,14 @@ import { Uuid2l } from '@vertexvis/scene-view-protos/core/protos/uuid_pb';
 import {
   ListItemModelViewsRequest,
   ListItemModelViewsResponse,
+  UpdateSceneViewRequest,
 } from '@vertexvis/scene-view-protos/sceneview/protos/scene_view_api_pb';
 import { SceneViewAPIClient } from '@vertexvis/scene-view-protos/sceneview/protos/scene_view_api_pb_service';
 import { UUID } from '@vertexvis/utils';
+import { FieldMask } from 'google-protobuf/google/protobuf/field_mask_pb';
 
 import { createMetadata, JwtProvider, requestUnary } from '../grpc';
+import { Scene } from '../scenes';
 import { mapListItemModelViewsResponseOrThrow } from './mapper';
 import { ModelViewListResponse } from './types';
 
@@ -23,7 +26,8 @@ export class ModelViewController {
   public constructor(
     private client: SceneViewAPIClient,
     private jwtProvider: JwtProvider,
-    private deviceIdProvider: () => string | undefined
+    private deviceIdProvider: () => string | undefined,
+    private sceneProvider: () => Promise<Scene>
   ) {}
 
   /**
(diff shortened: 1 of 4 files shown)


amvertex (approved):

packages/viewer/src/lib/model-views/controller.ts
```diff
@@ -63,4 +67,68 @@ export class ModelViewController {
...
+      req.setModelViewId(svUuid2l);
+
+      const mvUuid = UUID.toMsbLsb(modelViewId);
+      const mvUuid2l = new Uuid2l();
+      mvUuid2l.setMsb(mvUuid.msb);
+      mvUuid2l.setLsb(mvUuid.lsb);
+      req.setModelViewId(mvUuid2l);
+
+      const mask = new FieldMask();
+      mask.addPaths('sceneView.modelViewId');
```
sourada-e5, line 97: I'm not at all sure I understand the conventions around field masks, but looking at [this documentation](https://protobuf.dev/reference/protobuf/google.protobuf/#field-mask), it seems like the mask should be of the form `model_view_id`, so, not including the `sceneView` and using camel case for the field name (since this is gRPC Web rather than REST/JSON). Does that make sense to you (and @danschultz, if you have some opinions here)?

jdm717: Yeah, I think you're right. This at the very least seems like it needs to be snake cased rather than camel cased. https://github.com/Vertexvis/language-guides/blob/main/grpc/standard-fields.md suggests that it should be relative to the resource rather than the request, so I'm not certain if the update request implies we're operating on a `scene_view`, or if we should explicitly call that out

packages/viewer/src/lib/model-views/controller.ts
```diff
@@ -63,4 +67,68 @@ export class ModelViewController {
...
+      const req = new UpdateSceneViewRequest();
+
+      const svUuid = UUID.toMsbLsb(scene.sceneViewId);
+      const svUuid2l = new Uuid2l();
+      svUuid2l.setMsb(svUuid.msb);
+      svUuid2l.setLsb(svUuid.lsb);
+      req.setModelViewId(svUuid2l);
+
+      const mask = new FieldMask();
+      mask.addPaths('sceneView.modelViewId');
```
danschultz, line 124: `mask.addPaths('model_view_id')`

packages/viewer/src/testing/modelViews.ts
```diff
@@ -13,6 +18,33 @@ export function makeListItemModelViewsResponse(
...
+  if (modelViewId != null) {
+    const mvUuid = UUID.toMsbLsb(modelViewId);
+    const mvUuid2l = new Uuid2l();
+    mvUuid2l.setMsb(mvUuid.msb);
+    mvUuid2l.setLsb(mvUuid.lsb);
+    req.setModelViewId(mvUuid2l);
+  }
+
+  const mask = new FieldMask();
+  mask.addPaths('sceneView.modelViewId');
```
danschultz, line 42: ^

packages/viewer/src/lib/model-views/controller.ts
danschultz: BTW, the guide is incorrect as we decided to not use a "source" object.

jdm717: 👍 good to know. Thanks!

danschultz (approved):

PR #617, opened 2024-07-25
[PLAT-3943] Hide detached pins

amvertex:
## Summary
Detached pins and markers should be hidden from the viewer. This PR also updates the method to determine whether a point is detached to ensure that it returns the correct result.

## Test Plan
Verify detached pins are hidden

## Release Notes
Improvements to determining whether a point is detached

## Possible Regressions
Determining whether a point is detached from geometry

## Dependencies
None


jdm717 (approved):
LGTM!

PR #624, opened 2024-08-23
[PLAT-4800] Item model views with annotations filter

danschultz:
## Summary

Adds option to filter item model views with annotations.

https://vertexvis.atlassian.net/browse/PLAT-4800


amvertex (approved):
Looks good!

PR #629, opened 2024-09-09
[PLAT-4743] Include scene comparison information in stream attributes

amvertex:
## Summary
This PR updates the scene compare feature to include the a stream key to compare to on the stream attributes as opposed to within the experimental rendering options.

## Test Plan
Verify scenes can be compared by including a stream key in the comparison information in the stream attributes

## Release Notes
None

## Possible Regressions
Comparing scenes

## Dependencies
None


jdm717 (approved):
LGTM!

PR #630, opened 2024-09-10
[PLAT-4743] Remove ability to compare by scene id

amvertex:
## Summary
We only want to allow users to initiate a scene comparison by providing the stream key. Therefore, this PR removes support for including a scene id to compare to the existing scene.

## Test Plan
Verify version compare works as expected when providing a stream key

## Release Notes
None

## Possible Regressions
Version compare

## Dependencies
None


jdm717 (approved):
LGTM!

PR #634, opened 2024-11-18
[PLAT-5090] Use new scene alteration payload

amvertex:
## Summary
This PR includes support for the new scene alteration payload and a new scene.elements() method.

## Test Plan
Verify scene item interactions continue to work as expected

## Release Notes
- Includes support for the new scene alteration payload
- Includes new scene.elements() method

## Possible Regressions
Interacting with scene items

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/scenes/scene.ts
+++ b/packages/viewer/src/lib/scenes/scene.ts
@@ -70,7 +70,7 @@ export class SceneItemOperationsBuilder
    *
    * // Override the material for the item with the `item-uuid` ID to
    * // be red with an opacity of 0.5.
-   * await scene.items((op) => [
+   * await scene.elements((op) => [
    *   op
    *     .where((q) => q.withItemId('item-uuid'))
    *     .materialOverride(ColorMaterial.create(255, 0, 0, 0.5)),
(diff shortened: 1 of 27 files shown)


packages/viewer/src/lib/scenes/scene.ts
```diff
@@ -710,14 +710,57 @@ export class Scene {
...
+   * ```
+   *
+   * @see {@link RootQuery} for more information on available queries.
+   *
+   * @see {@link SceneItemOperationsBuilder} for more information on available operations to the scene items.
+   *
+   * @param itemOperations
+   */
+  public elements(
+    itemOperations: (q: SceneItemQueryExecutor) => TerminalItemOperationBuilder
```
jdm717, line 744: How do you envision this working on the PMI annotation side? Would it be a second function provided to the `elements()` function? Or did you have something else in mind?

Wondering if there might be a way to organize this so that it's more clear which data sets you're operating on. Just a thought, but maybe the `RootQuery` (or a new concept, but effectively the object passed to the `where` method of the `QueryExecutor`) could have internal definitions for both, so you'd have something like:

```typescript

scene.elements(op => [
  op.where(q => q.items.withItemId(...)).select(),
  op.where(q => q.annotations.withAnnotationId(...)).select(),
])
```

jdm717 (commented):
Overall this looks great - thanks for updating all the examples + documentation! Just had one thought on how we add the PMI annotation queries into this method

PR #640, opened 2025-01-23
[PLAT-1780] Blur scene-tree-table-cell elements on action pointerup

jdm717:
## Summary

Adds a `this.hostEl.blur()` when handling `pointerup` events for scene-tree actions (toggle expansion, isolate, toggle visibility). Because we call `preventDefault` on the `pointerup` event, this was causing the `document.activeElement` to remain as the `<vertex-scene-tree-table-cell>` element, rather than returning focus to the `<body>` element. Because of this, key presses would result in a focus outline appearing on the last pressed action button, which this PR corrects.

As a side-note, the above behavior can be tested by setting a `pointerup` breakpoint in Chrome, and checking the value of `document.activeElement` before and after the event handlers have run.

## Test Plan

- Click a scene-tree action button
- Without focusing another element, press a number or letter key
- Verify that no blue focus outline appears on the button that was clicked

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.spec.tsx
+++ b/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.spec.tsx
@@ -1,3 +1,5 @@
+jest.mock('./utils');
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { h } from '@stencil/core';
 import { newSpecPage, SpecPage } from '@stencil/core/testing';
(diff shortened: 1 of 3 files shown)


packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.spec.tsx
```diff
@@ -545,6 +552,31 @@ describe('<vertex-scene-tree-table-cell>', () => {
...
+        <vertex-scene-tree-table-cell expand-toggle isolate-button visibility-toggle></vertex-scene-tree-table-cell>
+      `,
+      node,
+    });
+
+    const expand = cell.shadowRoot?.querySelector('.expand-btn');
+    const visibility = cell.shadowRoot?.querySelector('.visibility-btn');
+    const isolate = cell.shadowRoot?.querySelector('.isolate-btn');
+
+    console.log(expand);
```
amvertex, line 569: Nit: do you still want this?

amvertex (approved):
Nice job figuring this out!

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.spec.tsx
jdm717: I do not, didn't mean to leave that in 🙃 - removed!

PR #645, opened 2025-02-05
[PLAT-5570] Release v0.23.0

jdm717:
## Summary

This release includes:
* Added support for loading and unloading ModelViews
* Added support for listing the PMI Annotations associated to ModelViews
* (beta) Added support for changing visibility of PMI Annotations
* Added support for detecting and styling "detached" pins that are not placed on visible geometry
* Added support for customizing scene-tree row styles using CSS variables
* Added support for partially applying the shading (noDefaultLights) saved in a scene view state
* Updated scene comparison to use a stream key instead of a scene ID
* Updated styling of the scene-tree table resize divider
* Fixed an issue where pressing any key on the keyboard was highlighting scene-tree action buttons after interacting

## Test Plan

N/A

## Possible Regressions

N/A

## Dependencies

N/A


packages/viewer/src/lib/scenes/scene.ts
```diff
@@ -587,6 +587,11 @@ export class SceneItemOperationsBuilder
 }
 
 /**
+ * @ignore
+ * @beta PMI Annotation operations are being actively developed. These APIs
+ * should be considered unstable, and will potentially change with following
+ * releases.
```
jdm717, line 593: Ignoring + adding a beta tag to all of the PMI Annotation operations so it's clear they're not ready for broader usage yet. Not all of the operations will work today, and we don't provide a formal element for interacting with ModelViews and PMI Annotations

amvertex (approved):

PR #654, opened 2025-03-28
[PLAT-5800] Support camera type overrides

amvertex:
## Summary
This PR adds support for optionally including a camera type override when starting a stream, applying a scene view, or resetting a scene. When the camera type override is included, the desired camera will be converted to the desired type before being applied to the scene.

## Test Plan
Verify a user can include a camera type override when starting a stream, applying a scene view, or resetting a scene 

## Release Notes
Ability to include a camera type override when starting a stream, applying a scene view, or resetting a scene

## Possible Regressions
Loading and resetting views

## Dependencies
Changes in FSS and SS

### DIFF
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -546,7 +546,7 @@ export class Viewer {
     }
 
     if (this.src != null) {
-      this.load(this.src).catch((e) => {
+      this.load(this.src, this.cameraType).catch((e) => {
         console.error('Error loading scene', e);
       });
     }
@@ -768,7 +768,7 @@ export class Viewer {
   @Watch('src')
   public handleSrcChanged(src: string | undefined): void {
     if (src != null) {
-      this.load(src);
+      this.load(src, this.cameraType);
     } else {
       this.unload();
     }
@@ -861,9 +861,10 @@ export class Viewer {
    *  * `urn:vertex:scene:<sceneid>`
    *
    * @param urn The URN of the resource to load.
+   * @param cameraType (Optional) The camera type to load. If not included, the default camera type for the resource will be used.
    */
   @Method()
-  public async load(urn: string): Promise<void> {
+  public async load(urn: string, cameraType?: FrameCameraType): Promise<void> {
     if (this.stream != null && this.dimensions != null) {
       const { EXPERIMENTAL_annotationPollingIntervalInMs } =
         this.getResolvedConfig();
@@ -880,7 +881,8 @@ export class Viewer {
         urn,
         this.clientId,
         this.getDeviceId(),
-        this.getResolvedConfig()
+        this.getResolvedConfig(),
+        cameraType
       );
       this.sceneReady.emit();
 
--- a/packages/viewer/src/components/viewer/viewer.tsx
+++ b/packages/viewer/src/components/viewer/viewer.tsx
@@ -112,6 +112,10 @@ interface ConnectionFailedStatus {
   errorMessage: string;
 }
 
+export interface LoadOptions {
+  cameraType?: FrameCameraType;
+}
+
 /**
  * Internal state values for the component. Used to preserve values across live
  * reload refreshes.
@@ -546,7 +550,7 @@ export class Viewer {
     }
 
     if (this.src != null) {
-      this.load(this.src, this.cameraType).catch((e) => {
+      this.load(this.src, { cameraType: this.cameraType }).catch((e) => {
         console.error('Error loading scene', e);
       });
     }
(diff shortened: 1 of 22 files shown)


packages/viewer/src/lib/types/loadableResource.ts
```diff
@@ -96,6 +105,8 @@ function fromQuery(query?: string): QueryValue[] {
           return { type: 'supplied-id', id: resourceId };
         case 'scene-view-state':
           return { type: 'scene-view-state', id: resourceId };
+        case 'camera-type':
+          return { type: 'camera-type', camera: resourceId as CameraType };
```
jdm717, line 109: Could we move this to avoid having to include it in the URN? I don't think this specific value belongs as part of it, since it's not used in the actual lookup of the resource to be loaded. Maybe we could add a `forcedCameraType` prop or similar to the viewer and include that with all requests if present instead?

jdm717 (commented):
Overall LGTM! Just had one thought on the URN-based camera type

packages/viewer/src/lib/types/loadableResource.ts
amvertex: The camera type will only be used when starting the stream, the same as the initial scene view state, which is why I was thinking it belonged here with the initial scene view state. I'm worried that if we add a _forcedCameraType_ parameter to the viewer, users won't be able to differentiate it from the existing _cameraType_ parameter. Would naming the variable 'initial-camera-type' be better?

jdm717: That's fair - I do think something like `initialCameraType` would be better in that case, maybe even something like `loadCameraType` + pulling that value whenever `src` changes or `load(...)` is called to be explicit that it's only for the loading functions?

amvertex: What do you think of this?

jdm717: So my initial thinking was mainly based on whether the URN schema is the appropriate place to include this type of "initial state", since those are primarily used to give a sort of template that can resolve to an item. Doing a bit of further digging, according to the [RFC q-component section](https://datatracker.ietf.org/doc/html/rfc8141#section-2.3.2), it looks like this is likely still a valid usage of those q-components of a URN. That being said, I still think I'd lean toward making this a viewer`Prop` for the ease of use.

For example, in an application that supports switching camera types + models, you'd have to write logic like:

```typescript

// src is `urn:vertex:stream-key:123/scene-view-states?scene-view-state=234`
function ViewerWrapper({ src, loadCameraType }): JSX.Element {
  const cameraTypeParam = `load-camera-type=${loadCameraType}`;
  const srcWithCameraType = src.includes('?')
      ? `${src}&${cameraTypeParam}`
      : `${src}?${cameraTypeParam}`;

  return <VertexViewer src={srcWithCameraType} />
}
```

to cover cases like the above, where `?scene-view-state=...` or similar parameters could be provided, since you'd have to account for the presence of existing parameters. Of course this logic could live at a higher point in the application to append all these values, but it does seem as though it complicates the approach for providing the initial type compared to:

```typescript
// src is `urn:vertex:stream-key:123/scene-view-states?scene-view-state=234`
function ViewerWrapper({ src, loadCameraType }): JSX.Element {
  return <VertexViewer src={src} loadCameraType={loadCameraType} />
}
```

It'd also make the `loadCameraType` live a bit closer to the `cameraType`, which I think makes sense from a discoverability perspective. All that said - still open to thoughts here, just trying to give this a bit of thought before we add to the URN, as we don't add to that very often

amvertex: Okay, I reworked it to listen to the cameraType parameter earlier than it used to, so it's much simpler now. What do you think?

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -861,9 +861,10 @@ export class Viewer {
    *  * `urn:vertex:scene:<sceneid>`
    *
    * @param urn The URN of the resource to load.
+   * @param cameraType (Optional) The camera type to load. If not included, the default camera type for the resource will be used.
    */
   @Method()
-  public async load(urn: string): Promise<void> {
+  public async load(urn: string, cameraType?: FrameCameraType): Promise<void> {
```
jdm717, line 867: One thought here - could we model this as a set of options versus a singular parameter? Thinking something along the lines of what we do for `applySceneViewState` or `resetView`. I could definitely see the `load` method expanding to support additional "initialization" features down the  road

packages/viewer/src/lib/types/loadableResource.ts
jdm717: Nice! Yeah, I like this approach. I think tying into the `cameraType` prop is even better than what I was originally thinking here, since that's primarily where the unexpected behavior comes in today when you pass a `cameraType` and the resource loads with a different camera type

jdm717 (approved):
LGTM! Just had one thought on how the `load` method accepts a camera type

packages/viewer/src/components/viewer/viewer.tsx
amvertex: Good point - updated!

jdm717 (approved):
LGTM!

PR #657, opened 2025-04-15
[PLAT-5942] Network config validation

jdm717:
## Summary

Introduces some basic validation/sanitation of the `network` properties provided in a configuration object. With this in place, the SDK will now automatically trim whitespace on either end of the URLs provided, and will validate that each URL begins with the appropriate scheme (`https` or `wss`), throwing a descriptive error in the case that these requirements are not met.

## Test Plan

- Verify that out of the box configurations continue to work as expected (`config-env`)
- Verify that providing custom network hosts continues to work as expected
- Verify that providing network hosts with additional whitespace now loads the resource as expected (`sceneTreeHost: 'https://scene-trees.platprod.vertexvis.io '` should now resolve and load a tree for instance)
- Verify that providing invalid network hosts throws a descriptive error (`sceneTreeHost: 'http://scene-trees.platprod.vertexvis.io'` should throw indicating that the `https://` scheme is required)

## Release Notes

N/A

## Possible Regressions

Configuration defaults

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/components/scene-tree/scene-tree.tsx
+++ b/packages/viewer/src/components/scene-tree/scene-tree.tsx
@@ -19,7 +19,13 @@ import { Node } from '@vertexvis/scene-tree-protos/scenetree/protos/domain_pb';
 import { SceneTreeAPIClient } from '@vertexvis/scene-tree-protos/scenetree/protos/scene_tree_api_pb_service';
 import { Disposable } from '@vertexvis/utils';
 
-import { Config, parseConfig, PartialConfig } from '../../lib/config';
+import {
+  Config,
+  parseAndValidateConfig,
+  PartialConfig,
+  sanitizeConfig,
+  validateConfig,
+} from '../../lib/config';
 import { Environment } from '../../lib/environment';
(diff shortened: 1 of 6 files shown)


packages/viewer/src/components/scene-tree/scene-tree.tsx
```diff
@@ -972,7 +978,9 @@ export class SceneTree {
   }
 
   private getConfig(): Config {
-    return parseConfig(this.configEnv, this.config);
+    const parsedConfig = parseAndValidateConfig(this.configEnv, this.config);
+
+    return validateConfig(sanitizeConfig(parsedConfig));
```
amvertex, line 983: It looks like `validateConfig` and `sanitizeConfig` are run as part of `parseAndValidateConfig`, so are they needed here too?

jdm717: Ah good catch. I pulled those two into `parseAndValidateConfig` after realizing we had quite a few places we needed to do this validation + sanitation, but missed removing those here. Updated!

amvertex (approved):

PR #661, opened 2025-04-29
[PLAT-5767] Measurement improvements

amvertex:
## Summary
This PR includes improvements to measurement in order to ensure consistent results.

## Test Plan
Verify precise and point-to-point measurement works as expected.

## Release Notes
None

## Possible Regressions
Measurement

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/types/entities.ts
+++ b/packages/viewer/src/lib/types/entities.ts
@@ -6,7 +6,7 @@
  */
 export enum EntityType {
   /**
-   * A value that represents the presence of a cross section.
+   * A value that represents the presence of a cross-section.
    */
   CROSS_SECTION = 0xff, // == 255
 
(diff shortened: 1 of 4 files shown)


packages/viewer/src/lib/types/entities.ts
```diff
@@ -26,7 +26,7 @@ export enum EntityType {
   IMPRECISE_EDGE = 0xa0, // == 160
 
   /**
-   * A value that represents the presence of an surface without BREP.
+   * A value that represents the presence of a surface without BREP.
```
jdm717, line 29: Nice catch! Honestly we could probably use a pass through the SDK to determine if there are other cases of typos - it doesn't look great in documentation and these seem keep popping up. Definitely not something to include here of course, just a thought. Thanks for fixing these!

jdm717 (approved):
LGTM!

PR #662, opened 2025-05-08
[PLAT-5521] Support reloading scene from viewer element

amvertex:
## Summary
This PR adds the ability to reload the scene from the viewer element. The method disconnects the stream and then immediately reconnects the same scene.

Further, this PR also exports the 'Orientation' type.

## Test Plan
Verify a viewer element can reload its scene

## Release Notes
Ability to reload the scene from the viewer element

## Possible Regressions
None

## Dependencies
None


jdm717 (approved):
LGTM!

PR #664, opened 2025-05-20
adding client to get scene item metadata

mcallydevelops:
## Summary
Adding new controller that allows for the retrieval of scene item metadata


packages/viewer/geometry/src/math.js.map
```diff
@@ -0,0 +1 @@
...E,GAAW;EAC3D,OAAO,IAAI,CAAC,GAAG,CAAC,GAAG,EAAE,IAAI,CAAC,GAAG,CAAC,GAAG,EAAE,KAAK,CAAC,CAAC,CAAC;AAC7C,CAAC;AAED;;;;;;;;;GASG;AACH,MAAM,UAAU,IAAI,CAAC,CAAS,EAAE,CAAS,EAAE,CAAS;EAClD,CAAC,GAAG,KAAK,CAAC,CAAC,EAAE,CAAC,EAAE,CAAC,CAAC,CAAC;EACnB,OAAO,CAAC,GAAG,CAAC,CAAC,GAAG,CAAC,CAAC,GAAG,CAAC,CAAC;AACzB,CAAC","sourcesContent":["/**\n * Clamps the given value between `min` and `max`.\n *\n * @param value The value to clamp.\n * @param min The min possible value.\n * @param max The max possible value.\n * @returns `value` or a value clamped to `min` or `max`.\n */\nexport function clamp(value: number, min: number, max: number): number {\n  return Math.max(min, Math.min(max, value));\n}\n\n/**\n * Linear interpolates a value between `a` and `b` by `t`. If `t` is 0, then the\n * result will be `a`. If `t` is 1, then the result will be `b`. `t` will be\n * clamped to a value between 0 and 1.\n *\n * @param a The start value.\n * @param b The end value.\n * @param t The interpolation value between 0 and 1.\n * @returns The interpolated value between `a` and `b`.\n */\nexport function lerp(a: number, b: number, t: number): number {\n  t = clamp(t, 0, 1);\n  return t * (b - a) + a;\n}\n"]}
```
jdm717, line 1: I'm not sure where this `packages/viewer/geometry` file is coming from (I see it get generated locally occasionally as well, but haven't dug into it), could we remove it?

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -364,6 +365,15 @@ export class Viewer {
    */
   @Prop({ mutable: true }) public pmi: PmiController | undefined;
 
+  /**
+   * The controller for accessing and viewing SceneItems.
+   *
+   * @readonly
+   */
+  @Prop({ mutable: true }) public sceneItemController:
```
jdm717, line 373: How would you feel about renaming this to `sceneItems` and omitting the `Controller` part to follow our other controllers exposed this way? (thining about `pmi` and `modelViews`)

packages/viewer/src/lib/scene-items/controller.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+export interface SceneItemMetadataResponse {
+  paging: PagingLinks;
+  entries: DomainPropertyEntry[];
+}
+
+export interface ListSceneItemMetadataOptions {
+  cursor?: string;
+  size?: number;
+}
```
jdm717, line 59: nit: could we potentially pull these types out into a separate `types.ts` file? I'm anticipating this list will only grow once we add the direct SceneItem endpoint, so it'd be nice to keep those definitions separate from the controller

packages/viewer/src/lib/scene-items/helper.ts
```diff
@@ -0,0 +1,130 @@
...
+    const valueDate = new PropertyValueDate();
+    valueDate.setValue(value);
+    this.value.setDate(valueDate);
+    return this;
+  }
+
+  public build(): PropertyValue {
+    return this.value;
+  }
+}
```
jdm717, line 130: nit: definitely minor, but it looks like these helpers are only used in testing - could we move those to the `src/testing` directory to make that clear?

jdm717 (commented):
Overall this looks great! Just had a few minor points of feedback.

One other thing I did notice is that the `components.d.ts` file has a fairly large diff. That can happen depending on whether `yarn start` or `yarn build` was run most recently, so we typically try to run `yarn build` to prevent that from changing with each PR - if you wouldn't mind running that here as well

jdm717 (approved):
LGTM! Though it looks like you may need to run `yarn format` 🙂

jdm717 (approved):

PR #665, opened 2025-05-21
[PLAT-3730] Option to always show scene tree icons

amvertex:
## Summary
This PR adds a parameter to the scene-tree-table-cell element to set the cell's icons to always be visible. If the parameter is not set, or set to false, then the cell's icons will only be visible when hovering over the cell.

## Test Plan
Verify the parameter can be set to always show the cell's icons

## Release Notes
Option to always show icons in a scene tree table cell

## Possible Regressions
None

## Dependencies
None


jdm717 (approved):
💯

PR #666, opened 2025-05-22
[PLAT-5774] Increase z-index of error and empty results screens

amvertex:
## Summary
This PR increases the z-index of the error and empty result screens in the screen tree in order to prevent some elements, namely the column resize dividers, from appearing over the desired screen.

## Test Plan
Verify only the error or empty result screens are visible when desired

## Release Notes
None

## Possible Regressions
Scene tree views

## Dependencies
None


packages/viewer/src/components/scene-tree/scene-tree.css
```diff
@@ -139,6 +139,7 @@
   box-sizing: border-box;
   justify-content: center;
   background-color: white;
+  z-index: 3000;
```
jdm717, line 142: Curious here, any particular reason for `3000`? Also - one thought would be that it'd be nice to be able to override this in the future via CSS variables. I suspect this may not behave exactly as we expect in every application, so being able to adjust that easily would be nice

jdm717 (approved):
LGTM! Just had one thought on moving to a CSS variable

packages/viewer/src/components/scene-tree/scene-tree.css
jdm717: Fair enough! It looks like I originally added those, so I'm guessing I pulled them from somewhere, but I can't recall exactly where... 🤔

jdm717 (approved):

PR #669, opened 2025-06-03
[PLAT-6168] Example for metadata display, searching and columns

MadisonEhlers-Vertex:
## Summary

These changes introduce a new example to display the metadata columns in the scene tree, searches, and also a way to find the available columns. 

https://github.com/user-attachments/assets/18f79230-b494-4771-9a5e-ec951c88b89e


amvertex (approved):
Looks great! I think this example will be helpful!

PR #682, opened 2025-07-28
[PLAT-6607] Prevent duplicate cursors

amvertex:
## Summary
Users could hit a race condition when using point-to-point measurement where multiple, identical measurement cursors would be added to the viewer's cursor manager. Once in this state, when the custom cursor was to be removed, only one of the measurement cursors was removed from the cursor manager and since a measurement cursor was still present in the cursor manager, the measurement cursor persisted.

Therefore, this PR adds an optional parameter to the 'add cursor' method to prevent adding a duplicate cursor. 

## Release Notes
Ability to prevent duplicate custom cursors

## Possible Regressions
Custom cursors

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/cursors.ts
+++ b/packages/viewer/src/lib/cursors.ts
@@ -115,9 +115,9 @@ export class CursorManager {
     preventDuplicate = false
   ): Disposable {
     if (preventDuplicate) {
-      const hasDuplicate = this.getExistingDuplicateCursor(cursor);
-      if (hasDuplicate != null) {
-        return hasDuplicate;
+      const duplicateCursor = this.getExistingDuplicateCursor(cursor);
+      if (duplicateCursor != null) {
+        return duplicateCursor;
       }
     }
 
--- a/packages/viewer/src/lib/cursors.ts
+++ b/packages/viewer/src/lib/cursors.ts
@@ -111,8 +111,16 @@ export class CursorManager {
    */
   public add(
     cursor: Cursor,
-    priority = CursorManager.NORMAL_PRIORITY
+    priority = CursorManager.NORMAL_PRIORITY,
+    preventDuplicate = false
   ): Disposable {
+    if (preventDuplicate) {
+      const hasDuplicate = this.getExistingDuplicateCursor(cursor);
+      if (hasDuplicate != null) {
+        return hasDuplicate;
+      }
+    }
+
     const id = ++this.nextId;
     const instance = { id, cursor, priority, dispose: () => this.remove(id) };
     this.cursors.push(instance);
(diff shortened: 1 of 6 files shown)


packages/viewer/src/lib/cursors.ts
```diff
@@ -111,8 +111,16 @@ export class CursorManager {
...
-    priority = CursorManager.NORMAL_PRIORITY
+    priority = CursorManager.NORMAL_PRIORITY,
+    preventDuplicate = false
   ): Disposable {
+    if (preventDuplicate) {
+      const duplicateCursor = this.getExistingDuplicateCursor(cursor);
+      if (duplicateCursor != null) {
+        return duplicateCursor;
+      }
+    }
```
jdm717, line 122: Curious here, is there a scenario where we specifically want to track duplicate cursors? Wondering if we should just consider making this the default behavior.

One other thought is that we may also want to reorder the list in this scenario, such that `duplicateCursor` gets shifted to the end of the array, since `getActiveCursor` prioritizes the cursor at the end of the array

jdm717 (commented):
Thanks for fixing this! Just had a couple of thoughts on the handling in the CursorManager, but otherwise LGTM

packages/viewer/src/lib/cursors.ts
```diff
@@ -113,6 +113,12 @@ export class CursorManager {
     cursor: Cursor,
     priority = CursorManager.NORMAL_PRIORITY
   ): Disposable {
+    // Do not add a duplicate cursor
+    const duplicateCursor = this.getExistingDuplicateCursor(cursor, priority);
+    if (duplicateCursor != null) {
+      return duplicateCursor;
+    }
```
jdm717, line 120: 💯 I think this makes sense! Would we still be able to resort `this.cursors` in this case so whatever cursor add is attempted gets shifted to the front of the queue and takes precedence over same `priority` cursors?

packages/viewer/src/lib/cursors.ts
```diff
@@ -133,6 +139,29 @@ export class CursorManager {
...
+   * @param priorityToCheck The priority level to check for duplicates against.
+   */
+  private getExistingDuplicateCursor(
+    cursorToCheck: Cursor,
+    priorityToCheck: number
+  ): CursorInstance | undefined {
+    const cursorMatches = this.cursors.filter(
+      (cursor) =>
+        cursor.cursor === cursorToCheck && cursor.priority === priorityToCheck
+    );
```
jdm717, line 156: One other small nit: this could probably be simplified to
```typescript
const cursorMatch = this.cursors.find(
  (cursor) =>
    cursor.cursor === cursorToCheck && cursor.priority === priorityToCheck
  );
```
since we only use the first result. One other nice side-effect is that the return value can actually just become the result of that `find`, since we don't have to check the array length 🙂

jdm717 (approved):
This looks great!

PR #686, opened 2025-08-11
[PLAT-7036] Update rotation behavior

amvertex:
## Summary
This PR updates the default rotation behavior in the viewer.  Now, when _rotateAroundTapPoint_ in the viewer is true and the user clicks on geometry, then the model will rotate around the point that was clicked. When _rotateAroundTapPoint_ is true and the user clicks in empty space (not on geometry), then the model will rotate around the center of the viewport. Finally, when _rotateAroundTapPoint_ is false, then the model will always rotate around the center of the viewport.

Previously, when _rotateAroundTapPoint_ was true and the user clicked in empty space (not on geometry), then the model rotated around the center of the model, not the center of the viewport, when if the center of the model was not in view.

## Test Plan
Verify models rotate as expected when _rotateAroundTapPoint_ is both true and false 

## Release Notes
Updates default rotation behavior

## Possible Regressions
Model rotation

## Dependencies
None


packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -404,57 +404,54 @@ export abstract class InteractionApi<T extends Camera = Camera> {
...
-          this.worldRotationPoint =
-            depthBuffer != null
-              ? this.getWorldPoint(point, depthBuffer, worldCenter)
-              : camera.lookAt;
-        }
+    return this.transformCamera(({ camera, viewport, depthBuffer }) => {
+      if (this.worldRotationPoint == null) {
+        this.worldRotationPoint =
+          depthBuffer != null
+            ? this.getWorldPoint(point, depthBuffer, camera.lookAt)
```
amvertex, line 411: This is the main change. When the clicked point has no depth (as in the user clicked in the empty space), we now fall back to the camera's lookAt point instead of the worldCenter point.

PR #689, opened 2025-08-25
[PLAT-7114] Update end item icon in the scene tree

amvertex:
## Summary
This PR updates the end item icon in the scene tree to only appear for parent, non-leaf nodes and appear in a lighter shade.

## Test Plan
- Verify the end item icon does not appear for leaf nodes
- Verify the end item icon appears as expected for parent, non-leaf nodes

## Release Notes
None

## Possible Regressions
End item icon

## Dependencies
None


packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.css
```diff
@@ -96,6 +96,12 @@
   cursor: default;
 }
 
+.end-item-icon {
+  width: 14px;
+  height: 14px;
+  fill: var(--neutral-600);
```
jdm717, line 102: Could we make this configurable? Maybe just something along the lines of `--scene-tree-secondary-font-color` (or something similar) with a default of `--neutral-600`? That might also make it possible to have the CSS definition be reusable for any future "secondary" icons we may add

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -507,7 +511,7 @@ export class SceneTreeTableCell {
   }
 
   private getEndItemIcon(): ViewerIconName | undefined {
-    if (this.endItemIndicator && this.node?.endItem) {
+    if (this.endItemIndicator && this.node?.endItem && !this.node?.isLeaf) {
```
jdm717, line 514: Could we potentially add a test case for this as well?

jdm717 (commented):
Overall LGTM! Just had a couple thoughts

packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.css
amvertex: Good idea - updated!

jdm717 (approved):
LGTM!

PR #690, opened 2025-08-29
[PLAT-7125] Support snapping transform widget while rotating

amvertex:
## Summary
This PR adds support for snapping the transform widget to a multiple of a given number when a user rotates the transform widget while holding the shift key.

## Test Plan
- When _degreeToSnapToWhenRotating_ is defined, verify the transform widget snaps to the nearest multiple of the given number when the user rotates the transform widget **while** holding the shift key
- When _degreeToSnapToWhenRotating_ is defined, verify the transform widget does not snap to the nearest multiple of the given number when the user rotates the transform widget **without** holding the shift key
- When _degreeToSnapToWhenRotating_ is not defined, verify the transform widget does not snap to the nearest multiple of the given number when the user rotates the transform widget regardless if the user is holding the shift key

## Release Notes
Support snapping transform widget while rotating

## Possible Regressions
Interacting with the transform widget

## Dependencies
None


jdm717 (commented):
This looks great! Just had a few thoughts, but this is a really nice improvement to the widget

PR #697, opened 2025-09-22
[PLAT-7223] Close WebSocket on vertex-viewer disconnectedCallback

jdm717:
## Summary

Introduces a `pause` and `resume` method to the `ViewerStream` maintained by the `<vertex-viewer>` element that are used when the element is disconnected or connected to the DOM. When disconnected, the `pause` method will close the open WebSocket and keep track of the state to allow the `resume` method to reconnect to the existing Stream when/if the element is re-connected. This fixes an issue where repeatedly disconnecting and reconnecting the element from the DOM could result in a number of open WebSocket connections.

## Test Plan

- Verify that the viewer properly closes WebSocket connections when disconnected from the DOM
- Verify that the viewer properly resumes a Stream when reconnected to the DOM
  - This used to be available for "free" since we kept the WebSocket open, but now is an intentional reconnect action

1. Create a viewer and obtain a reference to the viewer and its parentElement
2. Call `viewer.remove()` - (see that the WebSocket disconnects)
3. Call `viewerParent.appendChild(viewer)` - (see that the WebSocket reconnects)
4. Verify the Stream state remains the same after the reconnect

## Release Notes

N/A

## Possible Regressions

Maintaining of Stream state when the viewer is moved or removed/added within the DOM
- Prior to this change, the above four steps to test would keep a single WebSocket connection each time the actions were performed. This is now a more intentional disconnect + reconnect, with potential to lose state if something doesn't go as expected

## Dependencies

N/A

### DIFF
--- a/packages/viewer/src/lib/stream/stream.ts
+++ b/packages/viewer/src/lib/stream/stream.ts
@@ -109,6 +109,7 @@ export class ViewerStream extends StreamApi {
   private deviceId: string | undefined;
 
   private state: ViewerStreamState = { type: 'disconnected' };
+  private pausedState?: Connected;
   public readonly stateChanged = new EventDispatcher<ViewerStreamState>();
 
   private options: Required<Omit<FrameStreamOptions, 'loggingEnabled'>>;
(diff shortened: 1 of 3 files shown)


packages/viewer/src/lib/stream/stream.ts
```diff
@@ -204,6 +209,28 @@ export class ViewerStream extends StreamApi {
     }
   }
 
+  public pause(): void {
+    if (this.state.type === 'connected') {
+      this.pausedState = this.state;
```
amvertex, line 214: Where does `this.pausedState` get reset when the stream is reconnected?

PR #741, opened 2026-03-31
[PLAT-8217] PMI Documentation

jdm717:
## Summary

Adds a couple new readmes for the `modelViews` controller and `pmi` controller, along with links to both from the `@vertexvis/viewer` package readme. This PR also adds a `pmi` example that loads an NIST model with a handful of model views with PMI annotations to demonstrate loading the views and annotations along with interacting with the annotations.

Additionally, as I was testing some of the small example snippets, I discovered that the `sceneReady` event was being emitted by the `load()` method, even if the stream was unsuccessful in connecting, so a fix for that bug is included in the PR.

## Test Plan

- Verify that the `sceneReady` event is only emitted once when a stream successfully connects
  - This can be tested by repeatedly changing the stream-key, causing the stream to abort and reconnect. This should only result in `sceneReady` events for those streams which have connected successfully, not ones that have been aborted
- Run the `pmi` example and review the code
- Review the documentation added

## Release Notes

N/A

## Possible Regressions

Scene ready event emitting

## Dependencies

N/A


amvertex (commented):
This looks great! Just a few small suggestions.

amvertex (approved):
Looks great! Thanks for putting this documentation together.

PR #743, opened 2026-04-02
[PLAT-8321] Clear cross-section when unloading model view

amvertex:
## Summary
This PR ensures that the active cross-section planes are cleared when unloading a model view.

## Test Plan
Verify the active cross-section planes are cleared when unloading a model view

## Release Notes
Clear cross-section when unloading model view

## Possible Regressions
None

## Dependencies
None

### DIFF
--- a/packages/viewer/src/lib/model-views/controller.ts
+++ b/packages/viewer/src/lib/model-views/controller.ts
@@ -94,6 +94,16 @@ export class ModelViewController {
    * the scene view.
    */
   public async unload(): Promise<void> {
-    this.stream.updateModelView({}, true);
+    await this.stream.updateModelView({}, true);
+
+    // Clear any cross-section added from the model view
+    await this.stream.updateCrossSectioning(
+      {
+        crossSectioning: {
+          sectionPlanes: [],
+        },
+      },
+      true
+    );
   }
 }


packages/viewer/src/lib/model-views/controller.ts
```diff
@@ -94,6 +94,16 @@ export class ModelViewController {
...
+
+    // Clear any cross-section added from the model view
+    await this.stream.updateCrossSectioning(
+      {
+        crossSectioning: {
+          sectionPlanes: [],
+        },
+      },
+      true
+    );
```
jdm717, line 107: I'd wonder if there's ever a scenario where it'd be useful to maintain the section here, or whether this could potentially be something that's done on the FSS side as part of the `updateModelView` call. Not necessarily something we need to do right now, but maybe something to keep in mind as we move all of the cross-sectioning configuration to be stream attributes based

jdm717 (approved):
LGTM!

packages/viewer/src/lib/model-views/controller.ts
amvertex: Discussed this separately, and we're leaving this for now and plan to revisit when we go to support multi-plane cross sections in the future.

PR #744, opened 2026-04-02
[PLAT-8227] PDF paging support

jdm717:
## Summary

Adds support for paging a PDF document. As part of this work, the logic for where the PDF is actually drawn on the canvas has also been updated to keep it centered horizontally (similar to browser PDF viewers). The `index.html` for the doc-viewer package has also been updated with a toolbar for controlling the interaction mode, zoom level, and current page.

Additionally, as part of this work, I've fixed a bug with the `<viewer-walk-mode-tool>` element I noticed after trying to run all the examples after #741. The `onSlotchange` behavior appears to have regressed - I'm assuming this is related to underlying browser implementation, but ultimately adding the teleport tool as a direct child of the `<slot>` element was not properly adding the element, causing the `ensureTeleportToolConfigured` method to always append a new child. This would trigger the `onSlotchange` callback, and since it could not find the element, it would attempt to add it again, repeating this cycle indefinitely. I spent a bit of time trying to get a test to capture this, but ultimately left it as-is with slightly modified tests, as the behavior in the JSDOM simulated environment was not consistent with browser behavior.

## Test Plan

- Verify that the `<vertex-document-viewer>` element can page forward and backward
- Verify that the `<vertex-walk-mode-tool>` does not cause infinite re-renders when loaded without a slotted tool element

## Release Notes

N/A

## Possible Regressions

`<vertex-walk-mode-tool>` default behavior

## Dependencies

N/A

### DIFF
--- a/packages/doc-viewer/index.html
+++ b/packages/doc-viewer/index.html
@@ -29,12 +29,125 @@
         width: 100%;
         height: 100%;
       }
+
+      svg {
+        fill: currentColor;
+      }
+
+      .toolbar {
+        display: flex;
+        gap: 1rem;
+        position: absolute;
+        bottom: 10px;
+        left: 50%;
+        transform: translateX(-50%);
+        z-index: 10;
+        background-color: #eeeeee;
+        padding: 0.25rem;
+        border-radius: 0.5rem;
+      }
+
+      .toolbar-button {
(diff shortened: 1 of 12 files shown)


amvertex (approved):
Looks great!

PR #750, opened 2026-04-14
[PLAT-3828] Update basic pin position

amvertex:
## Summary
This PR updates the basic pin position such that the point of the pin aligns with the position the user clicked.

**Before:**
Note that the user is unable to place a basic pin right on the corner. However, they are able to place a text pin on the corner.
https://github.com/user-attachments/assets/b6d634b0-515e-4fea-96ae-9dc3d92e9302

**After:**
Note that the user is now able to place a basic pin right on the corner. The text pin behavior is unchanged and they are still able to place a text pin on the corner.
https://github.com/user-attachments/assets/2eceae96-bcb8-41c2-8158-abbec9bec0cf

## Test Plan
Verify basic pins appear in the expected position

## Release Notes
Update basic pin position

## Possible Regressions
None

## Dependencies
None


packages/viewer/src/components/viewer-pin-group/viewer-pin-group.css
```diff
@@ -90,6 +90,11 @@
   color: var(--viewer-annotations-pin-color);
   cursor: pointer;
   transition: opacity 0.3s ease-in;
+
+  /**
+   * Place bottom point of the pin at the location instead of the middle of the pin
+   */
+  transform: translate(0px, -12px);
```
aeischeid, line 97: I don't really have a good understanding of how this related to the .pin-anchor class above, but it makes me wonder if this is nested in that should would a `top:` or `bottom: 0` rule potentially be a bit easier to reason about? again, not at all clear to me yet if this is the case, but multiple transform rules can sometimes get a bit unwieldy, so I am curious if you had thoughts on that order too while tackling this.

aeischeid (approved):
❓ is there a convention on a PRs with visual changes for attaching a screenshot or mini gif recording type thing?

not sure how much my code review approval is worth just yet, but 👍

packages/viewer/src/components/viewer-pin-group/viewer-pin-group.css
amvertex: This pin class is [used by basic pins](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/viewer-pin-group/pin-renderer.tsx#L42), while the pin-anchor class above is [used by text pins](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/viewer-pin-group/pin-renderer.tsx#L27), so they are not used at the same time. 

I added videos to the PR description showing the two types of pins before and after this change.

jdm717 (approved):

PR #751, opened 2026-04-14
[PLAT-8195] Emit event when scene tree search has finished

amvertex:
## Summary
This PR adds an event emitter that emits an event when a search has completed in the scene tree.

## Test Plan
Verify the event is emitted when a search has complete in the scene tree

## Release Notes
Emit event when scene tree search has finished

## Possible Regressions
None

## Dependencies
None

### DIFF
--- a/packages/viewer/src/components/scene-tree-search/scene-tree-search.tsx
+++ b/packages/viewer/src/components/scene-tree-search/scene-tree-search.tsx
@@ -64,11 +64,17 @@ export class SceneTreeSearch {
   public value = '';
 
   /**
-   * An event that is emitted when a user has inputted or cleared the search
+   * An event that is emitted when a user has changed or cleared the search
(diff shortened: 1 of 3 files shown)


packages/viewer/src/components/scene-tree-search/scene-tree-search.tsx
```diff
@@ -64,12 +64,18 @@ export class SceneTreeSearch {
   public value = '';
 
   /**
-   * An event that is emitted when a user has inputted or cleared the search
+   * An event that is emitted when a user has changed or cleared the search
```
aeischeid, line 67: nice touch ❇️

aeischeid (approved):
think I am tracking along well enough with how this connects to https://github.com/Vertexvis/connect-app/pull/1144/ to feel like it is a good solution.

PR #757, opened 2026-04-24
[PLAT-8423] Markup repositioning support

jdm717:
## Summary

Adds support for repositioning of markup rendered by the `<vertex-viewer-markup>` element. This repositioning is based on a handful of new properties on the `<vertex-viewer-markup>` element and its corresponding child markup elements:
* `originatingViewport` - Represents the viewport the markup was created within. When provided, all NDC values will be relative to this viewport rather than the markup element's viewport.
* `centeringBehavior` - An optional value to control whether markup content should be centered horizontally, vertically, or both when the markup element's viewport becomes larger than the `originatingViewport`. This defaults to `none`, and will push markup up and to the left assuming that the visual content is placed there.
* `offset` - Another optional value that indicates the current offset of the markup element's viewport from the `originatingViewport` or markup element viewport. This value will transform the markup it is applied to support shifting the markup and displaying it off-screen or partially off-screen.
* `scale` - Another optional value that indicates the current scale of the markup element's viewport relative to the `originatingViewport`. When provided, this is used to recompute the points that make up a markup to ensure that it remains in the same position relative to the content.

All of the above values are used to compute a `transform` applied to SVG groups to shift content and compute points relative to the `originatingViewport` to support fixed markup positioning within dynamic 2D content.

In addition to the above changes around markup components, the `doc-viewer` package has had a couple updates to add new events (`pageLoaded`, `pageDrawn`, and `documentStateChanged`), along with a small update to the rendering behavior to trigger within a `window.requestAnimationFrame` callback. This allows for a better correlation of the rendering of 2D content to the canvas with external components, and was necessary to prevent a noticeable lag between the canvas visually updating and markup repositioning (though there is still a small amount here, it is hard to notice in my experience without looking for it).

## Test Plan

- Verify that 2D document rendering continues to work as expected
- Verify that markup without providing all of the new props mentioned above continues to work in the same way (i.e. it remains fixed at the expected position)
- Verify the behavior of providing the above props
  - Just a note here, `originatingViewport` alone should effectively result in the same visual markup, but providing a `scale` and `offset` alongside it should adjust the behavior. Similarly `centeringBehavior` will only come into play when the `scale` + `originatingViewport` results in a viewport smaller than the canvas size for the markup component.

## Release Notes

N/A

## Possible Regressions

Markup, 2D rendering

## Dependencies

N/A


packages/viewer/src/components/viewer-markup-freeform/viewer-markup-freeform.tsx
```diff
@@ -195,13 +241,12 @@ export class ViewerMarkupFreeform {
...
-  }
-
+  @Watch('originatingViewport')
+  @Watch('offset')
+  @Watch('scale')
   @Watch('bounds')
-  protected handleBoundsChange(): void {
+  @Watch('points')
+  protected recomputePointsFromProps(): void {
     this.updatePointsFromProps();
```
amvertex, line 250: I like that these are combined now!

packages/viewer/src/components/viewer-markup/markup-utils.ts
```diff
@@ -43,12 +44,37 @@ export function getBoundingBox2dAnchorPosition(
...
-  return Point.add(
-    Point.scale(pt, scaleFactor, scaleFactor),
-    Dimensions.center(canvasDimensions)
+  const canvasToContentScaleFactor = Math.min(
+    canvasDimensions.width / contentDimensions.width,
+    canvasDimensions.height / contentDimensions.height
+  );
+  const effectiveScalar = canvasToContentScaleFactor * scale;
+  const contentRelativePoint = Point.add(
+    Point.scale(pt, contentDimensions.height, contentDimensions.height),
```
amvertex: Is this line right?

packages/viewer/src/components/viewer-markup/markup-utils.ts
```diff
@@ -63,10 +89,26 @@ export function translatePointToBounds(
...
+  const canvasToContentScaleFactor = Math.min(
+    canvasDimensions.width / contentDimensions.width,
+    canvasDimensions.height / contentDimensions.height
+  );
+  const effectiveScalar = canvasToContentScaleFactor * scale;
+  const contentRelativeDimensions = Dimensions.scale(
+    contentDimensions.height,
+    contentDimensions.height,
+    dimensions
+  );
```
amvertex, line 107: Is this line right?

amvertex (approved):

packages/viewer/src/components/viewer-markup/markup-utils.ts
jdm717: Yeah, this was intentional to maintain the same behavior as [translatePointToScreen](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/viewer-markup/markup-utils.ts#L48-L52) and [translateDimensionsToScreen](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/viewer-markup/markup-utils.ts#L68) previously, where the NDC values are always computed relative to the canvas height. I spent a bit of time exploring this, but ultimately kept the scalar the same here to avoid breaking previously created / stored markup. Probably worth revisiting at some point though, since this does look a bit odd 🙂

packages/viewer/src/components/viewer-markup/markup-utils.ts
jdm717: Same as above here

packages/viewer/src/components/viewer-markup/markup-utils.ts
amvertex: Could you add a comment for this then? I could see something thinking it's wrong in the future and changing it.

amvertex (approved):
💯

PR #758, opened 2026-04-24
[PLAT-8122] Upgrade build pipeline scripts & bump build target to Es2019

aeischeid:
## Summary
cherry-picking some parts out from the React19 upgrade branch that weren't really React19 specific

- modernize github actions build files some.
- new examples pages and documentation to make initial demo experience a bit nicer
- introduce vite to serve examples and allow examples to be test cases for local builds more easily
- bump build target up to es2019 and remove some outdated polyfills 

## Test Plan
should be able to test with updated nextjs and nuxtjs starters, but also just using the build locally and testing examples should go fairly far too

## Release Notes
improved docs and build artifacts

## Possible Regressions
officially the 2019 target should be well within our supported browser matrix, but some older unsupported browser that might have worked okay might now not as much 

## Dependencies
es2019


.github/workflows/branch.yml
```diff
@@ -6,32 +6,27 @@ on:
...
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - name: "Checkout changes"
-        uses: actions/checkout@v4
-      - name: Set Node Version
```
amvertex, line 15: Are the `Set Node Version` and `Cache dependencies` steps no longer needed?

examples/pmi/README.md
```diff
@@ -0,0 +1,32 @@
+# PMI Example
```
amvertex, line 1: Thanks for adding this!

packages/viewer/readme.md
```diff
@@ -164,25 +162,6 @@ function main() {
 }
 ```
 
-## Polyfills
-
-If you plan on targeting IE11 or Edge <= 18, you'll also need to supply
```
amvertex, line 169: Glad to get this removed!

amvertex (approved):
Looks good! Thanks for cleaning this up.

.github/workflows/branch.yml
aeischeid: same idea is just handled differently - down within the "Setup NodeJS" step there is a `uses` and `with` section that

PR #760, opened 2026-04-24
[PLAT-8122] Remove isPromise typeGuard in favor of instanceof Promise

aeischeid:
This cleanup came from noticing something in the course of testing #729 and then trying to chunk bits out from that PR to keep it from getting to large and out of hand

## Test Plan
automatted tests should cover this pretty well I would think

## Release Notes
Code modernization 

## Possible Regressions
if implementations relied on isPromise being not so robust and allowing promise shaped objects then this could be a problem. Not clear to me how likely that any such cases exists is

## Dependencies
no change


packages/viewer/src/lib/meters.ts
```diff
@@ -40,7 +38,7 @@ export class TimingMeter {
    */
   public measure<T>(target: Promise<T>): Promise<T>;
   public measure<T>(target: (() => T) | Promise<T>): T | Promise<T> {
-    if (isPromise(target)) {
+    if (target instanceof Promise) {
```
aeischeid, line 41: this is the only real change. everything else is essentially cleanup. The big question I don't have a clear answer for is if the particular implementation of `isPromise` is important. A more precise name for it might have been `isPromiseLike`

amvertex (approved):
Thanks for cleaning this up!

PR #762, opened 2026-04-27
[PLAT-8424] Markup size scaling

jdm717:
## Summary

Introduces the ability to scale markup sizes based on the `scale` property introduced with #757. When provided, the `scale` will be multiplied against the various `stroke-width` properties to reduce the size of markup when scale is less than 1 or increase the size of markup when scale is greater than 1.

In addition to the stroke-width scaling, some additional logic has been added to the arrowhead point computation, as this was the only piece that needed recomputed points to scale markup up or down in size.

## Test Plan

- Verify that providing a `scale` value to markup increases or decreases the stroke-width such that the markup appears larger or smaller
- Verify that providing a `scale` value to an arrow markup increases or decreases the size of the arrowhead elements

## Release Notes

N/A

## Possible Regressions

Existing markup re-hydration if a stroke-width value without a length (i.e. `4` instead of `4px`). The defaults have been updated here, but we'll need to change this in Connect.

## Dependencies

N/A


packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
```diff
@@ -292,38 +305,34 @@ export class ViewerMarkupArrow {
         <polygon
           id="line-anchor-arrow-triangle"
           class="head"
-          points={arrowheadPointsToPolygonPoints(arrowheadPoints)}
+          points={arrowheadPointsToPolygonPoints(
+            arrowheadPoints,
+            this.scale ?? 1
```
amvertex: We use `this.scale ?? 1` multiple times. What about adding a private variable that's set once and then used many times in the file?

amvertex (approved):

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
jdm717: Good call! After taking a look at this logic again, I actually ended up just having the default for `scale` be set to 1, since that was always the default behavior, so now we don't have to keep setting a fallback everywhere

amvertex (approved):
💯

amvertex (approved):

PR #763, opened 2026-04-28
[PLAT-8122] React19, Stencil4, & other framework tooling upgrades

aeischeid:
## Summary
Modernization of dependencies and peer dependencies. Specifically to target official React 19 support, but along for the ride is TS 4 -> 5, Stencil 2 -> 4, Jest 27 -> 29, and few smaller bumps for things like ts-lib

### Packaging / Module Format
This release updates the SDK’s packaging to prefer ESM and align package resolution with modern Node/bundler behavior. CommonJS remains supported, so most users should not need to change anything. The main cases to watch for are custom deep imports into dist/ or other package internals, and any hard-coded CDN/file-path references to emitted bundle filenames, which may need to be updated. Viewer packages were also adjusted to be safer to import in SSR/Node contexts.

- @stencil/react-output-target v0.5 → v1.4.2: The new major version generates React wrappers using createComponent from @stencil/react-output-target/runtime instead of the old class-based proxy approach. The old react-component-lib/ directory is removed.
- The rollup.config.js for viewer-react is updated to preserve the generated 'use client' directive for React Server Components compatibility.
- TypeScript v4 → v5: Fixed several incompatibilities introduced by stricter TS5 checking:

## Notes
This PR replaces #729 as that one needed significant merge conflict resolution and was getting out of hand

## Test Plan
unit tests should pass
builds and link checks pass

will have -testing releases to test against nextjs starter with react 19. https://github.com/Vertexvis/vertex-nextjs-starter/pull/224 
nuxtjs-starter also tested with built packages: https://github.com/Vertexvis/vertex-nuxtjs-starter/pull/8 to ensure vue wrappers remain functional and compatible 

testing locally with yarn link 
use `yarn examples:start` in this repo to test new stencil components

connect app with testing builds or yarn linking should also look good

## Release Notes
New major version. 
Drop support for React < 17 

## Possible Regressions
- Beyond old React versions no known regressions, but anything noted in stencil breaking changes could apply https://github.com/stenciljs/core/blob/main/BREAKING_CHANGES.md
- applyPollyfils has been removed

## Dependencies
nextjs and nuxtjs starter apps
https://github.com/Vertexvis/vertex-nextjs-starter/pull/224
https://github.com/Vertexvis/vertex-nuxtjs-starter/pull/6


internal/build/rollup.config.js
```diff
@@ -1,7 +1,107 @@
...
+import fs from 'fs';
+import { builtinModules } from 'module';
+// import { config, input, output, typescript } from '@vertexvis/build-tools';
+import path from 'path';
+import ts from 'typescript';
 
-export default config(
-  input('src/index.ts'),
-  typescript(),
-  output({ enableInlineDynamicImports: true })
```
aeischeid, line 6: The common build-tool config was abandoned for this package. This was somewhat a decision I inherited from the initial draft PR ( #729 ). I have tried to understand if I could update `build-tools` instead, but haven't had too much success with that so far. Maybe for now there is a way to just incorporate this `enableInlineDynamicImports` bit into this custom config? But, also at a bit at a loss on how exactly to accomplish that, or how to test if it is done correctly. Looking for feedback on the later part from @amvertex since that was added recently by them.

aeischeid: think I figured this out, not how to test it, but should be ready to test

internal/build/package.json
```diff
@@ -25,26 +25,30 @@
...
     "lint": "eslint --ext .ts,.tsx,.js,.jsx --ignore-path ../../.gitignore .",
     "start": "jest --watch",
-    "test": "jest",
+    "test": "jest --runInBand",
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
+    "@rollup/plugin-commonjs": "^21.1.0",
+    "@rollup/plugin-node-resolve": "^13.3.0",
     "@vertexvis/build-tools": "^0.10.2",
```
amvertex: Can we remove this dependency then?

packages/doc-viewer-react/package.json
```diff
@@ -18,7 +18,6 @@
   "files": [
     "dist/*"
   ],
-  "private": "true",
```
amvertex: We still want this to be private

packages/doc-viewer/package.json
```diff
@@ -51,7 +51,6 @@
     "loader/",
     "LICENSE"
   ],
-  "private": "true",
```
amvertex: We also still want this to be private

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -214,8 +214,6 @@ export class ViewerMarkup {
    *
    * @param markup The markup to add.
    * @returns The markup element that was created.
-   * @see {@link ViewerMarkups.arrowTemplateId}
-   * @see {@link ViewerMarkups.circleTemplateId}
```
amvertex, line 218: Does
```
  * @see {@link arrowTemplateId}
  * @see {@link circleTemplateId}
```
work here?

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -415,7 +413,7 @@ export class Viewer {
    * Emits an event whenever the user taps or clicks a location in the viewer.
    * The event includes the location of the tap or click.
    *
-   * This event can be used in combination with the {@link VertexViewer.scene scene} method
+   * This event can be used in combination with the `scene` method
```
amvertex: Could `{@link scene}` work?

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -351,9 +351,7 @@ export class Viewer {
    * @internal
    */
   @Prop({ mutable: true })
-  public stencilBuffer: StencilBufferManager = new StencilBufferManager(
-    this.hostElement
-  );
+  public stencilBuffer!: StencilBufferManager;
```
amvertex, line 354: What was the reason behind this change?

packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -1356,7 +1364,10 @@ export class Viewer {
 
         const drawnFrame = await this.canvasRenderer(data);
 
-        if (drawnFrame != null) {
+        if (
+          drawnFrame != null &&
+          frameRenderVersion === this.frameRenderVersion
```
amvertex, line 1369: What was the reason behind this change?

packages/viewer/src/lib/walk-mode/interactions.ts
```diff
@@ -8,7 +8,7 @@ import { ViewerWalkModeOperation, WalkModeModel } from './model';
 export class WalkInteractionHandler implements InteractionHandler {
   private api?: InteractionApiPerspective;
 
-  private interval?: NodeJS.Timer;
```
amvertex, line 11: Would NodeJS.Timeout work here?

packages/viewer/stencil.config.ts
```diff
@@ -28,8 +29,7 @@ export const config: Config = {
   globalStyle: 'src/css/global.css',
   outputTargets: [
     reactOutputTarget({
```
amvertex, line 31: Can we add `stencilPackageName: '@vertexvis/viewer',` to this section?

package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "root",
   "private": true,
-  "nextVersionBump": "patch",
+  "nextVersionBump": "major",
```
amvertex, line 4: @jdm717 what do you think about releasing v1.0.0 on our next release?

packages/viewer/src/components/viewer/viewer.tsx
aeischeid: IIRC this one was around trying to solve memory leak issue. Short story was in tests many viewers get started but teardown was not always cleaning them up or cleaning them up well. so would end up in SIGSEGV type errors where things were just running out of memory. This seemed like it improved the situation though it didn't ultimately solve it. Felt worth keeping, though in practice I don't imagine many users of the component would bump into cases where it would make much difference

packages/viewer/src/components/viewer/viewer.tsx
aeischeid: Related to something in the Stencil 4 upgrade. `this` could be undefined, so `this.hostElement` would error. Setting it inside of connectedCallback avoided that

packages/viewer/src/lib/walk-mode/interactions.ts
aeischeid: potentially. I'll give it a try. 🤔

jest.setup.console.js
```diff
@@ -0,0 +1 @@
+jest.spyOn(console, 'debug').mockImplementation(() => {});
```
aeischeid, line 1: there was quite a bit of console noise in tests. The way I set this up it makes it pretty easy to mute and unmute that. just comment out this line and get all the debugging messages back.

internal/build/package.json
aeischeid: yes, we could/should remove it for now. good catch

packages/doc-viewer-react/package.json
aeischeid: Okay, this was changed in the initial draft PR i started from, so assumed this was desired change. Easy revert.

packages/viewer/src/lib/walk-mode/interactions.ts
aeischeid: works, and is a bit more consistant with other usages in this project, so going with that

package.json
aeischeid: I didn't really mean to do this, sort of just wasn't connecting dots. I was thinking 'major' in the current semver context where minor == major. Well, pushed this to publish-testing branch before I put that together, so now we have a `1.0.0-testing` build on NPM 😬  

Eh, gotta go 1.0 at some point right?! 😅

jdm717: I think I'd be on board with shipping a v1.0.0 with our next release to signal the significant changes being introduced! Plus we're probably a bit overdue to release v1.0.0 - we've been on v0.x.x for just a few years 🙂

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
aeischeid: Your suggestion for fixes for @link instead of just removing the problematic use was spot on. Applied that

packages/viewer/stencil.config.ts
aeischeid: added!

examples/camera-manipulations/index.html
```diff
@@ -14,10 +14,16 @@
...
-      nomodule
-      src="https://cdn.jsdelivr.net/npm/@vertexvis/viewer@0.24.x/dist/viewer.js"
-    ></script>
+    <!--
+        Local development: use the local viewer build output instead of the CDN.
+        Do `yarn workspace @vertexvis/viewer build` and `yarn examples:start`
+        before enabling this block.
+    -->
+
+    <!--
```
aeischeid: @jdm717 sonar doesn't like these commented out bits, but IMO they are quite handy. Is there a way to tell Sonar to ignore these `examples/**/index.html` files ?

packages/viewer-react/.eslintrc.cjs
aeischeid, line 1: these move to .cjs files for now. Longer term goal would be to move to the new eslint v9 config files.

packages/geometry/jest.config.js
```diff
@@ -1,7 +1,26 @@
-const jestConfig = require('@vertexvis/jest-config-vertexvis/jest.config');
+// import jestConfig from '@vertexvis/jest-config-vertexvis/jest.config';
+// TODO: return to using build-tool jest-config once supports ESM
```
amvertex, line 2: Do we have a ticket for this?

packages/doc-viewer/scripts/test-ssr-import.mjs
```diff
@@ -0,0 +1,9 @@
+const modulePath = new URL('../dist/components/vertex-document-viewer.js', import.meta.url);
```
amvertex, line 3: Do you want to merge this file?

examples/camera-manipulations/index.html
jdm717: It looks like I'm not able to change any of the rules there - I'll take a look at getting the correct permissions to do so next week, but in the meantime if you want to get these to stop showing up on the PR, you can always open the link and click the dropdown where it says `Open` and choose `Accept` in the cases where we're not going to take the action Sonar suggests, which should at least cause the Sonar check to pass. It does look like Sonar supports path exclusions though, so it should be possible to get the examples directory omitted later!

aeischeid: I got around this. Ended up just putting this comment in the examples dir readme. Easy enough to grab it from there if doing local testing

packages/geometry/jest.config.js
aeischeid: Not yet, but certainly a good idea. I have multiple followup type tasks I have noted locally. Think I just need to set aside some time for getting those onto a proper backlog

packages/doc-viewer/scripts/test-ssr-import.mjs
aeischeid: Mixed thoughts. There is no inclusion in build or CI, and I don't think that would be necessary, but is it useful enough to stick around for occasional testing if a developer happens to be working on something that would have a chance to have a dependency on browser specific API?

packages/geometry/jest.config.js
amvertex: Sounds good! I think it would be worthwhile to create the different tickets for future improvements.

amvertex (approved):
Looks great!

PR #766, opened 2026-05-01
[PLAT-8423] Fix arrow markup clipping

jdm717:
## Summary

After the changes in #757, `<vertex-viewer-markup-arrow>` elements could experience clipping of the left and right sides of the arrowhead. This was particularly noticeable if the `--viewer-markup-arrow-head-stroke-width` value was set higher than `4px`. This PR re-introduces the `filterUnits="userSpaceOnUse"` parameter we were previously using, but now accounts for the `scale` of the element as well to ensure we scale the default `x`, `y`, `width`, and `height` values up if the scale could cause the markup to appear outside the initially computed bounds via `transform`. In the case of a `scale` less than one, we maintain the default bounds, as the transform will not cause the markup to be clipped.

This PR also includes a small adjustment to the `exports` config of the `@vertexvis/doc-viewer` package to export the `doc-viewer.css` file.

## Test Plan

- Verify that an arrow with the default arrowhead drawn parallel to the X or Y axis on-screen does not have the left and right sides of the arrowhead clipped
- Verify that scaling markup up and down does not cause unexpected clipping

## Release Notes

N/A

## Possible Regressions

Markup clipping when `scale` is provided

## Dependencies

N/A

### DIFF
--- a/packages/doc-viewer-react/package.json
+++ b/packages/doc-viewer-react/package.json
@@ -18,7 +18,10 @@
   "files": [
     "dist/*"
   ],
-  "private": "true",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org",
+    "access": "public"
+  },
   "scripts": {
     "clean": "rm -fr ./dist && mkdir ./dist",
     "prebuild": "yarn clean",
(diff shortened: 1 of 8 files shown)


amvertex (approved):
Thanks for getting this fixed so quickly!

PR #770, opened 2026-05-05
[PLAT-8122] remove obsolete deps on uuid & is-plain-object

aeischeid:
## Summary
Cleanup of obsolete dependencies

## Test Plan
Automated unit tests should cover this pretty well, but will also get incidental coverage in manual and integration testing for 1.0 release

## Release Notes
Some packages that got type info from @types/uuid could just use `: string` if you just need to get the job done and don't care about strict enforcement, or use `${string}-${string}-${string}-${string}-${string}` if you want a bit more

## Possible Regressions
Shouldn't be

## Dependencies
Build off of #763

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -46,13 +46,10 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
-    "fast-deep-equal": "^3.1.3",
-    "is-plain-object": "^3.0.0",
-    "uuid": "^8.3.2"
+    "fast-deep-equal": "^3.1.3"
   },
   "devDependencies": {
     "@types/jest": "^29.5.14",
-    "@types/uuid": "^8.3.4",
     "@vertexvis/eslint-config-vertexvis-typescript": "^0.5.0",
     "@vertexvis/jest-config-vertexvis": "^0.5.4",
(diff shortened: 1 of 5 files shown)


packages/utils/src/uuid.ts
```diff
@@ -8,7 +6,7 @@ export interface UUIDMsbLsb {
 }
 
 export function create(): UUID {
-  return uuid();
+  return globalThis.crypto.randomUUID(); // synonomys with uuid.v4()
```
aeischeid, line 9: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

browser and node env support is solid, so compatibility shouldn't be an issue, but jsDom is an outlier, new versions of jest could help that, or vitest + happy-dom which is what newer versions of lerna guide towards

packages/utils/package.json
```diff
@@ -46,13 +46,10 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
-    "fast-deep-equal": "^3.1.3",
-    "is-plain-object": "^3.0.0",
```
amvertex, line 50: Nice to remove some dependencies!

amvertex (approved):

PR #771, opened 2026-05-05
[PLAT-8378] Markup support for 2D viewer

amvertex:
## Summary
This PR adds support for using the markup tools in both the 2D and 3D viewers.

## Test Plan
Verify the markup tools can be used in both the 2D and 3D viewers

## Release Notes
Markup support for 2D and 3D viewers

## Possible Regressions
Using markup in 3D viewers

## Dependencies
None


packages/viewer/package.json
```diff
@@ -75,6 +75,7 @@
   "dependencies": {
     "@improbable-eng/grpc-web": "^0.15.0",
     "@stencil/core": "^4.0.0",
+    "@vertexvis/doc-viewer": "0.24.5",
```
aeischeid: this getting added as a dependency only for the typescript definition feels not ideal, but real potential I am misunderstanding something here

packages/viewer/src/components/viewer-markup-tool/viewer-markup-tool.tsx
```diff
@@ -10,6 +10,7 @@ import {
   State,
   Watch,
 } from '@stencil/core';
+import {} from '@vertexvis/doc-viewer';
```
aeischeid: just importing for the HTMLVertexDocumentViewerElement type maybe?

packages/viewer/src/components/viewer-markup-freeform/viewer-markup-freeform.tsx
```diff
@@ -107,7 +107,7 @@ export class ViewerMarkupFreeform {
    * `<vertex-viewer-markup>` or `<vertex-viewer>` element.
    */
   @Prop()
-  public viewer?: HTMLVertexViewerElement;
+  public viewer?: HTMLVertexViewerElement | HTMLVertexDocumentViewerElement;
```
aeischeid: 🤔  Seeing this makes me wonder if those two should share a common base type or if one of them should extend from the other maybe.

amvertex: What do you think of this implementation?

aeischeid (approved):
This seems pretty good to me at this point.

PR #772, opened 2026-05-06
[PLAT-8378] Use 'vertex3d' domain

amvertex:
## Summary
This PR updates our email addresses and urls to use 'vertex3d' as the domain instead of 'vertexvis.'

## Test Plan
Verify the email addresses and urls are updated

## Release Notes
None

## Possible Regressions
None

## Dependencies
None


aeischeid (approved):
The churn in the generated readme files makes me wonder about some unpinned version in build tooling, a separate issue from that actual changes in the PR, but just more noticeable in a PR like this.

PR #773, opened 2026-05-08
[PLAT-8375] Fix markup positioning when drafting with scale + offset parameters

jdm717:
## Summary

The changes introduced in #757 and #762 added support for positioning the markup canvas viewport relative to another set of dimensions to support rendering this markup in a 2D viewer when zoomed in or out. These changes accounted for the display of existing markup elements, but did not properly account for drafting of new markup. This PR updates the `translatePointToRelative` method to support passing these scaling parameters, and maps the screen position back to relative coordinates for the `originatingViewport`.

## Test Plan

- Within a `<vertex-viewer-markup>` component with `originatingViewport`, `centeringBehavior`, `offset`, and `scale` set, attempt to draft markup and confirm that it has the following behavior:
  - It is created with NDC values relative to the `originatingViewport` rather than the canvas itself. This can be confirmed by drafting a markup and updating the `scale` or `offset`. 
  - It has its stroke width updated with the `scale` parameter while drafting (zoomed in the markup should be visually much larger when drafting and vice-versa for zoomed out)
- With a `<vertex-viewer-markup>` without any of the above properties set, verify that markup drafting continues to behave as expected

## Release Notes

N/A

## Possible Regressions

Markup drafting

## Dependencies

N/A


packages/viewer/src/components.d.ts
```diff
@@ -4,6601 +4,5022 @@
...
+         */
+        "toggleItemVisibility": (row: RowArg) => Promise<void>;
+        /**
+          * An instance of a `<vertex-viewer>` element. Either this property or `viewerSelector` must be set.
+         */
+        "viewer"?: HTMLVertexViewerElement | null;
+        /**
+          * A CSS selector that points to a `<vertex-viewer>` element. Either this property or `viewer` must be set.
+         */
+        "viewerSelector"?: string;
```
aeischeid, line 316: I've noticed some thrashing on style rules in these files. locally too, build will change them one way, format will change them another. Not related to the core of the PR, but sometime soon probably worth fixing whatever prettier or lint styling layers are competing to reduce diff noise

aeischeid (approved):
Code makes sense to me, though not the most confident in how to manually test it quite yet

packages/viewer/src/components.d.ts
jdm717: Agreed. I've noticed this difference between the way `stencil build --dev` and `stencil build` generate the file, but on some initial investigation it appears as though we don't have a lot of control over that. I'd wonder if we could move to just adding this to the `.gitignore` given it's generated on build anyway

jdm717:
> Code makes sense to me, though not the most confident in how to manually test it quite yet

Yeah, this one is a bit tricky without having used the `<vertex-viewer-markup>` component before. Here's a quick example that sets up drafting of markup:

```html
<!-- Note that this needs to be a child of a vertex-viewer element -->
<vertex-viewer-markup id="markup" mode="create" tool="arrow">
  <vertex-viewer-markup-tool></vertex-viewer-markup-tool>
</vertex-viewer-markup>
```

From there, in the `index.html` file for the `viewer` package (`doc-viewer` will still require the changes around the interaction handler initially), you can update the [main](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/index.html#L65-L70) function to set these properties. For example:

```javascript
async function main() {
  await window.customElements.whenDefined('vertex-viewer');

  const markup = document.getElementById('markup');

  markup.scale = 2;
  markup.offset = { x: 100, y: 100 }
}
```

Prior to the changes in this PR, the above example wouldn't properly position markup (it'd appear with the `offset` being applied to it unexpectedly), but with my changes this should now draft markup where expected. Additionally, the markup should appear larger due to the increased scale.

jdm717:
Also @aeischeid FYI - I pushed up another small commit to this PR (https://github.com/Vertexvis/vertex-web-sdk/pull/773/commits/7ff0052eae4267365f56b504a133e57523bb191f) after some testing in a consuming application

aeischeid (approved):

PR #774, opened 2026-05-11
[PLAT-8122] rendering event cleanup in markup components

aeischeid:
## Summary
While on the hunt to resolve a regression with arrow/circle/freeform markup components noticed this handler using some legacy stencil2 approach, that was worth doing some cleanup.

## Test Plan
markup should render, integration tests should pass

## Release Notes
n/a

## Possible Regressions
shouldn't be - was internal events


packages/viewer/src/components/viewer-markup/viewer-markup.tsx
```diff
@@ -444,20 +444,12 @@ export class ViewerMarkup {
...
-        el.removeEventListener('viewRendered', handleRender);
-        resolve();
-      };
-
-      el.addEventListener('viewRendered', handleRender);
-    });
-
+    if (typeof el.componentOnReady === 'function') {
+      await el.componentOnReady();
+    }
```
aeischeid, line 431: In Stencil 4, component rendering is microtask-based. When`addMarkup` appends the element to the DOM, Stencil schedules its initialization as a microtask *before* the `await addMarkup()` continuation microtask
runs. So `componentDidRender` (and thus `viewRendered`) fires before any listener can be registered — causing an unresolvable Promise deadlock that prevents `reset()` from ever being called. Since the permanent markup element has already rendered by the time execution resumes here, waiting for `viewRendered` is unnecessary

aeischeid (commented):
so far, none of these changes meanigfully fixes the markup not rendering regression, but does potentially quite down certain noise around it, and improve quality / performance somewhat

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
```diff
@@ -183,31 +183,25 @@ export class ViewerMarkupArrow {
...
-  @Event({ bubbles: true })
-  public viewRendered!: EventEmitter<void>;
-
   @Element()
   private hostEl!: HTMLVertexViewerMarkupArrowElement;
 
   @State()
   private elementBounds?: DOMRect;
 
-  private interactionHandler = new ArrowMarkupInteractionHandler(
```
amvertex, line 199: I think we should leave `interactionHandler` as it was. It's more difficult to manage when it can be undefined.

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
```diff
@@ -224,12 +218,6 @@ export class ViewerMarkupArrow {
     }
   }
 
-  protected componentDidRender(): void {
-    if (this.mode === '') {
-      this.viewRendered.emit();
```
amvertex, line 237: Why was this removed?

aeischeid: see this comment. https://github.com/Vertexvis/vertex-web-sdk/pull/774#discussion_r3222764354 was essentially unused because of new nuances in stencil 4 lifecycle

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
aeischeid: reverted

packages/viewer/src/components/viewer-markup-arrow/viewer-markup-arrow.tsx
aeischeid: I rebased this PR to only really include this bit of cleanup. Not urgent stuff

packages/viewer/src/components/viewer-markup/viewer-markup.tsx
amvertex: That makes sense. Thanks for the explanation!

README.md
```diff
@@ -79,16 +79,13 @@ Additionally, the project exposes a top-level script to check code coverage
 ## Bumping Versions
 
 The project's release scripts will automatically bump version based on the
-`nextBumpVersion` that's specified in the projects `package.json` file. When
+`nextVersionBump` that's specified in the projects `package.json` file. When
```
amvertex, line 82: 👍 Good catch!

amvertex (approved):

amvertex (approved):

PR #776, opened 2026-05-12
[PLAT-8122] Target es2022 with useDefineForClassFields: false

aeischeid:
## Summary
This was in part to attempt to address regression with markup rendering, but also just good next steps for modernizing the built stencil components

## Test Plan
ensure working examples, works with starter apps, & passing integration tests in connect-app 

## Release Notes
newer build target might require more polyfill setup if using old browsers, but es2022 should still be well within Vertex browser support matrix

## Possible Regressions
Platforms not supporting es2022 might see regressions.


internal/build/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "$schema": "http://json.schemastore.org/tsconfig",
   "extends": "@vertexvis/typescript-config-vertexvis/tsconfig",
   "compilerOptions": {
-    "importHelpers": true
+    "importHelpers": true,
+    "target": "es2022",
+    "lib": ["es2022", "dom"],
```
aeischeid, line 7: these are used by all other tsconfigs in packages in this repo

amvertex (approved):

PR #778, opened 2026-05-18
[PLAT-8473] Document markup adjustments

jdm717:
## Summary

Makes a couple small adjustments to the SDK behavior with markup in a document viewer. Specifically this PR:
- Adds an `injectViewerApi` helper similar to the `<vertex-viewer>`'s implementation to support auto-injecting this viewer into child elements (specifically `<vertex-viewer-markup>`)
- Updates the `<vertex-viewer-markup-*>` elements to call `this.handleScalingConfigurationChange` when the component loads. This fixes an issue where the interaction handler was not properly getting the values when initially constructed and markup was not rendering correctly

## Test Plan

- Verify the above items work as expected
  - Placing a `<vertex-viewer-markup>` under a `<vertex-document-viewer>` should auto-inject the viewer into the component
  - Drafting markup using a `<vertex-viewer-markup-tool>` and setting `scale`, `offset`, and `originatingViewport` should properly position markup at scale

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/packages/doc-viewer/src/components/document-viewer/document-viewer.spec.ts
+++ b/packages/doc-viewer/src/components/document-viewer/document-viewer.spec.ts
@@ -1,8 +1,18 @@
+jest.mock('../../lib/dom', () => {
+  const original = jest.requireActual('../../lib/dom');
+
+  return {
+    ...original,
+    getAllVertexElementChildren: jest.fn(() => []),
+  };
+});
+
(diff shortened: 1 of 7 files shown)


aeischeid (approved):
I've got things in a funny state locally so not in a good spot to really give it a test run, but code makes sense

PR #779, opened 2026-05-19
[PLAT-8473] Add pan wheel listener to host rather than window

jdm717:
## Summary

Updates the `<vertex-document-viewer>` to register a `wheel` event listener on the host element rather than the window. This prevents scenarios where the content was being panned when scrolling other elements within the window.

Additionally, this PR fixes a bug where the higher-level `wheel` listener was being unregistered when a `pointerup` occurred on the document itself, causing panning to stop working when the event occurred on markup elements.

## Test Plan

- Verify that scrolling content outside of the `<vertex-document-viewer>`'s viewport does not scroll the content of the `<vertex-document-viewer>`
- Verify that clicking within a `<vertex-document-viewer>` with markup added does not prevent panning that is initiated from over the top of a markup element

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


amvertex (approved):
Looks good!

PR #784, opened 2026-05-22
[PLAT-2469] Group stream attribute changes

amvertex:
## Summary
This PR groups stream attribute changes happening in quick succession (within 50 milliseconds of each other) in order to ensure that the correct attributes are applied to the stream. Previously, if multiple attributes changed at once, sometimes one or more of the new attribute values were not applied because they were overwritten by a change set that included a different updated value, but not that one.

## Test Plan
Verify stream attributes can be changed and are updated on the stream as expected

## Release Notes
Group stream attribute changes

## Possible Regressions
Stream attribute changes

## Dependencies
None


jdm717 (approved):
LGTM!

aeischeid (approved):

PR #786, opened 2026-05-27
Release v1.0.0

aeischeid:
## Summary

The highlight behind the move to 1.0 version is a major framework upgrade. Components are now built with Stencil 4 instead of Stencil 2. Stencil 4 has significant lifecycle changes, Mostly this should go unnoticed, but certain implementations may encounter problems if they were dependent on the old style. 

Other significant upgrades come along too including Typescript 5 and compatibility with React 19 while dropping support for React < 17.
And, of course, there are a good handful of other tangential dependency upgrades. 

Other potentially breaking changes:
- Browser shims for web-components removed
-  build target moved up to ES2019
- Default packages are ESM, while .cjs bundles are still provided 

---

- Upgrades (#763, #770) 
- Doc-viewer markup support (#771, #778, #779) 
- Include camera info for model view (#780)
- Fix pin label initialization (#785)
- Fix markup scale when drafting with scale + offset params (#773) 
- Add debug logging support to the measurement controller (#775)
- Group stream attribute changes (#784)


amvertex (approved):
💯

PR #789, opened 2026-05-29
[PLAT-8592] Swap to use hydrated attribute over class

jdm717:
## Summary

While investigating some fixes related to the Stencil upgrade, we encountered some odd behavior where the React wrapper + className updates would cause the `hydrated` class to be removed. This would cause some elements to disappear entirely, and required workarounds to manually re-add the class which was less than ideal.

This PR moves to the attribute-based `hydrated` flag, and updates the corresponding CSS previously referencing the class to play a bit nicer with the React wrapper. Additionally, a couple small changes were made to the `<viewer-dom-element>` to always immediately compute its updated matrix definition when a relevant property changes - this fixes an issue with the way we're initializing the values was not always causing the values to be synchronized correctly on first render.

## Test Plan

- Verify that components continue to get properly hydrated and become visible (particularly looking at any elements in packages/viewer/src/css/hydrated.css)
- Verify that `<viewer-dom-element>`s get properly initialized with the position, scale, and rotation provided before the first render (visually noticeable with how we display comment annotations where they would briefly display in the top-left corner of the viewer, which should not happen after this change)

## Release Notes

N/A

## Possible Regressions

Stencil element hydration

## Dependencies

N/A


packages/viewer/src/css/hydrated.css
```diff
@@ -7,30 +7,30 @@
...
+vertex-scene-tree-toolbar:not([hydrated]) {
   visibility: hidden;
 }
-vertex-scene-tree-toolbar.hydrated {
+vertex-scene-tree-toolbar[hydrated] {
   visibility: inherit;
 }
 
-vertex-scene-tree-toolbar {
+vertex-scene-tree-toolbar:not([hydrated]) {
```
amvertex: As Sonar says, these selectors are duplicated in the file. Can we remove the duplicates?

amvertex (commented):
One comment, but otherwise looks good!

amvertex (approved):

PR #793, opened 2026-06-15
[PLAT-5195] Fix re-render warnings & other console noise seen on test runs

aeischeid:
## Summary
Re-render warnings are potential runtime issue, so addressing those as part of exploring solutions for scroll wheel bug causing frames out of order, and while at it getting a little more cleanup of console complaints seen on test suite runs. 

## Test Plan
Beyond automatted tests doing some exploratory testing on touched components. 

## Release Notes
Preventative maintenance on component lifecycle

## Possible Regressions
Nothing known

## Dependencies
Doing similar work in https://github.com/Vertexvis/vertex-web-ui/pull/258

PR #800, opened 2026-06-18
[PLAT-5195] Improve zoom jitter with out of order frames

aeischeid:
## Summary
`handleMouseWheel` was doing a setTimeout loop with increasing delay, and with certain kinds of scroll patterns this could be problematic 

To solve we do a bit more dynamic loop length and zoom divisions, and also allow passing an extra timer delay offset to keep the interaction timer from hitting its reset as easily while these setTimeouts are still pending

## Test Plan
Setting the getInteractionDelay to return a very low value (eg 5) the issue used to replicate very easily even on less complex models - and especially when using a mouse wheel vs a touchpad. With this change I can very rarely get the jarring jitter I saw before. If I go really wild I can still occasionally get what seems like frames out of order, but they seem at least closer together and not flashing high res frames out of order

## Release Notes
Fix zoom scroll frame jitter

## Possible Regressions
No known

## Dependencies
currently this is branched from #793 because I initially has a suspicion that re-rendering could be part of the issue. Probably not it turns out, so this could be re-based and merged separately, but maybe that will merge ahead of this in any case

### DIFF
--- a/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
+++ b/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
@@ -300,7 +300,7 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
     }
   }
 
-  protected async handleMouseWheel(event: WheelEvent): Promise<void> {
+  protected handleMouseWheel(event: WheelEvent): void {
     event.preventDefault();
 
     if (
@@ -312,28 +312,32 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
         -this.wheelDeltaToPixels(event.deltaY, event.deltaMode) / 10;
       const rect = this.element.getBoundingClientRect();
       const point = getMouseClientPosition(event, rect);
+      const scrollSize = Math.abs(event.deltaY);
 
-      if (Math.abs(event.deltaY) < 10) {
+      if (scrollSize < 12) {
         // For small wheel movements, send a single zoom event.
-        await this.zoomInteraction.zoomToPoint(
+        void this.zoomInteraction.zoomToPoint(
           point,
           delta,
           this.interactionApi
         );
       } else {
-        // For large wheel movements, divide the delta into 10 equal zoom events.
-        // This results in a smoother zoom experience for the end user.
-        for (let index = 1; index <= 10; index++) {
+        const divisions = Math.min(10, Math.ceil(scrollSize / 12));
+        const zoomDelta = delta / divisions;
+        // For larger wheel movements, divide the delta into multiple zoom events with increasing delay
+        // which approximates a smooth zoom deceleration curve for the end user.
+        for (let i = 1; i <= divisions; i++) {
+          const delayMs = i * 5;
           window.setTimeout(() => {
             if (this.interactionApi != null) {
-              const zoomDelta = delta / 10;
               this.zoomInteraction.zoomToPoint(
                 point,
                 zoomDelta,
-                this.interactionApi
+                this.interactionApi,
+                delayMs
               );
             }
-          }, index * 5);
+          }, delayMs);
         }
       }
     }
--- a/packages/viewer/src/lib/interactions/mouseInteractions.ts
+++ b/packages/viewer/src/lib/interactions/mouseInteractions.ts
@@ -204,14 +204,14 @@ export class ZoomInteraction extends MouseInteraction {
   }
 
   /**
-   * if this value gets too low, can cause animation jitter in certain scenarios
+   * If this value gets too low, can cause animation jitter in certain scenarios
    */
   private getInteractionDelay(): number {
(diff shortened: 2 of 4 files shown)


packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -300,7 +300,7 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
     }
   }
 
-  protected async handleMouseWheel(event: WheelEvent): Promise<void> {
+  protected handleMouseWheel(event: WheelEvent): void {
```
aeischeid, line 303: Nothing I noticed waited for this, and I was having troubling thinking of scenarios where this couldn't be set and forget.  Not having it be a promise is just a really small micro optimization, but that can have an impact for interactions that can fire dozens of times per second

packages/viewer/src/lib/interactions/mouseInteractions.ts
```diff
@@ -211,14 +229,15 @@ export class ZoomInteraction extends MouseInteraction {
...
+    extraDelayMs?: number
   ): Promise<void> {
     if (!this.didTransformBegin) {
       await this.beginInteraction(api);
     }
 
-    this.resetInteractionTimer(api);
-    f();
+    this.resetInteractionTimer(api, extraDelayMs);
+    await f();
```
aeischeid: I went back and forth on whether I should add`await` on this line... Obviously I did, but I lean towards taking it back out, I didn't notice any difference in terms of the bug this PR is aiming to fix

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
amvertex: Is there any concern that this would impact how multiple handleMouseWheel events that are triggered very close together behave?

packages/viewer/src/lib/interactions/mouseInteractions.ts
```diff
@@ -191,15 +203,21 @@ export class ZoomInteraction extends MouseInteraction {
       : delta;
   }
 
+  /**
+   * if this value gets too low, can cause animation jitter in certain scenarios
```
amvertex: Nit: Can we capitalize the beginning of this comment and the one below?

amvertex (commented):
Looks good! I would lean towards merging this independently of the other PR instead of merging them together.

packages/viewer/src/lib/interactions/mouseInteractions.ts
aeischeid: For sure. I am the worst at capitalizing when I type! Can't seem to shake the habit.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
aeischeid: Not so far, as noting is awaiting this. Additional it would only impact sequences of small scroll moves, so as it was it had some potential for inconsistency, given the increased setTimeout loop would not await, and would need a different kind of overhaul to make it promise based instead.

amvertex (approved):
💯

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -312,28 +312,32 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
+          const delayMs = i * 5;
           window.setTimeout(() => {
             if (this.interactionApi != null) {
-              const zoomDelta = delta / 10;
               this.zoomInteraction.zoomToPoint(
                 point,
                 zoomDelta,
-                this.interactionApi
+                this.interactionApi,
+                delayMs
```
aeischeid, line 337: Passing this extra bit of delay that ends up getting added to the configured interaction delay helps ensure the interaction timer won't complete before this zoom event completes. Because when that happens the inherent network latency of a higher rez frame getting delivered followed by a subsequent zoom event which if started before that final frame arrives would cause the finalFrame to flash onto the screen out of order

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -312,28 +312,32 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
+        void this.zoomInteraction.zoomToPoint(
           point,
           delta,
           this.interactionApi
         );
       } else {
-        // For large wheel movements, divide the delta into 10 equal zoom events.
-        // This results in a smoother zoom experience for the end user.
-        for (let index = 1; index <= 10; index++) {
+        const divisions = Math.min(10, Math.ceil(scrollSize / 12));
```
aeischeid, line 325: We use to have either 1 or 10 zoom divisions. Now medium sized scroll events can be accounted for. Depending on scroll behavior, device, or usage pattern this could cause scroll to feel a little different.

PR #801, opened 2026-06-23
[PLAT-8593] migrate from rollup 2 to 4, along with plugins

aeischeid:
## Summary
Rollup 4 migration.

Changed:
- Updated Rollup to `^4` across root/shared/package build deps.
- Replaced deprecated Rollup 2-era plugins with `@rollup/plugin-commonjs`, `@rollup/plugin-node-resolve`, and `@rollup/plugin-terser`.
- Updated `@vertexwebsdk/build` peer to `rollup: ^4`.
- Added explicit `rollup` and `@vertexwebsdk/build` devDeps to wrapper packages that already use Rollup configs.
- Added a narrow `threads` resolver shim in [packages/viewer/stencil.config.ts](/Users/aaron/dev/Vertex/vertex-web-sdk/packages/viewer/stencil.config.ts) to preserve legacy worker resolution under the Rollup 4 resolver.

Verified:
- `yarn build` passes for all 11 projects.
- `yarn workspace @vertexwebsdk/build test` passes.
- Built artifact entry files still exist for Rollup-built packages: `dist/bundle.cjs`, `dist/bundle.js`, `dist/index.d.ts`, plus `utils` `dist/browser.js` and `geometry` CDN outputs.
- No remaining `rollup@^2` or deprecated `rollup-plugin-*` references in package manifests/lockfile/source.

Residual note: `@vertexvis/rollup-plugin-web-workers@0.1.0` still declares an old `rollup@^1` peer, so Yarn warns, but the full build passes with the compatibility shim.

## Test Plan
Shape of built packages should all be the same, however worth verifying that example apps like our `nextjs-starter` and `nuxtjs-starter` which depend on these don't have any issues.

## Release Notes
should be no difference in build pacakges available on npm, but if building from source will just need to make sure have latest via a `yarn install`

## Possible Regressions
should be none

## Dependencies
some of this could be simplified or offloaded back to build-tools after this https://github.com/Vertexvis/vertex-web-tools/pull/105


packages/viewer/stencil.config.ts
```diff
@@ -1,11 +1,12 @@
+import commonjs from '@rollup/plugin-commonjs';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import terser from '@rollup/plugin-terser';
 import { Config } from '@stencil/core';
 import { reactOutputTarget } from '@stencil/react-output-target';
 import { vueOutputTarget } from '@stencil/vue-output-target';
 import workers from '@vertexvis/rollup-plugin-web-workers';
+import * as path from 'path';
```
amvertex, line 8: Nit: could use `node:path` over `path`

packages/viewer/stencil.config.ts
```diff
@@ -90,3 +92,21 @@ function workerTypescript(): Plugin {
...
+function resolveThreadsEsm(): Plugin {
+  return {
+    name: 'resolve-threads-esm',
+    resolveId(source: string) {
+      if (source !== 'threads') {
+        return null;
+      }
+
+      // Preserve the legacy resolver behavior. The package exports default
+      // points at index.mjs, which omits `expose`; the module build includes it.
```
amvertex, line 105: 👍

amvertex (approved):
Looks good!

PR #802, opened 2026-06-23
[PLAT-1582] Improve transformations and matrix calculations

amvertex:
## Summary
This PR improves and simplifies the logic around of matrix calculations, particularly as they relate to transformations. We were using matrices in a mixture of row-major and column-major and this PR moves closer to using only row-major form. In general, it should now be clearer which form a given matrix has.

In particular, the _multiply_ function in the Matrix4 class was incorrect. Based on the method description, a user assumes that the method returns the value of the first parameter (A) multiplied by the second parameter (B). In actuality, the method returned transpose(transpose(A) * transpose(B)). The method has been changed to instead return A * B. Code that was previously using the method should flip the inputs to see the same result, since
`oldMultiply(A,B) = transpose(transpose(A) * transpose(B)) = transpose(transpose(B)) * transpose(transpose(A)) = B * A = newMultiply(B,A)`.  

When updating the uses of the _Matrix4.multiply_ function, the following facts were often used:
- The inverse of a rotation matrix is equal to its transpose (note this is not true of general transformation matrices)
- transpose(A * B) = transpose(B) * transpose(A)

## Test Plan
Verify transformations work as expected

## Release Notes
This PR improve transformations and matrix calculations. However, it has several breaking changes in the geometry package. To update to this version:
- _Matrix4.makeRotation(A)_ now returns a matrix in row-major form instead of column-major. Thus, previous uses of  _Matrix4.makeRotation(A)_ should be changed to  _Matrix4.transpose(Matrix4.makeRotation(A))_ to see the same result.
- _Matrix4.multiply(A,B)_ now returns A * B instead of B * A. Thus, previous uses of  _Matrix4.multiply(A,B)_ should be changed to  _Matrix4.multiply(B,A)_ to see the same result.

## Possible Regressions
Transformations and other uses of the geometry package

## Dependencies
None


packages/geometry/src/matrix4.ts
```diff
@@ -576,76 +531,37 @@ export function lookAt(
...
+  res[8] = z.x; res[9] = z.y; res[10] = z.z;
   /* eslint-enable prettier/prettier */
   return res;
 }
 
 /**
- * Returns a post-multiplied matrix.
+ * Returns a post-multiplied matrix equal to ab.
  */
 export function multiply(a: Matrix4, b: Matrix4): Matrix4 {
```
amvertex, line 557: Previously, a user assumes that the method returns the value of the first parameter (A) multiplies by the second parameter (B). In actuality, the method returned transpose(transpose(A) * transpose(B)). The method has been changed to instead return A * B. Code that was previously using the method should flip the inputs to see the same result, since
oldMultiply(A,B) = transpose(transpose(A) * transpose(B)) = transpose(transpose(B)) * transpose(transpose(A)) = B * A = newMultiply(B,A).

packages/geometry/src/matrix4.ts
```diff
@@ -206,26 +191,27 @@ export function makeTRS(
   rotation: Quaternion.Quaternion,
   scale: Vector3.Vector3,
 ): Matrix4 {
+  // T is row-major and r is column-major. Why does this work?
   const t = makeTranslation(translation);
-  const r = makeRotation(rotation);
+  const r = transpose(makeRotation(rotation));
```
jdm717: Curious here, did you want to revisit this? And is this still the case after your changes?

packages/geometry/src/matrix4.ts
jdm717: Interesting, great find! Definitely appreciate the context around this one

packages/viewer/src/components/viewer-transform-widget/__tests__/util.spec.ts
```diff
@@ -165,10 +165,12 @@ describe('vertex-viewer-transform-widget utils', () => {
         ),
       ).toMatchObject(
         Matrix4.multiply(
-          Matrix4.makeRotation(
-            Quaternion.fromAxisAngle(Vector3.left(), Angle.toRadians(45)),
+          Matrix4.makeIdentity(),
+          Matrix4.transpose(
+            Matrix4.makeRotation(
+              Quaternion.fromAxisAngle(Vector3.left(), Angle.toRadians(45)),
+            ),
```
jdm717, line 172: Curious here - and likely relates to my earlier comment on `makeTRS`, but what would need to happen to fix the `makeRotation` helper here such that `transpose` isn't necessary? And is that worth considering as part of this PR?

jdm717 (commented):
This is looking great! Nice to have all of these bugs fixed given how critical the utilities they're in are. I didn't quite get the chance to test this locally today, but had a few thoughts and will plan to do some more focused testing tomorrow

packages/viewer/src/lib/transforms/controller.ts
```diff
@@ -31,23 +31,18 @@ export class TransformController {
...
+          value: true,
+        },
       });
     }
   }
 
-  public async updateTransform(delta: Matrix4.Matrix4): Promise<void> {
+  public async updateTransformController(
+    delta: Matrix4.Matrix4,
+  ): Promise<void> {
```
jdm717: Did some local testing using connect-app and noticed this caused the Quick Mate feature to break without code changes. We can update that there of course, but one thought I had is that it might make sense to keep the `updateTransform` signature around here for backward compatibility and mark it as deprecated - mainly to avoid true breaking changes with this PR

packages/geometry/src/matrix4.ts
```diff
@@ -698,7 +621,7 @@ export function isIdentity(matrix: Matrix4): boolean {
 /**
  * Returns an object representation of a `Matrix4`.
  */
-export function toObject(m: Matrix4): Matrix4AsObject {
```
jdm717, line 701: Same thought here as around the `updateTransformController` method regarding true breaking changes - could we potentially leave `toObject` and have it default to returning the result of `toObjectColumnMajor` while marking it deprecated?

packages/geometry/src/matrix4.ts
```diff
@@ -67,28 +66,14 @@ export function fromValues(
...
-export function fromObject(obj: Matrix4AsObject): Matrix4 {
-  /* eslint-disable prettier/prettier */
-  return fromValues(
-    obj.m11, obj.m12, obj.m13, obj.m14,
-    obj.m21, obj.m22, obj.m23, obj.m24,
-    obj.m31, obj.m32, obj.m33, obj.m34,
-    obj.m41, obj.m42, obj.m43, obj.m44
-  );
-  /* eslint-enable prettier/prettier */
-}
```
jdm717, line 90: Same here as with the `toObject` method - could we potentially leave this here for backward compatibility? Also curious if there was any particular reason this method needed to be removed?

packages/geometry/src/matrix4.ts
```diff
@@ -243,37 +229,6 @@ export function makeBasis(
...
-
-  /* eslint-disable prettier/prettier */
-  return [
-    tx * x + c,       tx * y + s * z,   tx * z - s * y,   0,
-    tx * y - s * z,   ty * y + c,       ty * z + s * x,   0,
-    tx * z + s * y,   ty * z - s * x,   t * z * z + c,    0,
-    0,                0,                0,                1
-  ]
-  /* eslint-enable prettier/prettier */
-}
```
jdm717, line 275: Same thought here as with `fromObject` + `toObject` 🙂

packages/geometry/src/vector3.ts
```diff
@@ -364,9 +369,12 @@ export function rotateAboutAxis(
 }
 
 /**
- * Returns a vector that is multiplied with a matrix.
+ * Returns a vector that is multiplied with a matrix in column-major form.
  */
-export function transformMatrix(vector: Vector3, m: Matrix4.Matrix4): Vector3 {
```
jdm717, line 369: I think this is the last case related to backward compatibility - same thought here as the others, could this be left in and marked deprecated?

jdm717 (commented):
Tested locally using a linked build and this is looking great! Didn't notice any unexpected issues outside of the Quick Mate feature and had a couple thoughts around that

aeischeid (approved):
This is major! 🥳 🥁 

Along the way in reading I definitely had some thoughts about method names containing `ColumnMajor` rather than that being a argument to a shared method that defaults to a preferred/recommended  approach, but the advantages of explicitness probably outweigh tradeoffs involved in doing it otherwise. 

Good work on this!

packages/viewer/src/lib/transforms/controller.ts
amvertex: I reverted the name change.

packages/geometry/src/matrix4.ts
amvertex: I added the `toObject` method back.

packages/geometry/src/matrix4.ts
amvertex: I added the `fromObject` method back.

packages/geometry/src/vector3.ts
amvertex: I added this back as a deprecated method.

packages/viewer/src/components/viewer-transform-widget/__tests__/util.spec.ts
amvertex: As I worked through our existing logic, I found that we were using matrices in a mixture of row-major and column-major form. As a result, there were many places we were quietly performing a transpose. This PR makes those transpose operations more apparent, and hopefully in the future we can iterate on this logic more such that all matrices have the same form.

packages/geometry/src/matrix4.ts
amvertex: I don't think this one is used anywhere, so it seemed better to remove it than invest the time to update it and ensure that it's working correctly.

jdm717 (approved):
LGTM! Thanks for digging into all of this incorrect matrix math. Definitely going to help prevent future headaches trying to implement features requiring that math 😅

PR #804, opened 2026-06-24
Test config and doc viewer cleanup

aeischeid:
## Summary
The maxWorkers setting doesn't seem to be needed any more for viewer test suite after lifecycle cleanup

Doc-viewer still had some of those same re-render type warnings that have been cleaned up elsewhere so made an attempt to also address those here. 

## Test Plan
Exploratory testing in doc-viewer example app is merited.

My current thought is to get this re-based after #810 merges and make a testing release and double up some testing effort

## Release Notes
Cleanup of tests and potential extra runtime rerenders in doc-viewer component

## Possible Regressions
No known

## Dependencies
n/a


aeischeid:
@jdm717 This one is in draft as I haven't had time to test it on demo apps, but is potentially ready to go otherwise

amvertex (approved):

PR #806, opened 2026-06-26
[PLAT-5195] Further improve zoom jitter. 

aeischeid:
## Summary
- Minor name cleanups and cache object performance improvements
- Reuse cameraOptions with correlationId so rapidly subsequent zoom events update use same camera or positions instead of restarting fresh each time

## Test Plan
Exploratory scroll wheel zoom testing - previous issues were much more noticable with large / complex models

## Release Notes
Bug fix 

## Possible Regressions
Zoom for touchpads should be unaffected or also improved, but worth testing


packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -43,7 +43,11 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
 
   protected disableIndividualInteractions = false;
 
-  private computedBodyStyle?: CSSStyleDeclaration;
+  private bodyStyleCache?: {
```
aeischeid, line 46: This was already only used as a cache previously, now it is just a bit narrowed to the parts actually used, and named for its purpose rather than it origin

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -344,35 +348,35 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
+          parseFloat(bodyStyle.getPropertyValue('fontSize')) ||
+          DEFAULT_FONT_SIZE,
+        lineHeight:
+          parseFloat(bodyStyle.getPropertyValue('lineHeight')) ||
+          DEFAULT_LINE_HEIGHT,
+        height:
+          parseFloat(bodyStyle.getPropertyValue('height')) ||
+          window.innerHeight,
+      };
+      window.setTimeout(() => {
```
aeischeid, line 367: The cost of a cache miss is relatively low, to the point it doesn't justify cleartimeouts and resets.

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -268,12 +269,12 @@ export abstract class Camera {
     }
 
     try {
-      const corrId = UUID.create();
+      renderOptions.correlationId ??= UUID.create();
```
aeischeid, line 272: The idea is that because zoom animations can trigger in rapid succession we can help them share information better so that, say the starting location of the camera is inherited from the previous animation rather than getting stale info from the scene state. By doing this we will hopefully be able to then modify the zoom deceleration to a better system rather than a stack of setTimeouts.

packages/viewer/src/lib/interactions/interactionApi.ts
```diff
@@ -224,17 +228,10 @@ export abstract class InteractionApi<T extends Camera = Camera> {
...
   public async transformCamera(
     t: CameraTransform<T>,
-    renderOptions?: CameraRenderOptions
-  ): Promise<void>;
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public async transformCamera(...args: any[]): Promise<void> {
-    const t = args[0];
-    const renderOptions = args[1];
-
```
aeischeid, line 237: Nice to be able simplify the function signature here a bit

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -344,35 +348,35 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
   }
 
   protected wheelDeltaToPixels(deltaY: number, deltaMode: number): number {
-    if (this.computedBodyStyle == null) {
-      this.computedBodyStyle = window.getComputedStyle(document.body);
+    // Cached values are an optimization we can use given mouseWheel
+    // events can happen dozen or hundreds of times per scroll, but body style
+    // is very unlikely to change frequently or while doing wheel movements.
```
amvertex, line 353: Good idea. Thanks for adding the comment too!

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -344,35 +348,35 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
+        lineHeight:
+          parseFloat(bodyStyle.getPropertyValue('lineHeight')) ||
+          DEFAULT_LINE_HEIGHT,
+        height:
+          parseFloat(bodyStyle.getPropertyValue('height')) ||
+          window.innerHeight,
+      };
+      window.setTimeout(() => {
+        this.bodyStyleCache = undefined;
+      }, 4800); // For now hardcoded. Could be derived. eg mouseWheelInteractionEndDebounce * 12
```
amvertex, line 369: Is this worth a follow-up ticket?

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -344,35 +348,35 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
+        deltaY * (this.bodyStyleCache.fontSize * this.bodyStyleCache.lineHeight)
+      );
     } else if (deltaMode === 2) {
       // deltaMode 2 corresponds to DOM_DELTA_PAGE, which computes deltas in pages
-      return this.computedBodyStyle.getPropertyValue('height') != null &&
-        this.computedBodyStyle.getPropertyValue('height') !== '' &&
-        !isNaN(parseFloat(this.computedBodyStyle.getPropertyValue('height')))
-        ? deltaY * parseFloat(this.computedBodyStyle.getPropertyValue('height'))
-        : deltaY * window.innerHeight;
+      return deltaY * this.bodyStyleCache.height;
```
amvertex, line 379: This new logic is so much cleaner!

amvertex (approved):
💯

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
aeischeid: Unsure. I already have a next iteration of this work planned. Depending on how that mouseWheelInteractionEndDebounce config value ends up getting used or adjusted I may just do that then. My concern at the moment is there is no minimum enforced on that value, and I wouldn't want this value getting too low.

PR #810, opened 2026-07-01
[PLAT-8807] Default PDF background color

jdm717:
## Summary

Adds a default background color of `#cccccc` for PDFs such that the primary content of the PDF displays with a letterbox effect. Screenshot for example:

Screenshot 2026-07-01 at 5 53 22 PM

## Test Plan

- Verify that by default the PDF viewer displays with the `#cccccc` background where PDF content is not present
- Verify that this letterboxing is not visible if zoomed in such that the PDF content displays within the entire viewport
- Verify that passing a different value for `--viewer-background` to the `<vertex-document-viewer>` element updates the letterboxing color

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


amvertex (approved):

aeischeid (approved):
Looks good to me too! :shipit:

PR #812, opened 2026-07-06
[PLAT-5195] zoom accumulator + requestAnimationFrame

aeischeid:
## Summary
Use a zoom accumulator and requestAnimationFrame to handle zoomToPoint queue for a series of scroll wheel events instead of a loop of setTimeouts

## Test Plan
Do various zoom behaviors on example pages with a larger model especially with a external mouse with a scroll wheel. See fewer frames flashing

## Release Notes
Further improve zoom frame flashing issue

## Possible Regressions
None

## Dependencies
n/a

### DIFF
--- a/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
+++ b/packages/viewer/src/lib/interactions/baseInteractionHandler.ts
@@ -40,6 +40,7 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
   private lastMoveEvent?: BaseEvent;
   private interactionTimer?: number;
   private keyboardControls = false;
+  private pendingWheelZoom: { point: Point.Point; delta: number } | null = null;
 
   protected disableIndividualInteractions = false;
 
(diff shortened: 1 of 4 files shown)


packages/viewer/src/lib/interactions/mouseInteractions.ts
```diff
@@ -230,13 +222,12 @@ export class ZoomInteraction extends MouseInteraction {
...
     api: InteractionApi,
     f: () => void | Promise<void>,
-    extraDelayMs?: number,
   ): Promise<void> {
     if (!this.isTransforming) {
       await this.beginInteraction(api);
     }
 
-    this.resetInteractionTimer(api, extraDelayMs);
+    this.resetInteractionTimer(api, 48); // extra delay of approx 3 frames at 60fps
```
aeischeid, line 230: This also implicitly sets a floor for the configuration value mouseWheelInteractionEndDebounce.

packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -316,37 +318,25 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
...
-          window.setTimeout(() => {
-            if (this.interactionApi != null) {
-              this.zoomInteraction.zoomToPoint(
-                point,
-                zoomDelta,
-                this.interactionApi,
-                delayMs,
-              );
-            }
-          }, delayMs);
```
aeischeid, line 344: This approach is inherently problematic in that with rapidly subsequent large mouse wheel events you can end up with a strangely staggered queue of zoom animations. In my testing it doesn't seem this is the actual culprit of the remaining zoom jitter issue, but getting this simplified should be useful towards finding the root cause

PR #813, opened 2026-07-07
[PLAT-5195] default lineHeight more consistantly used as multiplier

aeischeid:
## Summary
The line height property being treated as a multiplier of the font size is potentially problematic. if line height were unspecified and the default of 1.2 was used there is no problem, but if line height were specified in px, like 22px, zoom could behave weirdly. if it is specified in em or rem units it is less so, but can still get weird. I think reducing the number of variables used for determining this zoom size behavior is probably a good thing, and this would be just a step in that direction. 

## Test Plan
Tested wheel zooming on a model in the viewer example page

## Release Notes
Changes some internal configs based on css properties that were used to determine wheel zoom size

## Possible Regressions
If were relying on a px size based line height to control zoom, which would likely be unintentional, zoom could behave or feel a bit different, slower or faster, or more or less responsive

## Dependencies
n/a


packages/viewer/src/lib/interactions/baseInteractionHandler.ts
```diff
@@ -353,13 +352,13 @@ export abstract class BaseInteractionHandler implements InteractionHandler {
     // is very unlikely to change frequently or while doing wheel movements.
     if (this.bodyStyleCache == null) {
       const bodyStyle = window.getComputedStyle(document.body);
+      const fontSize =
```
aeischeid, line 355: I think there is an underlying assumption of px values here, and with modern css that is less of a solid assumption. I wonder if we could move something a bit better for inferring these configuration values

jdm717: Yeah, I think that's fair. My guess is that we haven't encountered issues with this calculation lately since it looks like Firefox switched to using `DOM_DELTA_PIXEL` in a recent update. Not sure when that happened, but it used to use `DOM_DELTA_LINE`, and this accounted for that preventing extremely slow zooming. It looks like this was probably over-accounting for that though and scrolling much faster, so good catch!

jdm717 (approved):
LGTM!

PR #814, opened 2026-07-08
[PLAT-8622] Add extra frame validity checks to prevent flyTo on invalid cameras

aeischeid:
## Summary
Fix for https://vertexvis.atlassian.net/browse/PLAT-8622

Add extra frame validity checks which will ultimately prevent scene.flyTo if failed. Also expose that so that if you want to precheck before even attempting flyTo or rendering that is an option

something like 
``` js
import { FrameCamera } from '@vertexvis/viewer';

if (FrameCamera.isValidFrameCamera(modelView.camera)) {
  // Animate the camera to the new model view
  if (scene) {
    await scene
      .camera()
      .flyTo({ camera: modelView.camera })
      .render({
        animation: { milliseconds: 1200 },
      });
  }
} else {
  alert('oh no, bad camera?!')
}
```
Should now be be doable, so that if the modelView.camera is a bunch of zeroVectors that would be an invalid camera  you could catch that and alert the user

eg
```
fovHeight: 0
lookAt: {x: 0, y: 0, z: 0}
up: {x: 0, y: 0, z: 0}
viewVector: {x: 0, y: 0, z: 0}
```

## Test Plan
Test against models with PMI data with bad or missing camera data

## Release Notes
flyTo will more aggressively throw errors for certain camera positions

## Possible Regressions
None expected, but potentially if 0 sets were expected as valid for some reason this approach could be problematic

## Dependencies
n/a


packages/viewer/src/lib/types/frameCamera.ts
```diff
@@ -27,23 +27,41 @@ export type FrameCamera = PerspectiveFrameCamera | OrthographicFrameCamera;
...
+    const lookAtValid = isValidVector(asOrthographic.lookAt);
+    const upValid = isValidNonZeroVector(asOrthographic.up);
+    const fovHeightValid =
+      Number.isFinite(asOrthographic.fovHeight) && asOrthographic.fovHeight > 0;
 
     return viewVectorValid && lookAtValid && upValid && fovHeightValid;
   }
 }
 
+function isValidVector(vector?: Vector3.Vector3): boolean {
```
aeischeid: useful because camera object feeding into these is `Partial<FrameCamera>`

packages/geometry/src/vector3.ts
```diff
@@ -64,6 +64,10 @@ export function isValid({ x, y, z }: Vector3): boolean {
   return [x, y, z].every((v) => isFinite(v) && !isNaN(v));
 }
 
+export function isAllZero({ x, y, z }: Vector3): boolean {
```
amvertex: Nit: What do you think about calling this `isZeroVector`? The method could also benefit from a short description.

packages/viewer/src/lib/types/frameCamera.ts
```diff
@@ -27,23 +27,41 @@ export type FrameCamera = PerspectiveFrameCamera | OrthographicFrameCamera;
...
+    const lookAtValid = isValidVector(camera.lookAt);
+    const positionValid = isValidVector(camera.position);
+    const upValid = isValidNonZeroVector(camera.up);
+    return (
+      lookAtValid &&
+      positionValid &&
+      upValid &&
+      Vector3.magnitudeSquared(
+        Vector3.subtract(camera.lookAt, camera.position),
+      ) > 0
```
amvertex: Can this part be separated into its own variable? I think it would help with readability.

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -66,12 +69,12 @@ export class FlyToExecutor {
   }
 
   public withCamera(camera: FrameCamera.FrameCamera): TerminalFlyToExecutor {
-    return new TerminalFlyToExecutor({
-      flyTo: {
-        type: 'camera',
-        data: camera,
-      },
-    });
+    if (isValidFrameCamera(camera)) {
```
amvertex: We already do some checking for a valid camera for a flyTo operation: https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/stream-api/src/streamApi.ts#L276-L284. Why is that check not catching an invalid camera in this situation?

packages/geometry/src/vector3.ts
aeischeid: That is a better name by a bit. I will use that. Need similar check on our backend scene-service, and will aim to use same name convention there

packages/viewer/src/lib/scenes/camera.ts
aeischeid: I think you're right. The throw here where the flyTo doesn't do that was what appeared to solve the issue in a way, but the problem wasn't with the flyTo it was with a modelView.load that followed that also happened with the same bad camera. Not having a way to catch  or some signal to use to notify a user if a flyTo doesn't complete is a bit of a problem however. The scene/camera just not moving is one signal, I guess, but being able to deliver a toast message or somesuch about the problematic camera and flyTo might be an enhancement we could consider.

aeischeid: streamApi scene.flyTo also does validity checks but it uses the https://github.com/Vertexvis/vertex-web-sdk/blob/5cca7ff5dd9a412fc11608ac63f6296ad95ec67c/packages/stream-api/src/validators.ts#L29 and that ends up happening right after the check in camera.render here in viewer.  I am having trouble sorting out if we can put those checks all together in one spot, as it is it does seem there are redundant checks happening, but there might be some justification for that I am not yet aware of.

packages/viewer/src/lib/scenes/camera.ts
```diff
@@ -276,6 +280,7 @@ export abstract class Camera {
           this.flyToOptions,
           renderOptions.animation,
           this.toFrameCamera(),
+          this.hasValidFrameCamera && renderOptions.skipCameraValidation,
```
aeischeid: This skip option should end up being fairly safe, stream-api also has a layer of validation that is not skippable, so the better option may be to ditch the FrameCamera.isValidFrameCamera altogether or at least make it share logic parts with the stream-api/validator

packages/stream-api/src/validators.ts
```diff
@@ -51,133 +39,79 @@ export function validateCamera(
...
+    validatePositiveNumber(dimensions?.height) &&
+    validatePositiveNumber(dimensions?.width)
+  );
 }
 
-export function validateNumber(number: number): boolean {
-  return number != null && Number.isFinite(number);
+export function validateNumber(
+  number: number | null | undefined,
+): number is number {
```
aeischeid, line 71: TS type guard trickery! With strict null checks in play this lets us do less followup != null checks, because after checking validateNumber TS will know it IS a number

packages/stream-api/src/validators.ts
aeischeid, line 1: Cut this file in half line wise, and there should be some mild performance benefit, which could matter in certain cases where render loops end up calling this, but mostly I think it makes it easier to follow the logic.

packages/stream-api/src/paramsBuilder.ts
aeischeid, line 1: This changes are not related directly to the bug, just some TS improvements I noticed while poking around on stream-api files

packages/stream-api/src/paramsBuilder.ts
```diff
@@ -1,42 +1,38 @@
...
+  [K in keyof S]-?: NonNullable<S[K]> extends V ? K : never;
+}[keyof S];
+
+export function defineBoolean<
+  S,
+  P extends KeysOfType<S, boolean> = KeysOfType<S, boolean>,
+>(param: string, prop: P): ParamsBuilder<S> {
+  return defineValue(param, prop, (v) =>
+    typeof v === 'boolean' ? (v ? 'on' : 'off') : undefined,
+  );
```
jdm717: nit: could we potentially remove this nested ternary for readability? Maybe we could pull the on / off return to a small separate helper? What would you think of something like:

```typescript
export function defineBoolean<
  S,
  P extends KeysOfType<S, boolean> = KeysOfType<S, boolean>,
>(param: string, prop: P): ParamsBuilder<S> {
  return defineValue(
    param,
    prop,
    transformValue({
      guard: (v) => typeof v === 'boolean',
      transform: (v) => (v ? 'on' : 'off'),
    }),
  );
}

function transformValue<T, R>({
  guard,
  transform,
}: {
  guard: (v: unknown) => v is T;
  transform: (v: T) => R;
}): (v: unknown) => R | undefined {
  return (v) => (guard(v) ? transform(v) : undefined);
}
```

That'd also open up the ability to use this for the other methods, for example:

```typescript
export function defineNumber<S, P extends keyof S = keyof S>(
  param: string,
  prop: P,
): ParamsBuilder<S> {
  return defineValue(
    param,
    prop,
    transformValue({
      guard: (v) => typeof v === 'number',
      transform: (v) => v.toString(),
    }),
  );
}
```

Open to thoughts there, but thinking something like that may also make the behavior a bit more clear for all these functions

packages/stream-api/src/validators.ts
jdm717: I like this direction! This aligns a bit more with our server side validators, and lets us build up sets of these smaller validators that can be reused in the future 💯

packages/viewer/src/lib/scenes/mapper.ts
```diff
@@ -297,66 +297,53 @@ export function buildFlyToOperation(
...
+      duration: toProtoDuration(animation.milliseconds),
+    };
+  }
+
+  if (baseCamera != null) {
+    if (skipCameraValidation) {
+      payload.baseCamera = FrameCamera.toProtobuf(baseCamera);
+    } else if (isValidFrameCamera(baseCamera)) {
+      payload.baseCamera = FrameCamera.toProtobuf(baseCamera);
     }
```
jdm717: nit related to the Sonar warning: could this just be:

```suggestion
    if (skipCameraValidation || isValidFrameCamera(baseCamera)) {
      payload.baseCamera = FrameCamera.toProtobuf(baseCamera);
    }
```

jdm717 (approved):
Overall LGTM! Just had a couple thoughts

packages/stream-api/src/paramsBuilder.ts
aeischeid: I like this. 🥇 

No other notes, so basically going to apply it directly as you have suggested

PR #816, opened 2026-07-10
[PLAT-8853] Support including resolution of steam

amvertex:
## Summary
This PR adds support for including the resolution of the image on the stream when starting a stream or updating its dimensions. The resolution is saved in dots per CSS inch (dpi).

## Test Plan
Verify the resolution of the image is sent when starting a stream or updating its dimensions

## Release Notes
Support saving resolution of image

## Possible Regressions
Starting and updating streams

## Dependencies
None


aeischeid (approved):
Excellent! :shipit:

packages/viewer/src/lib/stream/stream.ts
```diff
@@ -186,7 +187,10 @@ export class ViewerStream extends StreamApi {
     ) {
       this.dimensions = fields.dimensions;
       this.ifState('connected', () =>
-        this.updateDimensions({ dimensions: this.getDimensions() }),
+        this.updateDimensions({
+          dimensions: this.getDimensions(),
+          resolution: toPbFloatValueOrThrow(this.getResolution()),
```
aeischeid, line 192: A little unclear to me just seeing the diff here how often this fires. If a lot I wonder if a cached value might be worth it. The chance of device resolution changing mid stream seems low to zero

amvertex: I wanted to fetch the resolution here as well since the dimensions also change when a user drags their browser window to a different screen, such as between their laptop and a secondary monitor. Different monitors have different resolutions, so the value can change.

PR #818, opened 2026-07-15
[PLAT-8643] Retry GetTree until view is ready

MadisonEhlers-Vertex:
## Summary

- Bump scene-tree protobufs to `0.1.33` and enable `require_view_ready` on GetTree requests.
- Retry the transient `NotFound` response when a TreeView is still initializing.
- Use bounded exponential backoff (100ms → 3.2s; 6.3s total) and preserve immediate failures for all other gRPC errors.
- Add coverage for the readiness retry path.

## Why

A GetTree request can arrive before the TreeView actor is ready, previously resulting in an intermittent initial-load failure. The new protobuf flag makes the server return an explicit readiness response, allowing the SDK to retry only this temporary condition.


packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -1223,17 +1225,44 @@ export class SceneTreeController {
...
+          `GetTree failed because the view is not ready. Retrying in ${delay}ms (attempt ${nextAttempt}/${maxAttempts}).`,
+        );
+        currentAttempt = nextAttempt;
+        await Async.delay(delay);
+      }
+    }
 
-      this.client.getTree(req, metadata, handler);
-    });
+    throw new Error('GetTree retry attempts were exhausted.');
```
jdm717: Minor nit: Could we potentially change this error message to be a bit less implementation specific? What would you think of something like:
```
throw new Error(`Failed to retrieve scene tree after ${SceneTreeController.MAX_GET_TREE_ATTEMPTS} attempts.`)
```

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -1223,17 +1225,44 @@ export class SceneTreeController {
...
     jwt: string,
   ): Promise<GetTreeResponse> {
-    return this.requestUnary(jwt, (metadata, handler) => {
-      const pager = new OffsetPager();
-      pager.setOffset(offset);
-      pager.setLimit(limit);
+    let currentAttempt = 1;
+    const maxAttempts = SceneTreeController.MAX_GET_TREE_ATTEMPTS;
+
+    while (currentAttempt <= maxAttempts) {
```
jdm717: One thought here - I'd wonder if this might be better as a recursive call with an `attempt` parameter alongside a JWT provider rather than the fixed JWT (maybe even moving the retry logic up to the `fetchPage` method since it wraps the call to this method in `ifConnectionHasJwt`). Main thought there is that there is some chance that the JWT could expire out from under the method and trigger a failure in scenarios we may not want it to fail

jdm717 (commented):
This is looking great! Definitely a nice improvement to reduce the risk of timeouts here. Had one thought around the JWT expiry, but otherwise LGTM!

jdm717 (approved):
LGTM!

PR #821, opened 2026-07-20
Upgrade to latest vertexvis common configs & back to using build-tools

aeischeid:
## Summary
Upgrade to latest vertexvis common configs & back to using build-tools

lets us move away from commonjs files for config, and set stage for newer esLint & jest versions

## Test Plan
Automatted suites should pass. Test builds in example / starter apps should get some basic exploratory testing. 

## Release Notes
build tool and configuration maintenance

## Possible Regressions
None expected 

## Dependencies
n/a


packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.spec.tsx
```diff
@@ -1265,7 +1265,7 @@ describe('vertex-viewer-transform-widget', () => {
       Vector3.create(1, 1, 1),
     );
     (computeHandleDeltaTransform as jest.Mock).mockImplementation(() =>
-      Vector3.create(2, 2, 2),
+      Matrix4.makeTranslation(Vector3.create(2, 2, 2)),
```
amvertex, line 1268: What was the motivation for this change?

tsconfig.docs.json
```diff
@@ -1,6 +1,11 @@
 {
   "$schema": "http://json.schemastore.org/tsconfig",
   "extends": "./packages/viewer/tsconfig.json",
+  "compilerOptions": {
+    // TODO: Remove once @vertexvis/frame-streaming-protos imports Long in its
+    // declarations and @vertexvis/web-workers/threads publish exports-compatible types.
```
amvertex, line 6: Do we have a ticket for either of these things?

amvertex (approved):

tsconfig.docs.json
aeischeid: No, but good callout, we should. I will create them

packages/viewer/src/components/viewer-transform-widget/viewer-transform-widget.spec.tsx
aeischeid: I worked on this last week, and have forgotten details, apart from test was failing and this is what was needed to get it to pass. I'll look back and see if I can jog my memory.

aeischeid: computeHandleDeltaTransform` is declared to return a `Matrix4`. Somehow older TS config was not catching this mismatch, and mismatch in the mock doesn't actually cause the test to fail apparently, but new config wants the correct type at least.

PR #827, opened 2026-07-29
[PLAT-8309] Add retry button for connection failures

jdm717:
## Summary

Adds a simple retry button to enable an out of the box `reload` call when the viewer enters `connection-failed` state.

Example (note this was forced to "time out" locally initially for demonstration):

https://github.com/user-attachments/assets/1f6a0614-8452-4565-8389-26b1243cc2a7

Additionally, this PR adjusts the TypeDoc setup slightly to ensure that the component-level READMEs are available (noticed the `@slot` was not present anywhere there.

## Test Plan

- Verify that when the `<vertex-viewer>` enters a `connection-failed` state, it displays a retry button
- Verify that providing an element to the `retry` slot overrides the default button

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -1383,6 +1401,12 @@ export class Viewer implements BasicViewer {
     }
   }
 
+  private handleRetry = (): void => {
+    this.reload().catch((e) => {
+      console.error('Error reloading scene', e);
+    });
+  };
```
aeischeid, line 1408: Wouldn't want to overcomplicate too much, but wonder about a minimalist debounce or simple setTimeout here to keep the retry button from being smashed in frustration. 🤔

aeischeid (approved):
Looks good! Comment about debounce not a blocker at all.

packages/viewer/src/components/viewer/viewer.tsx
jdm717: I played around with this a bit, and I'm actually leaning toward not adding the debounce, but open to other thoughts there! Basically the `Retry` button itself is only rendered if `this.errorMessage` is defined, and `reload` first calls `unload`, which clears that out [here](https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/viewer/viewer.tsx#L1102). There's probably a bit of a race condition there, but even at 20x CPU throttling I couldn't double-click the button fast enough to see two invocations (I even tried clicking with a mouse and trackpad simultaneously 🙂). With that in mind, I think it may just effectively add latency to the stream restart

PR #828, opened 2026-08-03
[PLAT-9120] Fix Typedoc publishing to Github pages

jdm717:
## Summary

Fixes our deployment of our Typedoc documentation to Github pages. After introducing the signed commit requirements, the previous approach was failing to push a commit to the `gh-pages` branch, which we've seen in some other repositories. This PR updates the workflow to use the `actions/upload-pages-artifact` and `actions/deploy-pages` actions to avoid the need to manually commit and push to the `gh-pages` branch.

For an example of this workflow succeeding, see https://github.com/Vertexvis/vertex-web-sdk/actions/runs/30845403521/job/91792838980. Note that this will still only run when we fully release the SDK, so merging this PR will not cause a new deployment to occur, but the workflow linked here for testing has pushed the latest documentation.

## Test Plan

- Verify updated docs are available at https://vertexvis.github.io/vertex-web-sdk/index.html
- Verify on next release that documentation is pushed successfully

## Release Notes

N/A

## Possible Regressions

N/A

## Dependencies

N/A

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -76,61 +76,28 @@ jobs:
         run: "yarn install"
       - name: "Test"
         run: "yarn test:ci"
-  # docs:
-  #   runs-on: ubuntu-latest
-  #   needs: [build]
-  #   steps:
-  #     - name: "Checkout changes"
-  #       uses: actions/checkout@v6
-  #     - name: "Download build artifacts"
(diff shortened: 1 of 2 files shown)


aeischeid (approved):

aeischeid:
Nice 🥇

PR #832, opened 2026-08-04
[PLAT-9142] Sonar issue cleanup

jdm717:
## Summary

Currently, Sonar analysis of the master branch is failing due to a number of miscellaneous existing issues. This PR attempts to address most or all of the issues causing some of the categories to be below the level needed to pass.

## Test Plan

- Unit tests / branch checks on this PR
- Verify Sonar checks pass on this PR
- Verify all workflow steps continue to work as expected (some script changes + workflow changes, but should all be fairly minor)
- Verify Sonar checks pass on the master branch after merging

## Release Notes

N/A

## Possible Regressions

Github workflows + scripts

## Dependencies

N/A

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Enable Corepack
         run: corepack enable
       - name: "Install"
-        run: "yarn install"
+        run: "yarn install --ignore-scripts"
       - name: "Lint"
         run: "yarn lint"
   test:
@@ -73,7 +73,7 @@ jobs:
       - name: Enable Corepack
         run: corepack enable
       - name: "Install"
-        run: "yarn install"
+        run: "yarn install --ignore-scripts"
       - name: "Test"
         run: "yarn test:ci"
   docs:
@@ -96,7 +96,7 @@ jobs:
       - name: Enable Corepack
         run: corepack enable
       - name: "Install"
-        run: "yarn install"
+        run: "yarn install --ignore-scripts"
       - name: "Generate docs"
         run: "yarn generate:docs"
       - name: "Test docs"
--- a/examples/metadata-operations/index.html
+++ b/examples/metadata-operations/index.html
@@ -20,6 +20,7 @@
       <vertex-viewer-toolbar data-viewer="viewer">
         <input id="metadata-key" type="text" placeholder="Metadata Key" />
         <input id="metadata-value" type="text" placeholder="Metadata Value" />
+        <input id="test" type="text" placeholder="Sonar testing" />
         <button id="search-btn">Search</button>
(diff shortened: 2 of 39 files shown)


examples/metadata-operations/index.html
```diff
@@ -16,13 +16,16 @@
     ></script>
   </head>
   <body>
-    <vertex-viewer id="viewer" class="viewer">
-      <vertex-viewer-toolbar data-viewer="viewer">
```
amvertex, line 20: Do we not need the toolbar anymore?

packages/viewer/src/components/scene-tree/lib/controller.ts
```diff
@@ -164,7 +164,7 @@ export class SceneTreeController {
   private pages = new Map<number, Page>();
   private activeRowRange = [0, 0];
   private metadataKeys: MetadataKey[] = [];
-  private debugLogs = false;
+  private debugLogs = true;
```
amvertex: Do you want to merge this change?

amvertex (commented):
Thanks for fixing all of these!

examples/metadata-operations/index.html
jdm717: Nope! I've updated this to show the controls outside the viewer - the toolbar was really only used here for positioning, but taking a look at the example, these felt like they fit outside the viewer better anyway

packages/viewer/src/components/scene-tree/lib/controller.ts
jdm717: Good catch! I definitely do not 🙂. This is leftover from some local testing I was doing elsewhere. Updated!

.github/workflows/branch.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Enable Corepack
         run: corepack enable
       - name: "Install"
-        run: "yarn install"
+        run: "yarn install --ignore-scripts"
```
aeischeid, line 53: probably should have noticed and added this when I upgraded Lerna a bit ago

aeischeid (approved):
Nice cleanups! Looks ready to go :shipit:

examples/metadata-operations/index.html
amvertex: Cool! I like how it looks now.

amvertex (approved):

.github/workflows/branch.yml
jdm717: Definitely an easy one to miss! The `yarn install` commands don't often get modified in PRs, so Sonar wasn't adding warnings to PRs frequently, and since the failure shows up post-merge, I've been missing this one as well 😅

PR #835, opened 2026-08-08
[PLAT-9051] Minor version upgrades along with jump to yarn4

aeischeid:
## Summary
While bumping to web-workers 0.2.1 and frame-streaming-protos to latest for some typescript fixes those contain I also went ahead with other minor version dependency upgrades. 

Yarn4 works well in all my testing so far. Think it may be a good moment to make the jump, but easy enough to roll that back if need be or if team would prefer to wait on that

## Test Plan
Exploratory testing in example pages here looking good. build and yarn link looked good in other vertex apps. Also did a `-testing` build release to npm and did further testing in vertex app and starter apps, everything looking good. 

## Release Notes
Dependency maintenance

## Possible Regressions
None expected

## Dependencies
new releases in `@vertexvis/web-workers` package and `@vertexvis/frame-streaming-protos`


package.json
```diff
@@ -50,5 +50,5 @@
   "engines": {
     "node": ">=22.0.0"
   },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  "packageManager": "yarn@4.17.1+sha512.ccbfabf7d7b6b32075088be9386fb9a2e00bb6887ef07fa56effabc890a56d53da1ccc4128d62db245fcbd3961b236d75335bdf7d5320ed6eafb7588b7ad4697"
```
amvertex, line 53: Can you write something up to share with the team about how to switch between yarn versions since we're not moving all of our repos over to yarn 4 right now?

tsconfig.spec.json
```diff
@@ -3,6 +3,7 @@
   "extends": "./internal/build/tsconfig-web.json",
   "compilerOptions": {
     "noEmit": true,
+    // TODO: Remove once threads is replaced or publishes exports-compatible types.
```
amvertex, line 6: Do you still want this?

amvertex (approved):

tsconfig.spec.json
aeischeid: Yes, though hopefully shortlived. I am writting a story to handle threads dependency issue, and move forward with either #837 or #836

package.json
aeischeid: Good idea. I have something rough drafted, but will get it further along and ready for sharing.

PR #837, opened 2026-08-14
[PLAT-9215] move to threadsx to avoid issues with threads dependency 

aeischeid:
## Summary
[vertex/web-workers](https://github.com/Vertexvis/web-workers) is a wrapper around threads that in part makes it nicer to use with rollup, but with threads appearing to be somewhat unmaintained it is worth looking at alternatives. 

Threadsx (https://threadsx.jmaleonard.com) fixes and adds enough to basically make the web-workers project obsolete. This is a working quick pass at using threadsx straight away. 

There is also [a PR to bump up web-workers to threadsx](https://github.com/Vertexvis/web-workers/pull/56) but with some of the enhancements available in threadsx the web-workers project looks to be potentially obsolete. If this merges I would propose to mark that project as deprecated and eventually to archive it. 

There is another draft approach to inline the core functionality that this project actually uses see #836. Obviously only one of these PRs would be needed, but I opened two for the sake of comparing the approaches in concrete terms

## Test Plan
Added browser test page to examples directory here to act as an in browser test for workers since Jest can't meaningfully cover that. Also tested locally in implementations of viewer and images look good. 

## Release Notes
TBD

## Possible Regressions
none expected

## Dependencies
Originally stacked on #835 for sake of using yarn4 for my own convenience but should be able to merge independently if wanted.


packages/viewer/package.json
```diff
@@ -82,12 +82,11 @@
     "@vertexvis/scene-view-protos": "^0.8.5",
     "@vertexvis/stream-api": "1.1.0",
     "@vertexvis/utils": "1.1.0",
-    "@vertexvis/web-workers": "^0.2.1",
     "camel-case": "^4.1.2",
     "classnames": "^2.5.1",
     "date-fns": "^2.30.0",
     "fast-deep-equal": "^3.1.3",
-    "fast-png": "^6.4.0",
+    "fast-png": "^8.0.0",
```
aeischeid, line 89: mostly move to ESM only, but also some mild runtime performance improvements as well as pretty significant bundle size improvement 

Image

aeischeid:
Built test release https://www.npmjs.com/package/@vertexvis/viewer/v/1.1.1-testing.2 from this branch for some further verification and it works great!

aeischeid:
since bringing in the logo for png-decoder testing anyway. Spruce up the examples and landing pages just a bit 

image

image

packages/viewer/package.json
amvertex: Exciting!

examples/metadata-operations/index.html
```diff
@@ -17,15 +17,15 @@
...
-      </vertex-viewer>
+      <vertex-viewer id="viewer" class="viewer"> </vertex-viewer>
       <div class="inputs">
         <label class="invisible" for="metadata-key">Metadata Key</label>
         <input id="metadata-key" type="text" placeholder="Metadata Key" />
         <label class="invisible" for="metadata-value">Metadata Value</label>
         <input id="metadata-value" type="text" placeholder="Metadata Value" />
         <button id="search-btn">Search</button>
         <button id="clear-btn">Clear</button>
+      </div>
```
amvertex, line 28: I'm surprised this didn't cause a problem sooner!

examples/README.md
```diff
@@ -28,6 +28,9 @@ yarn examples:start
 http://localhost:8080/examples/pmi/
 ```
 
+The PNG decoder worker smoke test uses the locally built viewer assets, so run
+the build step before opening `/examples/png-decoder-smoke/`.
```
amvertex, line 32: Is this just saying that step 3 above is required?

package.json
```diff
@@ -30,7 +30,7 @@
     "clean": "lerna clean",
     "validate": "yarn build && yarn test && yarn lint",
     "version:bump": "./scripts/version_bump.sh",
-    "examples:start": "vite --host 0.0.0.0 --port 8080 --strictPort"
+    "examples:start": "vite --host 0.0.0.0 --port 8088 --strictPort"
```
amvertex, line 33: Was this change intentional? If so, we should also update the documentation around running the examples.

aeischeid: Sort of, I can definitely move that back. Other repos especially java/jdk stuff tend to be on 8080 so I had made the change to avoid that for myself, and then it got bundled in the diff unintentionally.  But,  maybe others would like that change as well? If so, as you suggest, getting other docs consistent with it would be something I could take of quick

examples/README.md
```diff
@@ -28,6 +28,9 @@ yarn examples:start
 http://localhost:8080/examples/pmi/
```
amvertex: Can you update this url too?

aeischeid:
I did test this pretty well  using [1.1.1-testing.5](https://www.npmjs.com/package/@vertexvis/viewer/v/1.1.1-testing.5) Seems like things are in good shape!

examples/README.md
aeischeid: pretty much. removed and expanded step 3 a bit instead

amvertex (approved):

PR #838, opened 2026-08-25
[PLAT-9156] mark onTokenExpired viewer prop as deprecated & upgrade to EsLint 9

aeischeid:
## Summary
Marking as deprecatted as event is never emiited and token refresh is now handled internally automatically, so shouldn't be needed.

Eslint 9 is a major version bump, but fully compatible with current config, so no code changes needed though id did migrate code comments for prettier disable blocks to new format to avoid as many lint warnings.

## Test Plan
Ensure builds and tests pass, and documentation deploys

## Release Notes
Mark viewer `onTokenExpired` as deprecated.

## Possible Regressions
None expected

## Dependencies
unintentionally stacked on threadsx experimental branch, but easy to cherry pick off from there


packages/viewer/src/components/viewer/viewer.tsx
```diff
@@ -464,6 +464,8 @@ export class Viewer implements BasicViewer {
   /**
    * Emits an event when a provided oauth2 token is about to expire, or is about to expire,
    * causing issues with establishing a websocket connection, or performing API calls.
+   *
+   * @deprecated Token refresh is handled automatically internally; no replacement is needed
```
amvertex, line 468: 💯

amvertex (approved):
Looks good!

internal/build/eslint.config.js
```diff
@@ -3,4 +3,5 @@ import vertexvisTypescript from '@vertexvis/eslint-config-vertexvis-typescript';
 export default [
   { ignores: ['dist/**', 'coverage/**'] },
   ...vertexvisTypescript,
+  { linterOptions: { reportUnusedDisableDirectives: 'off' } },
```
aeischeid, line 6: if these are set to warn they get 'fixed' as in removed when running format script. Not quite what we want at this time.

PR #839, opened 2026-08-26
[PLAT-9323] Prettier fix

aeischeid:
## Summary
Prettier wasn't actually being broadly applied with the format command because each package does not have its own prettier script. And fine to let prettier be a mono-repo level thing than per package thing, so just switch it to yarn instead of Lerna 

## Test Plan
Nothing to test other beyond regular suite of test automation 

## Release Notes
Dev tooling maintenance

## Possible Regressions
None expected

## Dependencies
n/a


package.json
```diff
@@ -26,7 +26,7 @@
     "typecheck:spec": "tsc -p tsconfig.spec.json",
     "lint": "lerna run lint",
     "format:prettier": "prettier --write --list-different .",
-    "format": "lerna run format && lerna run format:prettier",
+    "format": "lerna run format && yarn format:prettier",
```
aeischeid, line 29: everything else in diff is a result of this change followed by running `yarn format`

I did explore using trailingComma: "es5" setting, just in case that quieted down the change from being as drastic, but that made the diff even bigger.

packages/geometry/src/matrix4.ts
```diff
@@ -12,10 +12,22 @@ import * as Vector3 from './vector3';
  */
 export type Matrix4 = [
   // prettier-ignore
-  number, number, number, number,
```
amvertex, line 15: Could we leave the changes in this file in their prior format? They are all marked as "// prettier-ignore", but that must need updated.

aeischeid: absolutely. Good eye. I will work on the comment line. I think it needs to just be moved up one line to take effect, but wonder how that will play with the doc comments, but either way I will figure that out

amvertex: Thanks! Can you do that for the other two matrices in this same file?

aeischeid: Aw dang, already did, but looks like I pushed without that commit.  sorry to make you look at it again.

amvertex: No worries, thanks!

amvertex (approved):

PR #840, opened 2026-08-27
[PLAT-9287] add ariaLabel for viewer-button

aeischeid:
## Summary
`viewer-button` accepts an optional aria label to improve accessibility 

Used on default toolbar icon button

## Test Plan
Should show up in DOM inspection or if using screen-reader tools. 

## Release Notes
Improve a11y for viewer-button. 

## Possible Regressions
None expected

## Dependencies
n/a - stacked on prettier branch so I don't have to potentially re-apply prettier fixes

PR #842, opened 2026-09-01
[PLAT-9287 ] Add aria labels on some elements not otherwise clearly labeled

aeischeid:
## Summary
Followup task from #840, finding and labeling some other elements that didn't have labels and could benefit from them

## Test Plan
inspect DOM and/or test with a11y tools 

## Release Notes
accessibility improvements for a few elements and inputs that were clearly labeled

## Possible Regressions
None expected

## Dependencies
n/a

### DIFF
--- a/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
+++ b/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
@@ -246,6 +246,7 @@ export class SceneTreeTableCell {
               onPointerUp={this.createActionPointerUpHandler(
                 this.toggleExpansion,
               )}
+              aria-label="Expand"
             >
               {expansionIcon && (
                 <vertex-viewer-icon
@@ -261,6 +262,7 @@ export class SceneTreeTableCell {
             <button
               class="end-item-btn no-shrink"
               data-testid={'end-item-' + this.node?.name}
+              aria-label="End item"
             >
               {endItemIcon && (
                 <vertex-viewer-icon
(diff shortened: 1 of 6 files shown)


packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx
```diff
@@ -246,6 +246,7 @@ export class SceneTreeTableCell {
               onPointerUp={this.createActionPointerUpHandler(
                 this.toggleExpansion,
               )}
+              aria-label="Expand"
```
amvertex: This button is can be either expand or collapse (https://github.com/Vertexvis/vertex-web-sdk/blob/master/packages/viewer/src/components/scene-tree-table-cell/scene-tree-table-cell.tsx#L507-L512), so maybe this label should change depending on the icon?

aeischeid: Good observation. I tend of avoid changing labels for these since I heard once from someone who uses a11y tools that it is better to have it just describe the function generically rather than have it change and need to re-read it (when thinking of a screen reader specifically). That may be outdated advice, I am not sure, but it can also avoid code to track state change, so maybe instead of "expand", "Expansion toggle" or "Expand/Collapse toggle"

aeischeid:
```suggestion
              aria-label="Cell expansion toggle"
```

amvertex: The advice makes sense to me. "Cell expansion toggle" sounds good!

amvertex (approved):

